PR #7, opened 2016-12-06
Set systemd as the chosen init manager

OYTIS:
We may/should probably investigate the possibility of running OSTree without systemd and initramfs.


cajun-rat:
Review Score: 11/11

PR #12, opened 2016-12-20
Make systemd optional

OYTIS:
Systemd is not really needed for OSTree, and it would be nice to give users a choice.


OYTIS:
The last commit mirrors this pull request in meta-oim: https://github.com/openivimobility/meta-oim/pull/138

alexhumphreys:
:+1: 

Makes sense, if it build ok then feel free to merge.

OYTIS:
Yes, works fine with and without systemd. Thanks!

PR #29, opened 2017-01-30
sota-client-cpp: Update version

leon-anavi:
Update the version of SOTA client C++ to the
latest version from 25 Jan 2017.


cajun-rat:
Tested on qemu

PR #35, opened 2017-02-20
Fix a bug with missing GNU_HASH in lshw under Yocto

OYTIS:
The bug was found when building against Yocto morty, but won't hurt having it in Krogoth as well.


cajun-rat:
We should think about a 'morty' branch for meta-updater, but I'd like to figure out how we are going to test the various combinations before getting too far down that path. I'd prefer to avoid 3 pull requests and 6 commits for every change to Aktualizr :)

OYTIS:
@cajun-rat I think it's OK to keep krogoth branch as a legacy and push all new features to morty. Otherwise we won't be able to maintain this repo, that's for sure.

PR #41, opened 2017-04-06
Fix small bugs, fix WIC for minnowboard

OYTIS:


cajun-rat:
Tested with OSTREE_BOOTLOADER = "grub"

PR #45, opened 2017-04-11
Add tweaks for compatibility with AGL

OYTIS:
Added _sota overrides to configuration snippets, so that they can be included regardless of whether we are using agl-sota or not. Fixed otaimg in wic.


cajun-rat:
Review Score: 11/11

PR #47, opened 2017-04-11
Add configuration for grub-efi on Minnowboard

OYTIS:


cajun-rat:
Tested that this boots on MinnowMax
Review Score: 11/11

PR #49, opened 2017-04-18
Create a wrapper so OSTree can find ostree-grub-generator

cajun-rat:
OSTree has a hardcoded path to ostree-grub-generator, this wrapper sets it
dynamically to avoid problems with sharing a SSTATE cache.


OYTIS:
Review score: 11/11

PR #51, opened 2017-04-19
Set default DISTRO, disable prelink

cajun-rat:
Setting the default DISTRO to include systemd means that the rvi-sota-client
starts by default, making it work 'out of the box'.  image-prelink seems to be
broken at the moment, so I've disabled it by default.


cajun-rat:
Since non-systemd boot isn't fully working, we'll remove support for it instead

PR #54, opened 2017-04-21
Mount the root filesystem by label rather than path

cajun-rat:
Support identifying the root filesystem via a kernel command line like
'ostree_root=LABEL=mylabel'. Also make it mount a filesystem labeled 'otaroot'
by default.

This means it is possible to test the Minnowboard build in Qemu.


OYTIS:
Review score: 11/11

PR #59, opened 2017-05-05
Add support for am335x-evm based board with WiFi

OYTIS:


cajun-rat:
Review Score: 11/11

Test built against minnowboard/qemu/rpi using meta-updater/scripts/qa

PR #60, opened 2017-05-09
Disable connman, systemd-networkd is used instead

OYTIS:
Even when VIRTUAL-RUNTIME_net_manager is set to "systemd", connman rewrites /etc/resolv.conf


cajun-rat:
Does this affect the 'normal' poky build?

OYTIS:
@cajun-rat 
Yet, I haven't even tested it with AGL, and do not think they will want this board supported anyway.
I've systemd-networkd at some point when a lot of things were broken just to be closer to the reference rootfs. Maybe we could return back to connman, I'll check that.

cajun-rat:
Review Score: 11/11

This only affects one board

PR #65, opened 2017-05-23
Extract common distro configuration into sota.conf.inc

cajun-rat:
Pull the common lines from the poky-sota.conf and poky-sota-systemd.conf
example distributions into a separate include file that can be pulled into
other environments.

This might be by including it in a user's own custom distro, or by directly
including it in local.conf (e.g. AGL).

  require conf/distro/sota.conf.inc


OYTIS:
Review score: 11/11

PR #71, opened 2017-05-26
PRO-3157 Add recipe for friendly name generator

tkfu:
This adds the recipe in preparation for https://github.com/advancedtelematic/rvi_sota_client/pull/234 getting merged.


tkfu:
This should be ready to merge, with the one caveat that I wasn't quite sure if I did the SRC_URI md5sum and sha256sum right. There weren't any errors or warnings about it when bitbaking, though, so it seems to be fine.

PR #74, opened 2017-06-01
Build OSTree with curl support

OYTIS:
libsoup has issues with mutual TLS authentication


cajun-rat:
Tested with a full update (adding man) on qemu
Review Score: 11/11

PR #76, opened 2017-06-06
Include package version number in build

cajun-rat:
Aktualizr will report its version number via the User-Agent header. This is
useful to identify which versions of the client are in the field. The version
information needs to be passed in to the cmake build.  In this case we use
${PV}, the package version number.


cajun-rat:
Included in #77

PR #79, opened 2017-06-08
Set default SOTA_CLIENT directly in sota.bbclass

OYTIS:
In some use cases (etc. AGL) we don't include sota.conf.inc which will result in unset SOTA_CLIENT. Surprisingly, unset variables ARE NOT resolved to empty strings by bitbake, so the build fails in these cases.

### DIFF
--- a/classes/sota.bbclass
+++ b/classes/sota.bbclass
@@ -5,6 +5,7 @@ python __anonymous() {
 
 OVERRIDES .= "${@bb.utils.contains('DISTRO_FEATURES', 'sota', ':sota', '', d)}"
 
+SOTA_CLIENT ??= aktualizr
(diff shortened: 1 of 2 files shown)


classes/sota.bbclass
```diff
@@ -5,6 +5,7 @@ python __anonymous() {
 
 OVERRIDES .= "${@bb.utils.contains('DISTRO_FEATURES', 'sota', ':sota', '', d)}"
 
+SOTA_CLIENT ??= aktualizr
```
cajun-rat: Are double quotes unnecessary/wrong here?

OYTIS: Oops. Yes, sure, we need double quotes.

cajun-rat:
Build tested OK on qemu/rpi/minnowmax
Review Score: 11/11

PR #87, opened 2017-06-26
Override IMAGE_ROOTFS_EXTRA_SPACE

cajun-rat:
AGL sets this to 500MB to allow extra space in the default image.  Set it to 0
in the initramfs.


leon-anavi:
Thanks! Looks good to me. With this patch and the latest version of meta-agl (as of today) I am able to successfully complete the booting process of AGL with agl-sota on Raspberry Pi. 

Btw some other known bugs with agl-sota remain. They are described in AGL SPEC-671 but they are not related to this patch.

Best regards,
Leon

PR #88, opened 2017-06-28
PRO-3341 PRO-3342 Bump aktualizr version to add sw/hw reporting

tkfu:


tkfu:
Tested on raspberry pi, everything seems to be in order.

cajun-rat:
Since Leon is happy with this revision too (see #89), merging.

PR #89, opened 2017-06-28
aktualizr: Update to 7ca3af4

leon-anavi:
Update Aktualizr to the latest version.


tkfu:
#88 bumps to the same rev, but also includes the new runtime dependency on lshw.

PR #96, opened 2017-06-29
Use compression on output file to save space

dl9pf:
This could also be an option for the user to choose.


cajun-rat:
Needs a port of #99 to pyro

OYTIS:
Is covered by https://github.com/advancedtelematic/meta-updater/pull/101

PR #102, opened 2017-07-05
image_types_ostree.bbclass: Fix path to unzip

leon-anavi:
The creation of images, such as rpi-basic-image failed
on branch Pyro due to missing/wrong path to unzip.
This patch ensures that unzip is called with its
full path from the bindir.

### DIFF
--- a/classes/image_types_ostree.bbclass
+++ b/classes/image_types_ostree.bbclass
@@ -5,7 +5,8 @@ inherit image
 IMAGE_DEPENDS_ostree = "ostree-native:do_populate_sysroot \ 
 			openssl-native:do_populate_sysroot \
 			virtual/kernel:do_deploy \
-			${OSTREE_INITRAMFS_IMAGE}:do_image_complete"
+			${OSTREE_INITRAMFS_IMAGE}:do_image_complete \
+			unzip-native"
 
 export OSTREE_REPO
 export OSTREE_BRANCHNAME


classes/image_types_ostree.bbclass
```diff
@@ -41,7 +41,7 @@ IMAGE_CMD_ostreecredunpack () {
 	if [ ${SOTA_PACKED_CREDENTIALS} ]; then
 		rm -rf ${DEPLOY_DIR_IMAGE}/sota_credentials
 
-		unzip ${SOTA_PACKED_CREDENTIALS} -d ${DEPLOY_DIR_IMAGE}/sota_credentials
+		${bindir}/unzip ${SOTA_PACKED_CREDENTIALS} -d ${DEPLOY_DIR_IMAGE}/sota_credentials
```
cajun-rat: Not sure this is the right fix

cajun-rat: Maybe we need to explicitly depend on unzip (-native)?

leon-anavi: I have updated the patch and added unzip-native as a dependency. 

Btw according [to the mega manual of the Yocto Project for Pyro](http://www.yoctoproject.org/docs/2.3/ref-manual/ref-manual.html#required-packages-for-the-host-development-system) (aka version 2.3) unzip is part of the essential packages that have to be installed on the host.

PR #103, opened 2017-07-14
aktualizr: Update to 207898e

leon-anavi:
Fix issues with the random generated device ID.


cajun-rat:
Tested on qemux86-64

Also the same Aktualizr commit is tested OK in #105

PR #105, opened 2017-07-20
Upgrade aktualizr. Full Uptane support for primaries, some uptane sup…

OYTIS:
…port for secondaries, bugs fixed

Tested on OTA stick. Updates successfully, reports status correctly, no redundant installs, old deployments garbage collected properly.


cajun-rat:
Dupe of #103

PR #107, opened 2017-07-21
Split rvi-sota-client into multiple binaries

taheris:
Part of [PRO-3377](https://advancedtelematic.atlassian.net/browse/PRO-3377).


cajun-rat:
Would it make more sense to keep the existing rvi_sota_client recipe name?  That way people won't have to change their existing code, and can continue to include 'rvi_sota_client' in their images as before?

PR #109, opened 2017-07-24
Do not enforce IMAGE_FSTYPES (board specific)

dl9pf:
IMAGE_FSTYPES += "otaimg wic" and IMAGE_CLASSES += "image_types_ota"   are board specific extensions, move down into board specific files.


cajun-rat:
@OYTIS I'm guessing we will need to add this back to the sota_$MACHINE classes?

dl9pf:
yes, that's the consequence. 
Maybe I missed a var. 
I think the main question is:
- what can a board that is not yet vetted  (e.g. does ostreepush make sense w/o otaimg ?)
-- i see a usecase here (aka just produce the repo, omit the image build for distribution)

- mid-term: it would be great if we would not need the custom image somehow. Initrd + image provided by the board .

cajun-rat:
@dl9pf Is this needed still, or do you have a solution already?

PR #119, opened 2017-08-10
Delete treehub.json from the credentials package on the device

OYTIS:


classes/image_types_ostree.bbclass
```diff
@@ -133,6 +134,8 @@ IMAGE_CMD_ostree () {
 	# deploy SOTA credentials
 	if [ -n "${SOTA_PACKED_CREDENTIALS}" ]; then
 		cp ${SOTA_PACKED_CREDENTIALS} var/sota/sota_provisioning_credentials.zip
+		# Device should not be able to push data to treehub
+		zip -d var/sota/sota_provisioning_credentials.zip treehub.json
```
Danten: What does this do if three is no `treehub.json`? I wonder for example if this gets renamed at some point and we forget to remove the new name. The `zip` program seems to exit with error code `12`, will that abort the whole build?

OYTIS: Yes, it will fail and abort the build (just checked).

OYTIS:
@Danten Sure. Just unpack it and repack again. But that's kind of a part of what we wanted to avoid doing. Do you think it is necessary/strongly desirable?

Danten:
@OYTIS, I want to avoid future security issues like this happening again, I don't really mind where we will catch those kind of issues, here at build time or that we make sure that whenever this zip file changes that we must check how that affects what gets built here. I suppose it is UI that creates this zip in the first place?

OYTIS:
@Danten Yes, Robert even said that the UI packs it on the fly with a JS code.
If we want to be flexible and safe at the same time, we could come up with nested archive, like
* credentials.zip
* * device_credentials.zip
* * * autoprov.p12
* * * autoprov.url
* * * whatever_else_we_want_to_be_on_the_device
* * treehub.json
* * whatever_else_we_need_during_build

PR #122, opened 2017-08-14
Make initramfs size more deterministic

dl9pf:
Set 
IMAGE_ROOTFS_EXTRA_SPACE = "0"
IMAGE_OVERHEAD_FACTOR = "1.0"

b/c we also define ramdisk_size=8192 on the kernel cmdline and thus do not support a larger image.

This makes sure we do not exceed the available default ramdisk size.

It avoids:
[    4.400242] RAMDISK: incomplete write (18440 != 32768)
[    4.405466] write error
[    4.456116] EXT4-fs (ram0): bad geometry: block count 12288 exceeds size of device (8192 blocks)


dl9pf:
IMAGE_ROOTFS_EXTRA_SPACE set twice ... fixing.

PR #124, opened 2017-08-14
Set IMAGE_OVERHEAD_FACTOR for fixed initrd size

dl9pf:
Set
IMAGE_OVERHEAD_FACTOR = "1.0"

Otherwise the initrd was 12M due to the default overhead (x1.3).


dl9pf:
This fixes 
[    4.400242] RAMDISK: incomplete write (18440 != 32768)
[    4.405466] write error
[    4.456116] EXT4-fs (ram0): bad geometry: block count 12288 exceeds size of device (8192 blocks)

cajun-rat:
Merges to Pyro don't (currently) need an ack from Ben, so please merge this when you are happy, Leon.

leon-anavi:
OK, I will do a end to end update for a test with this version and Poky Pyro on rpi3.

cajun-rat:
That would be cool, but a build+boot test should be enough

cajun-rat:
Thank you @dl9pf for the fix

PR #127, opened 2017-08-18
scripts/envsetup.sh: Use machine specific local.conf

leon-anavi:
Use the configurations for local.conf provided by
layer meta-updater-* (depending on the machine)
if variable TEMPLATECONF has not been set.
Otherwise, if the variable is not set or if the
layer does not provide any configurations then
the default will be used (just as before this change).


leon-anavi:
@patrickvacek thank you for the review and thank you for reminding me about the need of this change.

PR #129, opened 2017-08-21
Remove BUILD_TESTS=OFF because it does nothing.

patrickvacek:
Aktualizr tests are no longer built with 'make all' and there is no
longer a variable named BUILD_TESTS so setting it is neither necessary
nor meaningful.


patrickvacek:
Now based against pyro.

cajun-rat:
Just building now...

PR #130, opened 2017-08-23
ostree_git.bb: Depends on xz for liblzma

leon-anavi:
Move Yocto/OE recipe xz from the run-time dependencies
to the dependencies because liblzma is required for
the successful configuration of OSTree.


cajun-rat:
This needs an ack from Ben before merging

EpicNuts:
Successful build, connection and update. merging in.

PR #135, opened 2017-09-05
PRO-3923 Add persistence option to run-qemu script

tkfu:
Didn't realize the run-qemu script was in meta-updater too.


patrickvacek:
@tkfu What is the point of having it in both places? @cajun-rat made the copy, so perhaps he knows if you don't.

tkfu:
@patrickvacek I'm not sure, but @cajun-rat is on vacation until next week. @OYTIS, do you know why this got moved?

PR #136, opened 2017-09-08
Add recipes to test pkcs11 functionality in Aktualizr

OYTIS:
Also introduces SOTA_CLIENT_FEATURES variable which should control client's functionality. E.g. SOTA_CLIENT_FEATURES = "hsm hsm-test" to add HSM support and create a mock HSM token for testing.
HSM support inside aktualizr itself is coming once tested.

### DIFF
--- a/classes/sota.bbclass
+++ b/classes/sota.bbclass
@@ -18,7 +18,7 @@ EXTRA_IMAGEDEPENDS_append_sota = " parted-native mtools-native dosfstools-native
 # Please redefine OSTREE_REPO in order to have a persistent OSTree repo
 OSTREE_REPO ?= "${DEPLOY_DIR_IMAGE}/ostree_repo"
 # For UPTANE operation, OSTREE_BRANCHNAME must start with "${MACHINE}-"
-OSTREE_BRANCHNAME ?= "${MACHINE}-ota"
+OSTREE_BRANCHNAME ?= "${MACHINE}"
 OSTREE_OSNAME ?= "poky"
 OSTREE_INITRAMFS_IMAGE ?= "initramfs-ostree-image"
 
--- a/recipes-support/softhsm-testtoken/files/createtoken.service
+++ b/recipes-support/softhsm-testtoken/files/createtoken.service
@@ -1,0 +1,12 @@
+[Unit]
+Description=Create a mock smartcard for testing
+Before=aktualizr.service
+RequiredBy=aktualizr.service
+
+[Service]
+RestartSec=10
+Restart=on-failure
+ExecStart=/usr/bin/createtoken.sh
+
+[Install]
+WantedBy=aktualizr.service
--- a/recipes-sota/aktualizr/aktualizr_git.bb
+++ b/recipes-sota/aktualizr/aktualizr_git.bb
@@ -7,6 +7,11 @@ LIC_FILES_CHKSUM = "file://${S}/LICENSE;md5=9741c346eef56131163e13b9db1241b3"
 DEPENDS = "boost curl openssl jansson libsodium ostree"
 RDEPENDS_${PN} = "lshw"
 
+DEPENDS_append = "${@bb.utils.contains('SOTA_CLIENT_FEATURES', 'hsm', ' libp11', '', d)}"
+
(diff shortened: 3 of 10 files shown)


recipes-support/softhsm-testtoken/files/createtoken.service
```diff
@@ -0,0 +1,12 @@
+[Unit]
+Description=Create a mock smartcard for testing
+Before=aktualizr.service
+RequiredBy=aktualizr.servie
```
patrickvacek: Typo alert! "aktualizr.service".

OYTIS: Thanks!

patrickvacek:
This seems to continually fail for me:

```
Sep 11 12:15:58 qemux86-64 systemd[1]: Started Create a mock smartcard for testing.
Sep 11 12:15:58 qemux86-64 createtoken.sh[1005]: error: PKCS11 function C_Initialize failed: rv = CKR_GENERAL_ERROR (0x5)
Sep 11 12:15:58 qemux86-64 createtoken.sh[1005]: Aborting.
Sep 11 12:15:58 qemux86-64 createtoken.sh[1005]: ls: /var/sota/token/pkey.pem: No such file or directory
Sep 11 12:15:58 qemux86-64 createtoken.sh[1005]: ls: /var/sota/token/client.pem: No such file or directory
Sep 11 12:15:58 qemux86-64 systemd[1]: [[0;1;39mcreatetoken.service: Main process exited, code=exited, status=1/FAILURE[[0m
Sep 11 12:15:58 qemux86-64 systemd[1]: [[0;1;39mcreatetoken.service: Unit entered failed state.[[0m
Sep 11 12:15:58 qemux86-64 systemd[1]: [[0;1;39mcreatetoken.service: Failed with result 'exit-code'.[[0m
Sep 11 12:16:08 qemux86-64 systemd[1]: createtoken.service: Service hold-off time over, scheduling restart.
Sep 11 12:16:08 qemux86-64 systemd[1]: Stopped Create a mock smartcard for testing.
```

I'm not sure why ```! ls /var/sota/token/pkey.pem /var/sota/token/client.pem``` appears to result in the script aborting if the files don't exist. The return code of an ```ls``` of non-existent files is 1, and !1 should fail the if, right?

OYTIS:
@patrickvacek 
Regarding ls: in bash zero is interpreted as true, non-zero as false.
Regarding the libp11 error, we should take a look at it tomorrow, the script works for me.

patrickvacek:
@OYTIS I was confused about the bash true/false test, because the result of the pkcs11-tool command when it fails is 1, and yet the script appears to continue to the ls test instead of exiting.

classes/sota.bbclass
```diff
@@ -17,7 +17,7 @@ EXTRA_IMAGEDEPENDS_append_sota = " parted-native mtools-native dosfstools-native
 # Please redefine OSTREE_REPO in order to have a persistent OSTree repo
 OSTREE_REPO ?= "${DEPLOY_DIR_IMAGE}/ostree_repo"
 # For UPTANE operation, OSTREE_BRANCHNAME must start with "${MACHINE}-"
-OSTREE_BRANCHNAME ?= "${MACHINE}-ota"
+OSTREE_BRANCHNAME ?= "${MACHINE}"
```
patrickvacek: Perhaps a rather minor matter, but the comment there claims OSTREE_BRANCHNAME should start with "${MACHINE}-", but now the default is just "${MACHINE}" without the dash. Does that matter at all?

OYTIS: The comment is probably obsolete, current CI just uses branch name as hardware ID.

patrickvacek: That sounds about right. I hadn't seen that Uptane restriction enforced anywhere. Is there any need at all for the branch name to include MACHINE at all, then, or is it just a loose convention? If so, we should just remove the comment (not that you need to necessarily do that in this PR).

patrickvacek:
Okay, seems to be working now. I hadn't realized I needed to supply the keys myself in /var/sota/token. Was there more you wanted to add, though?

PR #139, opened 2017-09-11
Split Aktualizr binary and systemd service/configuration

leon-anavi:
Split Aktualizr binary and systemd service and configurations in two separate Yocto/OE recipes:
* aktualizr that provides the binary
* aktualizr-pro that provides the systemd services and configurations for manual or autoprovision of Aktualizr.

Best regards,
Leon

### DIFF
--- a/classes/sota.bbclass
+++ b/classes/sota.bbclass
@@ -6,7 +6,8 @@ python __anonymous() {
 OVERRIDES .= "${@bb.utils.contains('DISTRO_FEATURES', 'sota', ':sota', '', d)}"
 
 SOTA_CLIENT ??= "aktualizr"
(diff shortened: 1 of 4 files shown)


patrickvacek:
This looks good and works for me. My only request would be that we name the new autoprov configuration recipe something like "aktualizr-auto-prov", since this is a precursor to adding a new implicit provisioning configuration recipe that will probably be named something like "aktualizr-implicit-prov". I'm open to other schemes, though.

leon-anavi:
@patrickvacek  thank you for the feedback. I have renamed the recipe to **bitbake rpi-basic-image**.

patrickvacek:
@leon-anavi copy-paste error? :)

patrickvacek:
Hooray, I can merge now!

PR #141, opened 2017-09-13
Fixup hack to get glib-networking-native to build

cajun-rat:


patrickvacek:
I think this already got merged in as part of Anton's PR #136. It's identical as far as I can tell.

cajun-rat:
Good spot

PR #142, opened 2017-09-13
Make it easier to build from branches and run a specific image

cajun-rat:
Is is now easy to build a specific branch by adding the following line to
local.conf

BRANCH_pn-aktualizr = "mybranchname"

Also, the run-qemu-ota script can also take a path to an image file, which is
useful for long-term test images.


EpicNuts:
@cajun-rat, Both of those look super handy! :)

PR #143, opened 2017-10-04
Feat/pro 3799/implicit writer

patrickvacek:
To test this, you will need to use https://github.com/advancedtelematic/aktualizr/pull/362.

You will also need to add something like this to your local.conf:
```
DISTRO_FEATURES_append = " implicit-prov "
BRANCH_pn-aktualizr = "feat/PRO-3799/implicit-writer"
BRANCH_pn-aktualizr-native = "feat/PRO-3799/implicit-writer"
require classes/sota_bleeding.inc
```

You will need to use aktualizr_cert_provider to get the key and certificate on the device (unless you are using p11 or want to do it manually).

### DIFF
--- a/classes/image_types_ostree.bbclass
+++ b/classes/image_types_ostree.bbclass
@@ -4,11 +4,9 @@ inherit image
 
 IMAGE_DEPENDS_ostree = "ostree-native:do_populate_sysroot \
                         openssl-native:do_populate_sysroot \
-                        zip-native:do_populate_sysroot \
                         coreutils-native:do_populate_sysroot \
                         virtual/kernel:do_deploy \
-                        ${OSTREE_INITRAMFS_IMAGE}:do_image_complete \
-                        unzip-native"
+                        ${OSTREE_INITRAMFS_IMAGE}:do_image_complete"
 
 export OSTREE_REPO
 export OSTREE_BRANCHNAME
@@ -19,8 +17,6 @@ RAMDISK_EXT_arm ?= ".ext4.gz.u-boot"
 OSTREE_KERNEL ??= "${KERNEL_IMAGETYPE}"
 
 export SYSTEMD_USED = "${@bb.utils.contains('DISTRO_FEATURES', 'systemd', 'true', '', d)}"
-
-SOTA_IMPLICIT_PROV = "${@bb.utils.contains('DISTRO_FEATURES', 'implicit-prov', 'true', '', d)}"
 
 IMAGE_CMD_ostree () {
     if [ -z "$OSTREE_REPO" ]; then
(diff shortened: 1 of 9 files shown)


OYTIS:
A few points:

- I have a feeling that controlling provisioning type doesn't belong to DISTRO_FEATURES. It can almost be controlled just with SOTA_CLIENT_PROV variable, except...
- Writing provisioning credentials inside image_types_ostree.bbclass had been a wrong decision originally. I think once we have multiple types of provisioning, this should be moved inside individual provisioning recipes and the ostree.bbclass should just copy whatever is installed in /var/sota (or even better in /var/local, we can then rename /var/sota -> /var/local/sota).
- I can't see any reason why we shouldn't have a credentials.zip package with implicit provisioning (just containing autoprov.url and treehub.json in implicit case). We can then get rid of rewriting sota.toml as well.

patrickvacek:
1. You are probably right about DISTRO_FEATURES, since the 'implicit-prov' option is only used for setting SOTA_CLIENT_PROV. I think that could just be set directly and that would be fine.
2. I've wondered about that. I think I agree that provisioning details don't belong in the ostree bbclass. I think we can move the bit about deploying SOTA credentials to the aktualizr-auto-prov recipe, and I think we can remove the hacks with setting SOTA_CLIENT_CERTIFICATE, SOTA_CLIENT_KEY, and SOTA_ROOT_CA. Those were only meant to be shortcuts until we had implicit_writer ready. None of that conflicts with using SOTA_CLIENT_PROV to control which provisioning recipe is used, though.
3. I'm up for debate there, but currently I'm using the presence of provision.provision_path to control whether auto or implicit provisioning is triggered in aktualizr. Since the treehub.json needs to be removed from the zip before copying it to the device anyway, the only thing we'd need the zip file for on the device with implicit provisioning is the server URL, which can be controlled through the config file with implicit_writer anyway. So if it really is worth using the zip for that, we need to think of another way to control when implicit provisioning is triggered.

patrickvacek:
@OYTIS A couple further thoughts:
- Should image_types_ostree.bbclass install anything? Why not leave that up to the recipes?
- What's the advantage of using /var/local/?
- The other issue with having a different sota.toml is that we need the root CA file location to be different than it has been thus far. That can either be set in advance in the sota_implicit_prov.toml as is done now or it could be done by implicit_writer like is done for the server URLs. Which bothers you more: having a separate implicit provisioning sota.toml in Yocto or rewriting the file via implicit_writer?

OYTIS:
> I'm up for debate on whether the credentials.zip is necessary on the device for implicit provisioning

For me the most important advantage of always having credentials.zip is unification. And it's easier to keep track on hardcoded sota.toml file.

> Should image_types_ostree.bbclass install anything? Why not leave that up to the recipes?

ostree.bbclass is responsible for rearrangement of file system to match OSTree expectations (like /bin -> /usr/bin merge etc.). Cleaning the /var directory is a part of this rearrangement, because OSTree ignores whatever is in /var during the deployment process. So some code that treats /var/sota (or /var/local) specially has to live here.

> What's the advantage of using /var/local/?

It will mean that we provide a standard location where our users will be able to store their persistent writable data. AGL people need it occasionally for example. We can still keep /var/local/sota for ourselves.

> The other issue with having a different sota.toml is that we need the root CA file location to be different than it has been thus far.

Sorry, don't quite get it. Why can't we use the same location as before?

patrickvacek:
>> What's the advantage of using /var/local/?
>
> It will mean that we provide a standard location where our users will be able to store their persistent writable data. AGL people need it occasionally for example. We can still keep /var/local/sota for ourselves.

This sort of reorganization seems reasonable, but also seems like it is independent of this ticket. Should we make a new bug/task for that?

>> The other issue with having a different sota.toml is that we need the root CA file location to be different than it has been thus far.
>
> Sorry, don't quite get it. Why can't we use the same location as before?

Since the root CA file needs to be provided via Yocto and must be able to be updated, it can't live in /var as it has thus far. Hence, for implicit provisioning it will now live with the sota.toml in /usr/lib/sota.

OYTIS:
> This sort of reorganization seems reasonable, but also seems like it is independent of this ticket. Should we make a new bug/task for that?

We can create a task for moving /var/sota to /var/local/sota. But what is related to this PR is moving deployment of credentials.zip into individual "provisioning" recipes thus eliminating the need for implicit-prov DISTRO_FEATURE.

> Since the root CA file needs to be provided via Yocto and must be able to be updated, it can't live in /var as it has thus far. Hence, for implicit provisioning it will now live with the sota.toml in /usr/lib/sota.

Yes, that makes sense. I think it's better to have a different sota.toml for each provisioning scenario. We'll still need another one for HSM for example.

patrickvacek:
The new commit moves provisioning to the autoprov recipe and removes the implicit-prov DISTRO_FEATURE. It doesn't address any changes to the sota.toml or zip file for implicit provisioning. Any changes I do to that will require some deeper work, so I want to make sure we have a consensus on the right way forward before I rewrite any of that.

OYTIS:
Great, we just need to make sure that it works, i.e. the files deployed in the recipe really get into the final image.

OYTIS:
@patrickvacek 
bitbake can't find libsodium-native, some *.bbappend is probably missing in the commit.

OYTIS:
@patrickvacek 
OK, sorry, it build with pyro.

OYTIS:
@patrickvacek 
No it doesn't. aktualizr_native:do_install fails with

```
rm: cannot remove '/home/anton/updater-repo-qemu-testimplicit/build/tmp/work/x86_64-linux/aktualizr-native/1.0+gitAUTOINC+1004efa3f8-7/image/home/anton/updater-repo-qemu-testimplicit/build/tmp/work/x86_64-linux/aktualizr-native/1.0+gitAUTOINC+1004efa3f8-7/recipe-sysroot-native/usr/bin/aktualizr_cert_provider': No such file or directory
```

patrickvacek:
@OYTIS Hmm, that's a problem. I could change the 'rm' to 'rm -f' to silence the error, but we should understand why that file isn't there. It should be unless there is a problem with the build process.

OYTIS:
@patrickvacek 
Tested with both automatic and implicit provisioning. One last thing before merging that would be updating SRCREV for aktualizr (after merging the respective PR of course)

patrickvacek:
@OYTIS done.

PR #146, opened 2017-10-12
Support pkcs#11 in aktualizr and add softhsm token for testing

OYTIS:
Credentials triple (client.pem, pkey.pem, root.crt) should be placed in /var/sota/token after the build or on the running image. A convenient way to do it is by using cert_provider:

```
aktualizr_cert_provider -r -c /path/to/credentials -d /path/to/var/sota/token
```

### DIFF
--- a/classes/sota.bbclass
+++ b/classes/sota.bbclass
@@ -11,6 +11,9 @@ IMAGE_INSTALL_append_sota = " ostree os-release ${SOTA_CLIENT} ${SOTA_CLIENT_PRO
 IMAGE_CLASSES += " image_types_ostree image_types_ota"
 IMAGE_FSTYPES += "${@bb.utils.contains('DISTRO_FEATURES', 'sota', 'ostreepush otaimg wic', ' ', d)}"
 
+PACKAGECONFIG_append_pn-curl = "${@bb.utils.contains('SOTA_CLIENT_FEATURES', 'hsm', " ssl", " ", d)}"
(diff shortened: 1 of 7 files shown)


patrickvacek:
How can I tell if it is actually working right? softhsm2-util and pkcs11-tool show that there is a token, but I can't quite tell if it's getting used. Is there something in the log to indicate what's going on?

And are you sure you want to include the provision_path in the sota.toml? Including it bypasses the implicit provisioning logic at the moment, and it doesn't sound like you need it if you're using cert_provider.

Is there a use case for implicit provisioning without using an HSM? Should we be merging the implicit provisioning and HSM paths, or is worth allowing one without the other? (These may be questions for @cajun-rat or anyone else.)

OYTIS:
> And are you sure you want to include the provision_path in the sota.toml? Including it bypasses the implicit provisioning logic at the moment, and it doesn't sound like you need it if you're using cert_provider.

Good spot, actually there is something to be fixed yet.

patrickvacek:
How are you calling `aktualizr-hsm-test-prov.bb`? I don't see that recipe getting pulled in.

I can see the createtoken.sh working and doing its magic, so that part looks good.

patrickvacek:
I've confirmed that everything is working right, although I still recommend removing the credentials from the sota.toml (at least until we decide otherwise). I'm still wondering about merging the implicit and hsm code paths. If we want to go forward with that, then cert_provider should install the output directly into the token subdirectory.

OYTIS:
Thank you. I've updated the PR with current Aktualizr master commit hash and removed path to credentials package.

PR #147, opened 2017-10-16
Fix harmless error from not installing garage_deploy.

patrickvacek:


patrickvacek:
At what point is it worth explicitly specifying what files we want to install on the device and host instead of explicitly deleting those that we don't want?

cajun-rat:
Random question: what about pulling this stuff into a separate package like 'aktualizr-tools' or 'aktualizr-extras'?

patrickvacek:
> Random question: what about pulling this stuff into a separate package like 'aktualizr-tools' or 'aktualizr-extras'?

Wouldn't that mean recompiling aktualizr multiple times for each package?

I'm not particularly bothered by the current layout, just that there isn't an easy way to say "don't bother with x and y".

cajun-rat:
Nah, it it possible to have multiple RPMs from a single package.  The syntax is quite simple (I'll have to big up an example when I have a second), and isn't really any more complex than rm-ing the files we don't need.

cajun-rat:
Tested with an update on qemu against the production server

PR #150, opened 2017-10-25
Feat/remerge recipes

patrickvacek:


cajun-rat:
Weirdly, this works with adding the following in `local.conf`:

```
BRANCH_pn-aktualizr = "master"
```

but not with
```
BRANCH_pn-aktualizr = "tmp2"
```

patrickvacek:
@cajun-rat Annoyingly, the workaround is exactly what I didn't want to have to do. You need to set `BRANCH_pn-aktualizr` and `BRANCH_pn-aktualizr-native`. I'm going to investigate a proper fix since I don't want to have to do that.

patrickvacek:
I cleaned up a different (inconsequential) inconsistency I found, but I cannot get the issue at hand cleanly resolved. I tried several things but can't get around the fact that you ultimately have to set both `BRANCH_pn-aktualizr` and `BRANCH_pn-aktualizr-native`. It's particularly strange since sota_bleeding.inc only has to set `SRCREV_pn-aktualizr` for that to automatically work for both. I don't understand why.

For now, I would suggest accepting this workaround (since it's what we already had to do anyway) and keeping this in the back of our minds in case we ever find a better solution.

PR #152, opened 2017-10-30
Remove IMAGEDATESTAMP to make sstate input deterministic (morty)

OYTIS:
Branch name can be misleading, because the PR doesn't cover the initramfs case, we need to collect more info to fix that.


cajun-rat:
(Needs an ACK from Ben)

EpicNuts:
Looks good, test checklist passed

PR #160, opened 2017-11-03
Bump ostree version to get Shaun's fix.

patrickvacek:
Unnecessary error messages have been silenced by skipping some checks
that weren't necessary in our case.


patrickvacek:
Is that PV acceptable? I'm not sure where the older one came from. This was all I could find that made sense to use.

cajun-rat:
If it matches the version number that OSTree upstream uses for that commit, then I'm happy

patrickvacek:
It's both the output of `ostree --version` and the most obvious indicator from [the release page](https://github.com/ostreedev/ostree/releases/tag/v2017.13).

PR #163, opened 2017-11-06
Remove unused tools from latest aktualizr.

patrickvacek:
Use 'rm -f' so that there are no problems when not using the latest aktualizr.


OYTIS:
Closed for now, might do a major backport to morty in future if necessary.

PR #168, opened 2017-11-08
Fix race condition in run-qemu-ota

OYTIS:
Running qemu while overlay is being created is bad.


EpicNuts:
I'm going to approve this based on @patrickvacek's review, as qemu is still broken on my PC. Fixing it is on my to-do list, but it won't happen this week.

PR #169, opened 2017-11-09
Latest garage-sign with many bug fixes.

patrickvacek:


cajun-rat:
Tested with `oe-selftest -r garage_push`

PR #170, opened 2017-11-10
oe-selftest

leon-anavi:
Add notes about oe-selftest to the README and add a test case for hsm hsm-test.

Best regards,
Leon


leon-anavi:
Hi @patrickvacek, thank you for the positive feedback. Your proposition for extending the test cases with checks of the files in the image is excellent. Thanks, Leon

PR #173, opened 2017-11-13
Add managing targets.json by garage-sign

OYTIS:
Tested on qemux86-64


recipes-sota/garage-sign/garage-sign.bb
```diff
@@ -6,14 +6,14 @@ LICENSE = "CLOSED"
 LIC_FILES_CHKSUM = "file://${S}/docs/LICENSE;md5=3025e77db7bd3f1d616b3ffd11d54c94"
 DEPENDS = ""
 
-PV = "0.2.0-29-gf6f095a"
+PV = "0.2.0-35-g0544c33"
```
patrickvacek: Any reason to not use the brand-new version 37?

OYTIS: None, I just didn't notice that it updated again. We can bump it from time to time.

PR #177, opened 2017-11-16
Don't add bblayers if they were already added.

patrickvacek:
This was introduced by b48fe3ed766250dac8564a0f6bd0b05988e35f5c (back in
August) but we just never noticed or cared enough to look into it until
now.

### DIFF
--- a/scripts/envsetup.sh
+++ b/scripts/envsetup.sh
@@ -32,10 +32,11 @@ if [[ ! -f "${BUILDDIR}/conf/local.conf" ]]; then
   else
     # Use the default configurations or TEMPLATECONF set by the user
     source "$METADIR/poky/oe-init-build-env" "$BUILDDIR"
+    # Fill in the expected layers.
+    echo "METADIR  := \"\${@os.path.abspath('${METADIR}')}\"" >> conf/bblayers.conf
+    cat "${METADIR}/meta-updater/conf/include/bblayers/sota.inc" >> conf/bblayers.conf
+    cat "${METADIR}/meta-updater/conf/include/bblayers/sota_${MACHINE}.inc" >> conf/bblayers.conf
   fi
-  echo "METADIR  := \"\${@os.path.abspath('${METADIR}')}\"" >> conf/bblayers.conf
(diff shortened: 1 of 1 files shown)


OYTIS:
Great that you've found it, it produces annoying errors with bbappends. I have an impression that it's config templates that need to be modified/deleted though. Looks like we have two conflicting ways to add specify layers and it's not really maintainable.

patrickvacek:
> Great that you've found it, it produces annoying errors with bbappends. I have an impression that it's config templates that need to be modified/deleted though. Looks like we have two conflicting ways to specify layers and it's not really maintainable.

I think the `TEMPLATECONF` mechanism that Leon used is good, and I think that's a bit more maintainable. The other mechanism is just a fallback solution that I left in case we have anything that uses it. I'm fine going other way, but it probably makes sense to go with whatever the "standard" is. My guess is that the `TEMPLATECONF` method is preferred, but I don't actually know.

OYTIS:
The [original idea](https://github.com/advancedtelematic/meta-updater/commit/82a3c13a0b0c9979fa0812143d619748aa89c85d) was to separate layers common to all platforms from platform-specific ones and keep them all in a single place. The config templates were not deleted though. Later @leon-anavi [added](https://github.com/advancedtelematic/meta-updater/commit/b48fe3ed766250dac8564a0f6bd0b05988e35f5c) code to use them again.
I like separated snippets more, but if there are reasons in favor of TEMPLATECONF am also OK with them. It's also interesting to know Leon's reasons.

OYTIS:
Maybe the primary reason for intoducing '*.inc's was that local.conf is constructed much easier that way and bblayers.conf generation was changed for the sake of consistency.

OYTIS:
Yes, and then we've rewritten local.conf includes to bbclasses, so only ones for bblayers.conf are left.

OYTIS:
Alternative solution: https://github.com/advancedtelematic/meta-updater/pull/181

patrickvacek:
Closing since Anton's solution seems better.

PR #178, opened 2017-11-16
Autodetect kvm

cajun-rat:


scripts/qemucommand.py
```diff
@@ -67,7 +68,15 @@ def __init__(self, args):
             self.mac_address = random_mac()
         self.serial_port = find_local_port(8990)
         self.ssh_port = find_local_port(2222)
-        self.kvm = not args.no_kvm
+        if args.kvm is None:
+            # Autodetect KVM using 'kvm-ok'
+            try:
+                check_output(['kvm-ok'])
```
patrickvacek: On shovel, this returns true for me, yet launching the oe-selftest qemu_test complains with:
```
Could not access KVM kernel module: Permission denied
failed to initialize KVM: Permission denied
```
And the hostname is actually empty, which to me kind of implies to me that the boot (and thus the test) actually failed. I actually can't launch `run-qemu-ota` on shovel: I get the same errors. Even with `--no-kvm` I get this: `Could not initialize SDL(No available video device) - exiting`.

OYTIS: You need to run ssh with -X flag in order to forward Xserver requests. It also tends to lose connection after a while, reconnecting helps in this case.
To use kvm you need to run runqemu as superuser, maybe it's also worth adding this check to the script.

patrickvacek: > You need to run ssh with -X flag in order to forward Xserver requests.
Ah, of course, let me try that.

> To use kvm you need to run runqemu as superuser
I'm confused; I never run it as superuser, and `-enable-kvm` is always present. Is that not supposed to work?

Also, by the way, I think I've got a fix for the issue with the test succeeding despite the qemu boot failing.

OYTIS: > I'm confused; I never run it as superuser, and -enable-kvm is always present. Is that not supposed to work?

I'm too now. It never worked for me without sudo.

patrickvacek: I seem to recall @cajun-rat telling me about some trick with non-sudo kvm back in my first week here.

patrickvacek: But failing a mechanism to do that trick (if it does really exist), we should perhaps "enhance" kvm-ok to default to disable if it needs root and doesn't have it. Running oe-selftest as root is probably not a good option.

cajun-rat: @patrickvacek  I don't remember that, but "sudo addgroup `whoami` kvm" fixed it on shovel for me.

cajun-rat: (I've added a bunch of people to the kvm group on shovel, so a logout/login should make everything work for you)

patrickvacek: @cajun-rat I was wondering if it was something like that, but I'm not in the kvm group on my box.

patrickvacek: This is working for me on shovel now. This is clearly an improvement as is, but we may still want to think about the groups/superuser issues at some point.

cajun-rat: In the spirit of blaming a someone else for a bug, I claim that `kvm-ok` is the one at fault here :) I would expect it to answer in a way that matches qemu -enable-kvm

PR #181, opened 2017-11-16
Ignore configuration templates to avoid duplicate inclusion of stuff

OYTIS:
Also add IMAGE_ROOTFS_EXTRA_SPACE to qemu configuration which was
the original motivation for returning TEMPLATECONF processing (see
https://github.com/advancedtelematic/meta-updater-qemux86-64/pull/9 )

### DIFF
--- a/classes/sota_qemux86-64.bbclass
+++ b/classes/sota_qemux86-64.bbclass
@@ -9,3 +9,5 @@ UBOOT_MACHINE_sota = "qemu-x86_defconfig"
 OSTREE_BOOTLOADER ?= "u-boot"
 
 OSTREE_KERNEL_ARGS ?= "ramdisk_size=16384 rw rootfstype=ext4 rootwait rootdelay=2 ostree_root=/dev/hda"
+
+IMAGE_ROOTFS_EXTRA_SPACE = "${@bb.utils.contains('DISTRO_FEATURES', 'sota', '65536', '', d)}"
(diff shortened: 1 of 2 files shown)


patrickvacek:
Oddly, if I change it to `IMAGE_ROOTFS_EXTRA_SPACE = "65536"`, it seems to work.

OYTIS:
@patrickvacek Yes, it's a known problem. Overrides have higher priority that normal assignments, and it sometimes has unexpected consequences (like initramfs image bloated in this case).

OYTIS:
OK, this one seems to work.

patrickvacek:
Confirmed, it works fine. One last suggestion: could you make it a `?=` so that it can still be overridden in `local.conf` if desired?

OYTIS:
No, sorry, it doesn't work. The first assignment in poky/meta/conf/bibake.conf (to "0") will then take precedence.

patrickvacek:
Crap, you're right. Well, how do can we override this if necessary, then, other than changing the entry in `classes/sota_qemux86-64.bbclass`?

OYTIS:
Yes, configuration snippets have some advantages over bbclasses, but there were some things that could only be made with bbclasses and I'm not keen on mixing the two. If the user really wants to, he/she can use overrides (_sota or one's own).

OYTIS:
Yes, overrides will not work either, since the user will not want initramfs to bloat. We could add support for a new variable to override the default "65536" if you think it could be useful.

patrickvacek:
That could work! I wouldn't worry about it for now, though, since it doesn't seem to come up much.

PR #183, opened 2017-11-17
Test/oe selftest grub

patrickvacek:


patrickvacek:
Is there anything holding this PR back? It happens to also include a fix for detecting the machine name for the qemu test since if you build for multiple machines the autodetection fails.

cajun-rat:
Nope, leeme try it

cajun-rat:
Tested with oe-selftest

PR #184, opened 2017-11-17
Feat/pro 4252/rocko

OYTIS:
I'll create a PR to updater-repo in order to test rocko/master branches.


conf/include/bblayers/sota.inc
```diff
@@ -1,4 +1,6 @@
-
+BBLAYERS += "${METADIR}/poky/meta"
+BBLAYERS += "${METADIR}/poky/meta-poky"
+BBLAYERS += "${METADIR}/poky/meta-yocto-bsp"
```
patrickvacek: This seems to cause my new bblayers.conf to have duplicated entries for each of these three layers. Do we need this?

OYTIS: Um... no, I was sure I've deleted that.

OYTIS: Just fixed these, thanks. There are some minor fixes coming for RPi3 and maybe Minnowboard, but I'm not sure if it's worth adding them to this PR.

patrickvacek:
I'm stuck on bitbake errors like the following:
```
ERROR: gettext-minimal-native-0.19.8.1-r0 do_patch: Command Error: 'quilt --quiltrc /home/patrick/Code/qemu_rocko/build/tmp/work/x86_64-linux/gettext-minimal-native/0.19.8.1-r0/recipe-sysroot-native/etc/quiltrc push' exited with 0  Output:
/bin/sh: 1: quilt: not found
```
I cannot figure out why it can't find quilt. That's pretty basic.

OYTIS:
Never seen that, looks like the environment is fundamentally broken. I think quilt should be pulled by some bbclasses belonging to oe-core on early stages. Do you use updater-repo to pull everything together?

patrickvacek:
Yeah, I'm pretty confused. This was a new build directory, and yes, I'm using updater-repo.

PR #186, opened 2017-11-21
Check if the package and UPTANE target got to the backend

OYTIS:


classes/image_types_ostree.bbclass
```diff
@@ -227,9 +227,20 @@ IMAGE_CMD_garagesign () {
             bberror "Couldn't push to garage repository"
             exit 1
         fi
-    else
-        bbwarn "SOTA_PACKED_CREDENTIALS not set. Please add SOTA_PACKED_CREDENTIALS."
```
patrickvacek: Is there a use case for running this without providing credentials?

classes/image_types_ostree.bbclass
```diff
@@ -227,9 +227,20 @@ IMAGE_CMD_garagesign () {
...
     fi
 }
 
+IMAGE_TYPEDEP_garagecheck = "ostreepush garagesign"
+IMAGE_DEPENDS_garagecheck = "aktualizr-native:do_populate_sysroot"
+IMAGE_CMD_garagecheck () {
+    if [ -n "${SOTA_PACKED_CREDENTIALS}" ]; then
+        # if credentials are issued by a server that doesn't support offline signing, exit silently
+        unzip -p ${SOTA_PACKED_CREDENTIALS} root.json targets.pub targets.sec repo.url 2>&1 >/dev/null || exit 0
```
cajun-rat: Why wont this work when people are using the old (signed on the server) mode?

cajun-rat: sorry, dumb question--the tuf repo is only visible through the device gateway in the 'old world order'

cajun-rat:
Tested with oe-selftest

PR #187, opened 2017-11-22
Rename repo.url -> tufrepo.url

OYTIS:
Also pass this URL to garage-sign.

Waiting for targets keys format being fixed to test that.

### DIFF
--- a/classes/image_types_ostree.bbclass
+++ b/classes/image_types_ostree.bbclass
@@ -183,7 +183,7 @@ IMAGE_DEPENDS_garagesign = "garage-sign-native:do_populate_sysroot"
 IMAGE_CMD_garagesign () {
     if [ -n "${SOTA_PACKED_CREDENTIALS}" ]; then
         # if credentials are issued by a server that doesn't support offline signing, exit silently
-        unzip -p ${SOTA_PACKED_CREDENTIALS} root.json targets.pub targets.sec repo.url 2>&1 >/dev/null || exit 0
+        unzip -p ${SOTA_PACKED_CREDENTIALS} root.json targets.pub targets.sec tufrepo.url 2>&1 >/dev/null || exit 0
 
         java_version=$( java -version 2>&1 | awk -F '"' '/version/ {print $2}' )
         if [ "${java_version}" = "" ]; then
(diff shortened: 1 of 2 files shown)


classes/image_types_ostree.bbclass
```diff
@@ -198,11 +198,7 @@ IMAGE_CMD_garagesign () {
             garage-sign init --repo ${GARAGE_SIGN_REPO} --home-dir ${GARAGE_SIGN_REPO} --credentials ${SOTA_PACKED_CREDENTIALS}
         fi
 
-        if [ -n "${GARAGE_SIGN_REPOSERVER}" ]; then
-            reposerver_args="--reposerver ${GARAGE_SIGN_REPOSERVER}"
-        else
-            reposerver_args=""
-        fi
+        reposerver_args="--reposerver $( unzip -p ${SOTA_PACKED_CREDENTIALS} tufrepo.url )"
```
patrickvacek: Do you need to cat that URL or can we just use the file as is?

OYTIS: That is the interface as was described by Simao. We can ask @simao or @jochenschneider to process tufrepo.url internally instead, but it doesn't work yet to my understanding.

classes/image_types_ostree.bbclass
```diff
@@ -183,7 +183,7 @@ IMAGE_DEPENDS_garagesign = "garage-sign-native:do_populate_sysroot"
 IMAGE_CMD_garagesign () {
     if [ -n "${SOTA_PACKED_CREDENTIALS}" ]; then
         # if credentials are issued by a server that doesn't support offline signing, exit silently
-        unzip -p ${SOTA_PACKED_CREDENTIALS} root.json targets.pub targets.sec repo.url 2>&1 >/dev/null || exit 0
+        unzip -p ${SOTA_PACKED_CREDENTIALS} root.json targets.pub targets.sec turrepo.url 2>&1 >/dev/null || exit 0
```
jochenschneider: Spelling: "turrepo.url" -> "tufrepo.url"

classes/image_types_ostree.bbclass
patrickvacek: No problem, just checking.

classes/image_types_ostree.bbclass
OYTIS: Thanks!

PR #188, opened 2017-11-24
Bump aktualizr version for good

OYTIS:
Should fix PRO-4260


patrickvacek:
@stevana Does this actually work for you right now? I'm getting a problem with garage-sign. I sent the details to @jochenschneider and I think he's looking into it.

stevana:
Not checked, was going to after merge (as we get the images built automatically by TC then).

PR #190, opened 2017-11-28
Fix builds outside the .repo directory

cajun-rat:
The repo tool searches up the directory tree to find the .repo directory.
Cleanly handle the case where it can't find anything.

### DIFF
--- a/classes/image_repo_manifest.bbclass
+++ b/classes/image_repo_manifest.bbclass
@@ -14,9 +14,9 @@ HOSTTOOLS_NONFATAL += " repo "
 # Write build information to target filesystem
 buildinfo () {
   if [ $(which repo) ]; then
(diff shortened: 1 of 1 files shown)


patrickvacek:
This is pretty less than critical. I added this feature a couple weeks ago just so that when we are debugging a device we can go find the manifest that it was built with. That's all it does. You won't even miss it. You probably didn't even know it was there.

patrickvacek:
On further thought, it is probably quite difficult to find the right place to run repo from. However, we could try setting a variable for that in bblayers.conf or local.conf. It could even get automatically set by running `envsetup.sh`. And otherwise I think we should maybe just leave a comment in the file in question that explains that various failure cases and expected outcomes.

PR #192, opened 2017-11-30
Use *.toml files provided in aktualizr github repo

OYTIS:
This enables backwards incompatible changes to configuration format.

Tested for autoprovisioning case on qemu.


recipes-sota/aktualizr/aktualizr_git.bb
```diff
@@ -40,6 +40,10 @@ do_install_append_class-target () {
 do_install_append_class-native () {
     rm -f ${D}${bindir}/aktualizr
     rm -f ${D}${bindir}/aktualizr-info
+    install -d ${D}${libdir}/sota
+    install -m 0644 ${S}/config/sota_autoprov.toml ${D}/${libdir}/sota/sota_autoprov.toml
+    install -m 0644 ${S}/config/sota_hsm_test.toml ${D}/${libdir}/sota/sota_hsm_test.toml
+    install -m 0644 ${S}/config/sota_implicit_prov.toml ${D}/${libdir}/sota/sota_implicit_prov.toml
```
patrickvacek: This means all of the important configs will be installed on every device, right? I think I actually like that.

recipes-sota/aktualizr/aktualizr-auto-prov.bb
```diff
@@ -40,7 +39,7 @@ do_install_append() {
         install -d ${D}/${systemd_unitdir}/system
         install -m 0644 ${WORKDIR}/aktualizr-autoprovision.service ${D}/${systemd_unitdir}/system/aktualizr.service
         install -d ${D}${libdir}/sota
-        install -m "0644" ${WORKDIR}/sota_autoprov.toml ${D}${libdir}/sota/sota.toml
+        install -m "0644" ${STAGING_DIR_NATIVE}/${libdir}/sota/sota_autoprov.toml ${D}${libdir}/sota/sota.toml
```
cajun-rat: What about creating a symlink to the thing installed by the aktualizr recipe? Adding a dependency on aktualizr-native feels a bit weird

recipes-sota/aktualizr/aktualizr_git.bb
OYTIS: Um... No, that's actually only for native installation. They are later used by provisioning recipes. But we might have this option as well.

recipes-sota/aktualizr/aktualizr-auto-prov.bb
OYTIS: You mean install all of the *.toml files on every device and then symlink to the correct one?

recipes-sota/aktualizr/aktualizr_git.bb
patrickvacek: Ah, now I see. I read too fast and didn't see the context that that was only in the `-native` part.

patrickvacek:
Looks like this needs a rebase before merging. But otherwise seems good to me.

OYTIS:
Thank you, rebased.

PR #193, opened 2017-11-30
Remove aktualizr-manual-provision.service

OYTIS:
We don't support this scenario any more and it makes testing more complicated.


recipes-sota/aktualizr/aktualizr-auto-prov.bb
```diff
@@ -51,7 +50,7 @@ do_install_append() {
       fi
     else
         install -d ${D}/${systemd_unitdir}/system
```
jochenschneider: I guess you could factor these 2 lines out and put them in front of the "if".

PR #195, opened 2017-12-01
Add spaces before "_append" clauses

OYTIS:
Under some conditions missing space can lead to a corrupted environment variable


patrickvacek:
No way, really!? The space at the end of the first `EXTRA_OECMAKE` line wasn't enough? Sorry for that; I think that was my doing.

PR #196, opened 2017-12-04
Provide user interface to add legacy secondary bridge

OYTIS:
This is a bit of a hack, but I can't imagine non-hacky way to do that (well, we can start supporting proper configuration fragments like kernel recipes do). SOTA_LEGACY_SECONDARY_INTERFACE variable will not work until https://github.com/advancedtelematic/aktualizr/pull/425 is merged.


patrickvacek:
@stevana This is actually not too far from what you were asking about earlier!

Where does `SOTA_LEGACY_SECONDARY_INTERFACE` come from? Is that just to be set in local.conf?

patrickvacek:
Also, do want to make a similar interface for `--secondary-config`, or is that not worth it since it's only used for virtual secondaries?

OYTIS:
@patrickvacek Yes, the variable is supposed to be set by user to some valid path on device's file system.

OYTIS:
@patrickvacek It exports a bitbake variable as a shell environment variable inside a recipe.

recipes-sota/aktualizr/environment.inc
```diff
@@ -0,0 +1,17 @@
...
+
+  AKTUALIZR_PARAMETERS_CONFIGFILE="--config /usr/lib/sota/sota.toml"
+  for sec in ${SOTA_VIRTUAL_SECONDARIES}; do
+    AKTUALIZR_PARAMETERS_VIRTUALSECS="${AKTUALIZR_PARAMETERS_VIRTUALSECS} --secondary-config $sec"
+  done
+
+  echo "AKTUALIZR_CMDLINE_PARAMETERS=${AKTUALIZR_PARAMETERS_CONFIGFILE} ${AKTUALIZR_PARAMETERS_LEGACYSEC} ${AKTUALIZR_PARAMETERS_VIRTUALSECS}" > ${D}${libdir}/sota/sota.env
+}
+
+FILES_${PN}_append = " ${libdir}/sota/sota.env"
```
patrickvacek: This needs a mkdir right before it.

patrickvacek: Well, really probably `install -d "${D}/${libdir}/sota"` before the echo.

patrickvacek:
This looks good except for my one minor complaint and the fact that getting this to work in practice is difficult. (The latter is not a reasonable ground to delay merging.)

cajun-rat:
I think this might be breaking oe-selftest.  Does `oe-selftest -r updater` work for you?

patrickvacek:
> Does oe-selftest -r updater work for you?

It works for me.

cajun-rat:
oe-selftest clearly hates me :)

PR #197, opened 2017-12-05
Fix calculate_size in case $SIZE < $MIN

benkard:
calculate_size had a typo that made one of the cases fail.  This patch fixes it.


cajun-rat:
Nice catch, thank you!

PR #198, opened 2017-12-05
Fix garage-sign targets add.

patrickvacek:


cajun-rat:
Running oe-selftest now

PR #204, opened 2017-12-07
Properly install example-interface on target and remove it on host.

patrickvacek:
Not ready yet since we will also want to bump the aktualizr version once https://github.com/advancedtelematic/aktualizr/pull/437 is merged.

### DIFF
--- a/recipes-sota/aktualizr/aktualizr_git.bb
+++ b/recipes-sota/aktualizr/aktualizr_git.bb
@@ -18,7 +18,7 @@ PR = "7"
 SRC_URI = " \
   git://github.com/advancedtelematic/aktualizr;branch=${BRANCH} \
   "
-SRCREV = "57e9cdb8aa1e8ee9e682628bd67031d9be7aaafa"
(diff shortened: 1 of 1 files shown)


OYTIS:
@patrickvacek 
I would let user enable/disable installation of example interface with some SOTA_CLIENT_FEATURE.

OYTIS:
@patrickvacek And the version of aktualizr also needs to be bumped.

patrickvacek:
> I would let user enable/disable installation of example interface with some SOTA_CLIENT_FEATURE.

I was wondering about that. I took the easy route for now but that seems reasonable.

OYTIS:
Let's merge it before the demo time and then add the options. I have a local branch for that on my machine, just don't want to do all the testing right now.

patrickvacek:
Cool, give me just a minute to update this...

patrickvacek:
@OYTIS Look ready to you?

patrickvacek:
Rebased and updated with the proper fix for only installing example-interface if explicitly requested.

PR #212, opened 2017-12-13
Fixes for Spekulatius

OYTIS:
- New garage-sign interface
- Remove garage-sign recipe (now installed with aktualizr-native)
- Small but critical bugfixes in aktualizr


lib/oeqa/selftest/updater.py
```diff
@@ -31,23 +31,13 @@ def test_deploy_help(self):
...
-
-    @classmethod
-    def setUpClass(cls):
-        logger = logging.getLogger("selftest")
-        logger.info('Running bitbake to build garage-sign-native')
-        bitbake('garage-sign-native')
-
-    def test_help(self):
+    def test_garagesign_help(self):
         bb_vars = get_bb_vars(['SYSROOT_DESTDIR', 'bindir'], 'garage-sign-native')
```
patrickvacek: Looks like we missed this one: instead of `garage-sign-native` this should be `aktualizr-native`.

recipes-sota/garage-sign/garage-sign.bb
```diff
@@ -1,33 +0,0 @@
...
-    install -d ${D}${bindir}
-    install -m "0755" -t ${D}${bindir} ${S}/bin/*
-    install -d ${D}${libdir}
-    install -m "0644" -t ${D}${libdir} ${S}/lib/*
-}
-
-FILES_${PN} = " \
-  ${bindir}/garage-sign.bat \
-  ${bindir}/garage-sign \
-  ${libdir}/* \
```
patrickvacek: Does the installation of garage-deploy in the aktualizr recipe somehow automatically install garage-sign? I'm concerned that we aren't explicitly installing it anywhere, but it magically appears anyway.

classes/image_types_ostree.bbclass
```diff
@@ -194,23 +194,20 @@ IMAGE_CMD_garagesign () {
             exit 1
         fi
 
-	if [ ! -d "${GARAGE_SIGN_REPO}" ]; then
-            garage-sign init --repo ${GARAGE_SIGN_REPO} --home-dir ${GARAGE_SIGN_REPO} --credentials ${SOTA_PACKED_CREDENTIALS}
-        fi
-
-        reposerver_args="--reposerver $( unzip -p ${SOTA_PACKED_CREDENTIALS} tufrepo.url )"
+        rm -rf ${GARAGE_SIGN_REPO}
+        garage-sign init --repo ${GARAGE_SIGN_REPO} --home-dir ${GARAGE_SIGN_REPO} --credentials ${SOTA_PACKED_CREDENTIALS} --no-auth
```
patrickvacek: I think using `--no-auth` is causing garage-sign to fail for me with `The supplied authentication is not authorized to access this resource`.

patrickvacek: Now I understand that `--no-auth` is necessary for Spekulatius but I think that breaks everything else.

recipes-sota/garage-sign/garage-sign.bb
OYTIS: Strangely enough it worked for me, but we need to add it of course.

patrickvacek: Yeah it worked for me too, but I don't understand why yet. I tried removing my /usr/bin/garage-sign and it wasn't even using that.

OYTIS:
@patrickvacek 
The intermediate result of our discussions about root.ca as far as I understood it was the we're not going to update root certificate until we have a common understanding about how to do it properly. Therefore, there is no need to treat root.ca specially for now, so I've just set `--no-root` (you meant `--no-root`, right?) unconditionally.

patrickvacek:
@OYTIS I thought there was a `--no-auth` flag for `garage-sign init`, but now I'm not seeing it. What happened to that?

OYTIS:
@patrickvacek 
Ah, OK. There's no plan, garage-sign now just looks into the credentials (namely authentication objects in treehub.json) and acts accordingly.

patrickvacek:
Cool, sounds good to me.

PR #217, opened 2017-12-22
Add support of ISO/TP legacy secondaries and serial CAN

OYTIS:
Current implementation is quite hacky, created a ticket in Jira to make it (and similar tasks) better.

https://advancedtelematic.atlassian.net/browse/PRO-4572


patrickvacek:
I probably broke this with https://github.com/advancedtelematic/aktualizr/pull/470, but that was necessary for Vladimir and probably a good idea anyway. All we need to do is add BUILD_ISOTP if necessary.

patrickvacek:
Sorry, I probably should have fixed that myself. I can confirm that it builds, so if you still want to go this route (as opposed to going more down the route suggested in the Jira ticket), this is fine by me.

PR #220, opened 2018-01-05
Bugfix/pro 4632/morty pseudo bugs

patrickvacek:
Backport some pyro fixes to morty to fix issues around pseudo.

First, sqlite needs to be built with -fPIC on some systems to make linking it with pseudo-native work.

Second, in some installations, pseudo-native would fail because of a previously backported patch getting doubly applied. The problem was some cleverness in envsetup.sh that accidentally created duplicate layers.

### DIFF
--- a/recipes-support/sqlite/sqlite3_3.14.1.bbappend
+++ b/recipes-support/sqlite/sqlite3_3.14.1.bbappend
@@ -1,0 +1,1 @@
+CFLAGS_append = " -fPIC"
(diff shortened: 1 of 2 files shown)


EpicNuts:
Successfully built and pushed a new morty image using this branch. Closing ticket.

patrickvacek:
@EpicNuts Don't you have to be the one to merge? Or can @cajun-rat ? In any case, I cannot.

PR #222, opened 2018-01-08
Delete dependency on libsoup

OYTIS:
It makes supporting different platforms (latest Archlinux in my case) somewhat easier.


OYTIS:
Cool, thanks.

PR #228, opened 2018-01-16
Get the newest aktualizr using sqlstorage

OYTIS:


patrickvacek:
BTW I have an aktualizr branch in progress that I would like to merge in before this PR is merged. (So presumably this should get updated with the new aktualizr branch once it's ready.) It fixes a typo and a minor bug, but since it changes the schemas I'd like to get that out of the way to avoid complications.

patrickvacek:
Updated with latest aktualizr version. I'm happy with this now.

PR #229, opened 2018-01-16
Deploy startup.nsh before composing flashable image

OYTIS:
Additionally don't create extra images to save space and not confuse the user


OYTIS:
Depends on https://github.com/advancedtelematic/meta-updater-minnowboard/pull/3. Some integration work may be needed for AGL, haven't looked into that yet.

cajun-rat:
Assigning to @EpicNuts to test on hardware

PR #233, opened 2018-01-18
Fix some basic oe-selftest errors.

patrickvacek:
Grub, HSM, and qemu hostname tests still fail for reasons I haven't figured out yet.

### DIFF
--- a/lib/oeqa/selftest/cases/updater.py
+++ b/lib/oeqa/selftest/cases/updater.py
@@ -13,6 +13,7 @@ class SotaToolsTests(OESelftestTestCase):
 
     @classmethod
     def setUpClass(cls):
+        super(SotaToolsTests, cls).setUpClass()
         logger = logging.getLogger("selftest")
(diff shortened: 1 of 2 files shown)


lbonn:
I also got an issue with the `kvm-ok` check and have a fix for that. It makes sense to bundle it here IMO.

lbonn:
Actually, I didn't find `kvm-ok` on any Debian package, so maybe we'd like to have an alternative way to check for that.

patrickvacek:
@lbonn No complaint from me. It's probably reasonable that if kvm-ok fails for any reason, we should just assume it shouldn't be used. That said, kvm ought to be an option for your Debian machine, so we should try to find a way to check for that.

patrickvacek:
Ha! Thinking the same thing at the same time.

cajun-rat:
On my Ubuntu machine kvm-ok is part of the cpu-checker package:

    # dpkg -S /usr/sbin/kvm-ok
    cpu-checker: /usr/sbin/kvm-ok

PR #236, opened 2018-01-19
Deploy initial primary version information

OYTIS:
A bit hacky method, but OSTree refhash is only known to image_types_ostree and image_types_ota bbclasses, so I couldn't figure out a better one.
I couldn't test it on staging yet, where it has most sense, but on prod it seems to send a reasonable manifest.
Should work with SQLite as well, because of how FS->SQL migration works (and yes, it's also a hack).
Making a PR to pyro, we'll have to merge a couple of things again when we make the final transition to rocko.

### DIFF
--- a/classes/image_types_ostree.bbclass
+++ b/classes/image_types_ostree.bbclass
@@ -12,6 +12,8 @@ IMAGE_DEPENDS_ostree = "ostree-native:do_populate_sysroot \
 export OSTREE_REPO
 export OSTREE_BRANCHNAME
 
+export GARAGE_TARGET_NAME
+
 RAMDISK_EXT ?= ".ext4.gz"
 RAMDISK_EXT_arm ?= ".ext4.gz.u-boot"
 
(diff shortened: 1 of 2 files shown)


patrickvacek:
I'm not sure I understand what's going on here and what this is fixing.

OYTIS:
@patrickvacek 
In order to build image version manifest for primary ECU aktualizr should know the connection between UPTANE package name and OSTree refhash (we assume that UPTANE package version and OSTree refhash is the same thing). It is done by writing `<package_name> <refhash>` pair into installed_versions file (or respective table in SQLite storage) upon successful installation.
But initial deployment after yocto build is done without aktualizr, so the build process should take care of delivering this information to it. If it is not done, aktualizr reports its first version as `unknown-<refhash>`. Until recently there was a hack in the Garage UI that will show the installed version just based on its hash, so the user didn't care that the package name is in reality `unknown`. But this hack caused some troubles with testing secondaries, so we decided to remove it and display installed version honestly. Therefore it is now necessary to deploy UPTANE package name to a newly created flashable image.

patrickvacek:
@OYTIS Okay, cool. Thanks for the detailed explanation! Makes a lot more sense, but then where does `GARAGE_TARGET_NAME` actually get set?

OYTIS:
@patrickvacek `sota.bblass` sets it to `OSTREE_BRANCHNAME` by default and it can also be overridden independently of `OSTREE_BRANCHNAME` in local.conf/site.conf

PR #241, opened 2018-01-24
Compile aktualizr with asn1c help from yocto

lbonn:
As you'll see, this required a little hack.

Required for https://github.com/advancedtelematic/aktualizr/pull/509 (which is not ready yet)

### DIFF
--- a/recipes-sota/asn1c/asn1c.bb
+++ b/recipes-sota/asn1c/asn1c.bb
@@ -8,7 +8,8 @@ LIC_FILES_CHKSUM = "file://LICENSE;md5=ee8bfaaa7d71cf3edb079475e6716d4b"
 inherit autotools native
 
 PV = "0.9.28"
-SRC_URI = "https://github.com/vlm/asn1c/releases/download/v${PV}/asn1c-${PV}.tar.gz"
+SRC_URI = "https://github.com/vlm/asn1c/releases/download/v${PV}/asn1c-${PV}.tar.gz \
+           file://skeletons_dir_fix.patch"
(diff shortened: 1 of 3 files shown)


recipes-sota/asn1c/asn1c.bb
```diff
@@ -0,0 +1,18 @@
+SUMMARY = "ASN.1 to C compiler"
+DESCRIPTION = "Generates serialization routines from ASN.1 schemas"
+HOMEPAGE = "http://lionet.info/asn1c"
+SECTION = "base"
+LICENSE = "BSD"
+LIC_FILES_CHKSUM = "file://LICENSE;md5=ee8bfaaa7d71cf3edb079475e6716d4b"
+DEPENDS = "asn1c-native"
```
patrickvacek: This seems like a recursive dependency.

recipes-sota/asn1c/files/skeletons_dir_fix.patch
```diff
@@ -0,0 +1,32 @@
...
++        if(stat(skeletons_dir, &sb)) {
+             /*
+              * The default skeletons directory does not exist,
+              * compute it from my file name:
+-             * ./asn1c/asn1c -> ./skeletons
++             * ./asn1c/asn1c -> ./share/asn1c
+              */
+             char *p;
+             size_t len;
++            printf("HI\n");
```
patrickvacek: Is that printf intentional?

lbonn: obviously not, thanks!

cajun-rat:
I know the asn1c question is still being investigated, but I think we should merge this, and if we end up using protobufs or something else then we can rip it out again.

PR #246, opened 2018-01-31
Magic incantation to make ExternalProject work with yocto

OYTIS:


patrickvacek:
Ah, sorry, my PR just gets rid of the need for the external project, but if this works in the meantime, then cool.

patrickvacek:
I finally fixed the problem in my PR, so as soon as that passes CI, hopefully we can just go with that if you'd prefer.

OYTIS:
@patrickvacek Thank you, I'll close this one.

PR #248, opened 2018-02-01
Merge our current progress from Rocko to master

OYTIS:


cajun-rat:
My assumption is that meta-updater master should aim to work with Yocto master (i.e. whatever will be Rocko+1)

PR #249, opened 2018-02-02
Latest aktualizr with ostree ca cert fix.

patrickvacek:


EpicNuts:
Tested with Qemu, RaspPi and Minnowboard. 
Successful device activation, provisioning and update.

PR #250, opened 2018-02-02
Bugfix/rocko update ostree

patrickvacek:
Mostly just for the ostree ca cert bugfix and latest garage-sign.


cajun-rat:
Tested with oe-selftest and a manual update

PR #254, opened 2018-02-09
Replace INITRAMFS_FSTYPES with OSTREE_INITRAMFS_FSTYPES

rsalveti:
User is able to customize the init script via OSTREE_INITRAMFS_IMAGE but
there is no way to set INITRAMFS_FSTYPES as it gets defined by
sota.bbclass. Create a new variable called OSTREE_INITRAMFS_IMAGE to
handle the INITRAMFS_FSTYPES update, and also allow the user to override
it.


rsalveti:
Can't just change INITRAMFS_FSTYPES to "?=" as it gets set to cpio.gz by default, breaking the default sota logic.

leon-anavi:
Looks good to me but my attempt to build a Poky image from master with this patch for Raspberry Pi 3 failed. The error doesn't seem directly related to this GitHub PR as it affects the generation of a wic file.

PR #260, opened 2018-02-20
Test/pro 4481/hsm

patrickvacek:
This greatly expands the HSM testing of oe-selftest and improves a bunch of other tests as well.

PR #261, opened 2018-02-21
oe-selftest for implicit provisioning

lbonn:
This PR is based on https://github.com/advancedtelematic/meta-updater/pull/260

### DIFF
--- a/lib/oeqa/selftest/updater.py
+++ b/lib/oeqa/selftest/updater.py
@@ -345,7 +345,7 @@ class HsmTests(oeSelfTest):
         bb_vars = get_bb_vars(['SOTA_PACKED_CREDENTIALS'], 'aktualizr-native')
         creds = bb_vars['SOTA_PACKED_CREDENTIALS']
         bb_vars_prov = get_bb_vars(['STAGING_DIR_NATIVE', 'libdir'], 'aktualizr-hsm-prov')
-        config = bb_vars_prov['STAGING_DIR_NATIVE'] + bb_vars_prov['libdir'] + '/sota/sota_implicit_prov.toml'
+        config = bb_vars_prov['STAGING_DIR_NATIVE'] + bb_vars_prov['libdir'] + '/sota/sota_hsm_prov.toml'
 
         akt_native_run(self, 'aktualizr_cert_provider -c {creds} -t root@localhost -p {port} -r -s -g {config}'
                        .format(creds=creds, port=self.qemu.ssh_port, config=config))
--- a/lib/oeqa/selftest/updater.py
+++ b/lib/oeqa/selftest/updater.py
@@ -217,14 +217,9 @@ class GrubTests(oeSelfTest):
 
 class ImplProvTests(oeSelfTest):
 
-    @classmethod
-    def setUpClass(cls):
-        logger = logging.getLogger("selftest")
-        logger.info('Running bitbake to build aktualizr-native tools')
-        bitbake('aktualizr-native')
-
     def setUpLocal(self):
         self.append_config('SOTA_CLIENT_PROV = " aktualizr-implicit-prov "')
+        # note: this will build aktualizr-native as a side-effect
         self.qemu, self.s = qemu_launch(machine='qemux86-64')
 
     def tearDownLocal(self):
@@ -294,15 +289,10 @@ class ImplProvTests(oeSelfTest):
 
 class HsmTests(oeSelfTest):
 
-    @classmethod
-    def setUpClass(cls):
-        logger = logging.getLogger("selftest")
-        logger.info('Running bitbake to build aktualizr-native tools')
-        bitbake('aktualizr-native')
-
     def setUpLocal(self):
         self.append_config('SOTA_CLIENT_PROV = "aktualizr-hsm-prov"')
         self.append_config('SOTA_CLIENT_FEATURES = "hsm"')
+        # note: this will build aktualizr-native as a side-effect
         self.qemu, self.s = qemu_launch(machine='qemux86-64')
 
     def tearDownLocal(self):
--- a/lib/oeqa/selftest/updater.py
+++ b/lib/oeqa/selftest/updater.py
@@ -16,32 +16,17 @@ class SotaToolsTests(oeSelfTest):
     @classmethod
     def setUpClass(cls):
         logger = logging.getLogger("selftest")
-        bb_vars = get_bb_vars(['SYSROOT_DESTDIR', 'base_prefix', 'libdir', 'bindir'],
-                              'aktualizr-native')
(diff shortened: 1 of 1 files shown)


lib/oeqa/selftest/updater.py
```diff
@@ -339,11 +296,9 @@ class HsmTests(oeSelfTest):
...
     @classmethod
     def setUpClass(cls):
-        bb_vars = get_bb_vars(['SYSROOT_DESTDIR', 'base_prefix', 'libdir', 'bindir'],
-                              'aktualizr-native')
-        cls.sysroot = bb_vars['SYSROOT_DESTDIR'] + bb_vars['base_prefix']
-        cls.sysrootbin = bb_vars['SYSROOT_DESTDIR'] + bb_vars['bindir']
-        cls.libdir = bb_vars['libdir']
+        logger = logging.getLogger("selftest")
+        logger.info('Running bitbake to build aktualizr-native tools')
+        bitbake('aktualizr-native')
```
patrickvacek: Is this extra bitbake necessary here? Running `qemu_launch` calls `bitbake core-image-minimal`, and that should already include `aktualizr-native`.

lib/oeqa/selftest/updater.py
```diff
@@ -509,4 +454,27 @@ def qemu_send_command(port, command):
     stdout, stderr = s2.communicate()
     return stdout, stderr, s2.returncode
 
+
+def akt_native_run(testInst, cmd, **kwargs):
```
patrickvacek: I think I'm a fan of writing utility functions like this. I hadn't looked into it too much yet but seems like it should make life easier.

lib/oeqa/selftest/updater.py
lbonn: Oh I see, that's why you removed this line in your commit. Thanks!

patrickvacek:
@lbonn Is the HSM test working for you? It isn't for me and I think the fix is https://github.com/advancedtelematic/aktualizr/pull/603. Once that is merged, we'll need to bump the aktualizr version again.

lbonn:
Oh I actually have a problem indeed...
This must have got broken by https://github.com/advancedtelematic/aktualizr/pull/601

patrickvacek:
> This must have got broken by advancedtelematic/aktualizr#601

Yeah, no problem, it's easy enough to fix.

lib/oeqa/selftest/updater.py
```diff
@@ -174,19 +187,219 @@ def tearDownLocal(self):
...
+        softhsm2_command = 'softhsm2-util --show-slots'
+        stdout, stderr, retcode = self.qemu_command(softhsm2_command)
+        self.assertNotEqual(retcode, 0, 'softhsm2-tool succeeded before initialization: ' +
+                        stdout.decode() + stderr.decode())
+
+        # Run cert_provider.
+        bb_vars = get_bb_vars(['SOTA_PACKED_CREDENTIALS'], 'aktualizr-native')
+        creds = bb_vars['SOTA_PACKED_CREDENTIALS']
+        bb_vars_prov = get_bb_vars(['STAGING_DIR_NATIVE', 'libdir'], 'aktualizr-hsm-prov')
+        config = bb_vars_prov['STAGING_DIR_NATIVE'] + bb_vars_prov['libdir'] + '/sota/sota_implicit_prov.toml'
```
patrickvacek: This should be `/sota/sota_hsm_prov.toml`.

lib/oeqa/selftest/updater.py
```diff
@@ -174,19 +187,219 @@ def tearDownLocal(self):
...
+
+        akt_native_run(self, 'aktualizr_cert_provider -c {creds} -t root@localhost -p {port} -r -s -g {config}'
```
patrickvacek: I think each of these `aktualizr_cert_provider` calls (and presumably any other sota/aktualizr tool) need the complete path to the binary. I used to have it as something like `p = bb_vars['SYSROOT_DESTDIR'] + bb_vars['bindir'] + "/aktualizr_cert_provider"`. Otherwise the test may use whatever is in the host's path, which for me is nothing.

lib/oeqa/selftest/updater.py
lbonn: This is also present in your branch actually, I'll reference that there :)

lib/oeqa/selftest/updater.py
lbonn: Actually this should be handled by `native_sysroot` argument to oeqa's `runCmd` which adds this dir in front of `PATH`. But yes, it's probably more robust if `akt_native_run` does the translation itself to avoid any ambiguity.

patrickvacek: Ah, fair enough. Up to you what the best option is. I got thrown off when I was debugging.

lib/oeqa/selftest/updater.py
patrickvacek: Crap, you're right. I guess they were close enough that it didn't matter until the `cert_provider` change.

patrickvacek: Actually, for the sake of not making you rebase, does it make sense for you to just make the change here?

lbonn: Sure, will do

patrickvacek:
After upgrading the aktualizr version to `SRCREV = "75c20855a8023063203ff4479a413d12637f562d"`, I get this:
```
======================================================================
FAIL: test_provisioning (oeqa.selftest.updater.HsmTests)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/patrick/Code/updater_qemu_pyro/meta-updater/lib/oeqa/selftest/updater.py", line 363, in test_provisioning
    p11_out.decode() + hsm_err.decode() + hsm_out.decode())
AssertionError: False is not true : pkcs11-tool or softhsm2-tool failed: error: PKCS11 function C_Initialize failed: rv = CKR_GENERAL_ERROR (0x5)

Aborting.
ERROR: Could not initialize the PKCS#11 library/module: /usr/lib/softhsm/libsofthsm2.so
```

lbonn:
@patrickvacek bisected to https://github.com/advancedtelematic/aktualizr/commit/45e62ad7107d5e9571c3a7476cde076692c81c45

Oops that's also me...

cajun-rat:
(Can you also rebase this now that #260 is merged, please?)

lbonn:
Yes, this will need a bump of aktualizr after https://github.com/advancedtelematic/aktualizr/pull/607 is merged to fix HSM tests (for patrick's problem, I don't know about yours yet, Phil)

patrickvacek:
Works for me after "bitbake -c cleanall aktualizr aktualizr-native aktualizr-implicit-prov aktualizr-hsm-prov", then an explicit "bitbake aktualizr aktualizr-native aktualizr-implicit-prov aktualizr-hsm-prov", then running the tests again.

PR #262, opened 2018-02-21
Use FIT image on RPi

OYTIS:
To be tested together with https://github.com/advancedtelematic/meta-updater-raspberrypi/pull/17
Kernel parameters are still non-updateable, for which I can imagine two possible solutions:

1. Build device tree overlay containing command line options as a bitbake artifact. The problem with this approach is that bitbake's infrastructure is created to build all the device trees from the kernel source tree, so every kernel will need a patch adding our overlay before build.

2. Add kernel command line to OSTree pretty much the same way as kernel, initramfs and dtb work today.

There is another problem with u-boot passing extra parameters to kernel via /chosen/bootargs in the device tree overwriting whatever was there before. That may be not the behaviour expected by the user, who wants them to be kept. It may make sense to add something like bootargs_extra to u-boot.

### DIFF
--- a/classes/sota_raspberrypi.bbclass
+++ b/classes/sota_raspberrypi.bbclass
@@ -1,11 +1,20 @@
 RPI_USE_U_BOOT_sota = "1"
-KERNEL_IMAGETYPE_sota = "uImage"
+
+KERNEL_CLASSES_append_sota = " kernel-fitimage"
+KERNEL_IMAGETYPE_sota = "fitImage"
+
 PREFERRED_PROVIDER_virtual/bootloader_sota ?= "u-boot"
-UBOOT_MACHINE_raspberrypi2_sota ?= "rpi_2_defconfig"
-UBOOT_MACHINE_raspberrypi3_sota ?= "rpi_3_32b_defconfig"
+UBOOT_ENTRYPOINT_sota ?= "0x00008000"
 
 IMAGE_FSTYPES_remove_sota = "rpi-sdimg"
 OSTREE_BOOTLOADER ?= "u-boot"
 
 # OSTree puts its own boot.scr to bcm2835-bootfiles
-IMAGE_BOOT_FILES_remove_sota += "boot.scr"
+IMAGE_BOOT_FILES_sota = "bcm2835-bootfiles/* u-boot.bin;${SDIMG_KERNELIMAGE}"
+
+# Just the overlays that will be used should be listed
+KERNEL_DEVICETREE_raspberrypi2_sota ?= " bcm2709-rpi-2-b.dtb "
+KERNEL_DEVICETREE_raspberrypi3_sota ?= " bcm2710-rpi-3-b.dtb overlays/vc4-kms-v3d.dtbo overlays/rpi-ft5406.dtbo"
+
+# Kernel args normally provided by RPi's internal bootloader. Non-updateable
+OSTREE_KERNEL_ARGS_sota ?= " 8250.nr_uarts=1 bcm2708_fb.fbwidth=720 bcm2708_fb.fbheight=480 bcm2708_fb.fbswap=1 vc_mem.mem_base=0x3ec00000 vc_mem.mem_size=0x40000000 dwc_otg.lpm_enable=0 console=ttyS0,115200 usbhid.mousepoll=0 "


classes/sota_raspberrypi.bbclass
```diff
@@ -1,11 +1,22 @@
 RPI_USE_U_BOOT_sota = "1"
-KERNEL_IMAGETYPE_sota = "uImage"
+
+KERNEL_CLASSES_append_sota = " kernel-fitimage"
+KERNEL_IMAGETYPE_sota = "fitImage"
+
 PREFERRED_PROVIDER_virtual/bootloader_sota ?= "u-boot"
 UBOOT_MACHINE_raspberrypi2_sota ?= "rpi_2_defconfig"
-UBOOT_MACHINE_raspberrypi3_sota ?= "rpi_3_32b_defconfig"
+UBOOT_MACHIE_raspberrypi3_sota ?= "rpi_3_32b_defconfig"
```
patrickvacek: Typo? Should be `UBOOT_MACHINE...`, right?

OYTIS: Hm... I wonder how it worked for me.

OYTIS: Actually these assignments are redundant, it's already done in machine configuration.

cajun-rat:
@leon-anavi Can you check this out, please?

classes/sota_raspberrypi.bbclass
```diff
@@ -1,11 +1,20 @@
...
+
 PREFERRED_PROVIDER_virtual/bootloader_sota ?= "u-boot"
-UBOOT_MACHINE_raspberrypi2_sota ?= "rpi_2_defconfig"
-UBOOT_MACHINE_raspberrypi3_sota ?= "rpi_3_32b_defconfig"
+UBOOT_ENTRYPOINT_sota ?= "0x00008000"
 
 IMAGE_FSTYPES_remove_sota = "rpi-sdimg"
 OSTREE_BOOTLOADER ?= "u-boot"
 
 # OSTree puts its own boot.scr to bcm2835-bootfiles
```
rsalveti: Please update this comment as well, since boot.scr is not removed anymore.

classes/sota_raspberrypi.bbclass
```diff
@@ -1,11 +1,20 @@
 RPI_USE_U_BOOT_sota = "1"
-KERNEL_IMAGETYPE_sota = "uImage"
+
+KERNEL_CLASSES_append_sota = " kernel-fitimage"
+KERNEL_IMAGETYPE_sota = "fitImage"
```
rsalveti: This will force fitImage for all rpi targets, and it will also require the extra patch that is available at meta-updater-raspberrypi (which seems to be changing only rpi_3_32b_defconfig atm).

I'm curious if we should move most of this logic into that layer instead, as it is a bit weird to have some hw support here and the rest in another layer, and you can't really use one without the other.

OYTIS: @rsalveti Historically meta-updater was developed in AGL codebase and only later we decided to create a stand-alone environment with updater-repo. When we did it it turned out that AGL's platform-specific layers contained some useful paches that we needed for standalone use case as well. But as meta-updater should still have been compatible with AGL, they couldn't have been added to meta-updater itself, hence meta-updater-<platform> layers.
And still we want to control most of platform-specific logic inside meta-updater itself, because every change to meta-updater-<platform> has to be mirrored to meta-agl-bsp/<platform> and I'd like to minimise this effort.
Looks ugly, I agree, but I couldn't come up with a better solution yet.

classes/sota_raspberrypi.bbclass
OYTIS: @rsalveti Yes, probably, but I see some problems here
1. Aktualizr doesn't currently support updating the bootloader, and I don't know how soon it will.
2. Even if it supported updating the boot partition, the synchronization problem would arise. If I updated my kernel/dtb with a new version where support for some parameters was dropped, while others have been added, how do I make sure that the new kernel is always used with the new bootloader? It gets even more complicated once aktualizr supports rollbacks.

So my idea (might be a wrong one of course) was that user doesn't update bootloader to enable new functionality, it is all done with kernel/filesystem updates (which are synchronized by OSTree). Bootloader update when it is supported should only be used to fix bugs/vulnerabilities in bootloader itself.

classes/sota_raspberrypi.bbclass
OYTIS: @rsalveti Your note about support for raspberrypi 2 is totally legit, I think we should either support it properly or explicitly only support RPi3

classes/sota_raspberrypi.bbclass
rsalveti: Yeah, looking again it seems that the only really safe way to handle this with OSTree is by using fit format for devices requiring device tree.

It will make it incompatible with the traditional config.txt rpi file (e.g. when adding a new device tree overlay), but it is indeed a lot better to let the system handle what is supported or not.

classes/sota_raspberrypi.bbclass
rsalveti: > Looks ugly, I agree, but I couldn't come up with a better solution yet.

Got it, thanks for explaining the background.

> I think we should either support it properly or explicitly only support RPi3

I would be fine supporting just rpi3, as it is now the default rpi target anyway.

cajun-rat:
Tested with ostree deploy, server problems blocked an end-to-end test.

PR #263, opened 2018-02-22
Bugfix/fix libp11 openssl again

patrickvacek:


cajun-rat:
Fixes oe-selftest -u updater.  This bug was caused by a change in upstream rocko

PR #272, opened 2018-03-07
image_types_ota: make default grub.cfg a link to loader/grub.cfg

rsalveti:
/boot/grub2/grub.cfg should reflect the grub.cfg used by the boot process
instead of being an empty file.

uEnv.txt should have a similar logic, but will get to that once I'm sure it will not cause any regression.


cajun-rat:
@OYTIS Does this bug exist in rocko too?

PR #273, opened 2018-03-08
image_types_ota.bbclass: sync deploy /var with the content from rootfs/var

rsalveti:
As /var gets bind-mounted from deploy /var, the original content from
rootfs/var needs to be duplicated when creating the final ota image,
otherwise the final distro behavior will not be aligned with what gets
produced in the original rootfs.

### DIFF
--- a/classes/image_types_ota.bbclass
+++ b/classes/image_types_ota.bbclass
@@ -96,11 +96,12 @@ IMAGE_CMD_otaimg () {
 
 		ostree admin --sysroot=${PHYS_SYSROOT} deploy ${kargs_list} --os=${OSTREE_OSNAME} ${ostree_target_hash}
 
-		# Copy deployment /home and /var/sota to sysroot
+		# Copy deployment /home and /var to sysroot
 		HOME_TMP=`mktemp -d ${WORKDIR}/home-tmp-XXXXX`
-		tar --xattrs --xattrs-include='*' -C ${HOME_TMP} -xf ${DEPLOY_DIR_IMAGE}/${IMAGE_LINK_NAME}.rootfs.ostree.tar.bz2 ./usr/homedirs ./var/sota ./var/local || true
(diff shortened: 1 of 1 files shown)


cajun-rat:
Do you have a package that is creating stuff in /var?  OSTree doesn't update files in /var (by design) so applications are responsible for maintaining version compatibility of any data that is in there.

If an application is installed via a software update push, then there is no way for /var to get automatically populated correctly.  My worry is that if we copy the contents of /var over, then applications will start depending on this behavior, and will fail when installed for the first time via an OTA update.  This is also great for flushing out bugs where applications assume a certain directory exists in /var.  I prefer to use systemd tmpfiles to create these things (or have the application do it)

rsalveti:
My problem in particular is related to the directory structure that the distro expects to have in /var instead of the files itself.

One example of the top level var dir:
`
backups
cache
lib
local
lock
log -> volatile/log
run
spool
tmp -> volatile/tmp
volatile
`

And the entire list of directories:
`
.
./run
./volatile
./volatile/log
./volatile/log/journal
./volatile/tmp
./cache
./cache/ldconfig
./cache/ldconfig/aux-cache
./cache/opkg
./backups
./lib
./lib/sudo
./lib/sudo/lectured
./lib/arpd
./lib/systemd
./lib/alsa
./lib/alsa/asound.state
./lib/misc
./lib/dbus
./lib/NetworkManager
./lib/dhcp
./log
./local
./tmp
./spool
./spool/mail
./lock
`

It's not common for distros to manage files in var since the standard structure already defines that the packages shouldn't rely on /var over time, but the initial set of folders are expected to be available (e.g.  dbus would fail without /var/lib/dbus).

systemd tmpfiles should be fine as well (need to try that myself), so I guess if we want to protect the user by not using this patch we should at least warn in readme that meta-updater expects the user to align the distro var directories in a systemd tmpfiles config in order for everything to work.

rsalveti:
Let me give this a try with systemd tmpfiles and will propose another pull request to update the documentation. Will close the pull request for now then, thanks for the review!

PR #275, opened 2018-03-08
Move sota.env from /usr/lib to /etc

rsalveti:
The current way for the user to replace the default aktualizr
environment configuration is by copying the original aktualizr.service
to /etc/systemd/system and changing EnvironmentFile to point to a
different file, which is not user friendly.

Change the default environment path from /usr/lib/sota/sota.env to
/etc/sota/sota.env to allow user changes during runtime.

### DIFF
--- a/recipes-sota/aktualizr/environment.inc
+++ b/recipes-sota/aktualizr/environment.inc
@@ -11,7 +11,8 @@ do_install_append() {
     AKTUALIZR_PARAMETERS_VIRTUALSECS="${AKTUALIZR_PARAMETERS_VIRTUALSECS} --secondary-config $sec"
   done
 
-  echo "AKTUALIZR_CMDLINE_PARAMETERS=${AKTUALIZR_PARAMETERS_CONFIGFILE} ${AKTUALIZR_PARAMETERS_LEGACYSEC} ${AKTUALIZR_PARAMETERS_VIRTUALSECS}" > ${D}${libdir}/sota/sota.env
+  install -d ${D}${sysconfdir}/sota
+  echo "AKTUALIZR_CMDLINE_PARAMETERS=${AKTUALIZR_PARAMETERS_CONFIGFILE} ${AKTUALIZR_PARAMETERS_LEGACYSEC} ${AKTUALIZR_PARAMETERS_VIRTUALSECS}" > ${D}${sysconfdir}/sota/sota.env
 }
 
-FILES_${PN}_append = " ${libdir}/sota/sota.env"
+FILES_${PN}_append = " ${sysconfdir}/sota/sota.env"
--- a/recipes-sota/aktualizr/files/aktualizr-serialcan.service
+++ b/recipes-sota/aktualizr/files/aktualizr-serialcan.service
@@ -8,7 +8,7 @@ Requires=network-online.target
 [Service]
 RestartSec=10
 Restart=always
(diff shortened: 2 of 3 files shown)


cajun-rat:
I'm a bit worried by this, because it conflicts with our (unstated) goal of moving toward a Stateless system (in the systemd sense of the word).

The intention is that users can put their own settings in /etc, and use those to override our default settings in /usr.  In an ideal world we would have a clean split between:

* /etc For local administrator changes (this would often be completely empty)
* /var For data files generated by applications (erasing this would be a 'factory reset')
* /usr Is read-only application data, managed by the software update system

I think your use-case is reasonable, and we should support it.  Would an alternative be to copy aktualizr.service to /etc, then add modify the config directly using Environment=FOO=bar or by editing the command line directly?

Can you describe a bit more about what you are aiming to change?

rsalveti:
> I'm a bit worried by this, because it conflicts with our (unstated) goal of moving toward a Stateless system (in the systemd sense of the word).

That's fine, good to know there is also a push to support stateless systems, as then I can also be aligned with that from our side.

> I think your use-case is reasonable, and we should support it. Would an alternative be to copy aktualizr.service to /etc, then add modify the config directly using Environment=FOO=bar or by editing the command line directly?

My only issue with duplicating the service file itself is that it can also easily create update issues over time (in case another part of that file gets changed in a newer rev). I guess one possible solution could be having a default environment as we have today, but change the service file to also consume an environment file in /etc if available, something like:

```
[Service]
RestartSec=10
Restart=always
EnvironmentFile=/usr/lib/sota/sota.env
EnvironmentFile=-/etc/sota/sota.env
ExecStart=/usr/bin/aktualizr $AKTUALIZR_CMDLINE_PARAMETERS
```

That would make aktualizr work after a factory reset and would also allow the user to change the startup options if required/wanted.

Coming back to the use case itself, our main use case is just playing with aktualizr log level in order to facilitate debugging (but other options could also be set by the user if required/wanted).

Would this work for you?

cajun-rat:
That is a great solution! I wasn't aware of the '-' prefix on EnvironmentFile, but it is exactly what we want.

I've been stopping the aktualizr service and running it from the console to up the log level, so this will help me too.

If you are debugging the client/server communication, then I recently wrote up some notes for how to dump the ssl session keys so that wireshark can decrypt the traffic: https://github.com/advancedtelematic/aktualizr/blob/master/docs/debugging-tips.adoc#dumping-ssl-encrypted-traffic  These are a bit sparse at the moment, because they are partly there to remind me how to do it.

Thank you!

patrickvacek:
Is this sufficiently fixed by https://github.com/advancedtelematic/meta-updater/pull/283 to be satisfactorily resolved?

rsalveti:
> Is this sufficiently fixed by #283 to be satisfactorily resolved?

@patrickvacek yes, thanks for the fix and sorry for the delay, travels after ELC.

rsalveti:
> If you are debugging the client/server communication, then I recently wrote up some notes for how to dump the ssl session keys so that wireshark can decrypt the traffic: https://github.com/advancedtelematic/aktualizr/blob/master/docs/debugging-tips.adoc#dumping-ssl-encrypted-traffic These are a bit sparse at the moment, because they are partly there to remind me how to do it.

Awesome, thanks for the notes!

PR #277, opened 2018-03-20
run-qemu-ota support for secondaries

cajun-rat:
Resolves: PRO-5060
Resolves: PRO-5077


lbonn:
Tested with qemu, the dhcp setup works.
Just get `'vlan' is deprecated. Please use 'netdev' instead.` from qemu but I suppose we can ignore that for now.

cajun-rat:
You and your crazy-new Arch Linux :)

patrickvacek:
@cajun-rat I actually get the same deprecation warning on Ubuntu 17.10.

PR #279, opened 2018-03-27
Improve Secondary support

cajun-rat:


patrickvacek:
Does this need to work for other provisioning besides automatic? oe-selftest looks like it works, although as I mentioned I recommend merging the two Secondary tests since they each spawn a qemu instance. When I try to run everything myself, though, I get `bind: Address already in use` on the primary. What am I doing wrong now?

PR #280, opened 2018-03-28
Add aktualizr-check-discovery to shipped files

lbonn:


lbonn:
Actually, I was a bit too fast, it needs a commit bump as well...

patrickvacek:
Oh, right, I have that in my local.conf right now. I was the one that was too fast...

PR #286, opened 2018-04-02
image_types_ostree.bbclass: clean up GARAGE_SIGN_REPO after push

rsalveti:
Avoid exposing the sign repo after the build is completed.


patrickvacek:
I think this makes sense, at least for now. Originally, we wanted to keep that repo around and reuse it, but since we ran into problems with that when switching between servers and such, we remove it before using it. I suppose there isn't much reason to keep it if we aren't going to use it further. Any objection from @OYTIS?

PR #287, opened 2018-04-02
aktualizr: include default configs as part of host-tools

rsalveti:
Provide the default sota configs as part of host-tools so they can be used by aktualizr_cert_provider and aktualizr_implicit_writer.

This pull request depends on #284.

### DIFF
--- a/recipes-sota/aktualizr/aktualizr_git.bb
+++ b/recipes-sota/aktualizr/aktualizr_git.bb
@@ -45,6 +45,11 @@ do_install_append () {
     rm -fr ${D}${libdir}/systemd
     rm -f ${D}${libdir}/sota/sota.toml # Only needed for the Debian package
     install -d ${D}${libdir}/sota
+    install -m 0644 ${S}/config/sota_autoprov.toml ${D}/${libdir}/sota/sota_autoprov.toml
+    install -m 0644 ${S}/config/sota_autoprov_primary.toml ${D}/${libdir}/sota/sota_autoprov_primary.toml
+    install -m 0644 ${S}/config/sota_hsm_prov.toml ${D}/${libdir}/sota/sota_hsm_prov.toml
+    install -m 0644 ${S}/config/sota_implicit_prov.toml ${D}/${libdir}/sota/sota_implicit_prov.toml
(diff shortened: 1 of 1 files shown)


rsalveti:
Rebased pull request on top of latest master.

rsalveti:
There is no harm in producing packages for everything that gets created/built (usually a good practice), as the user can later decide what to install when creating the image.

Having everything that gets used in the build process as packages is useful when debugging, as the packages can be exported in an external repository, for example.

I initially had a similar change that moved the configs and host tools to a package called aktualizr-extras, but since we now have a host-tools package, I decided to just have the configs as part of that instead.

patrickvacek:
> There is no harm in producing packages for everything that gets created/built (usually a good practice), as the user can later decide what to install when creating the image.
> 
> Having everything that gets used in the build process as packages is useful when debugging, as the packages can be exported in an external repository, for example.

Fair points, that makes sense. From my perspective, I don't see why we would want the config files in non-native packages, though. I think adding them to `FILES` for the host-tools package is the right thing to do, but if we do that, then we probably need to update each of the `-prov.bb` recipes to depend on `aktualizr-host-tools-native` instead of `aktualizr-native`.

rsalveti:
> From my perspective, I don't see why we would want the config files in non-native packages, though. 

I think as long the core and extra packages are separating the files correctly, there should be no need to separate files between target and native, as that just complicates the packaging even more (SDK would contain different features and files for the same package name).

> I think adding them to FILES for the host-tools package is the right thing to do, but if we do that, then we probably need to update each of the -prov.bb recipes to depend on aktualizr-host-tools-native instead of aktualizr-native.

Sure, that makes sense and would also be more aligned with the recipes since only the host-tools are used by them.

rsalveti:
>> I think adding them to FILES for the host-tools package is the right thing to do, but if we do that, then we probably need to update each of the -prov.bb recipes to depend on aktualizr-host-tools-native instead of aktualizr-native.

> Sure, that makes sense and would also be more aligned with the recipes since only the host-tools are used by them.

Looking at this again, I don't think that would be needed because aktualizr-native is not the package itself, but the recipe (for native), which will result in every installed file to be available in the sysroot.

So current patch should already cover all the use cases, unless I'm missing something.

PR #289, opened 2018-04-03
Move dev settings to main README. Add table of contents.

patrickvacek:


patrickvacek:
FYI @jerrytrieu: I've cleaned up the docs about the config settings we were talking about earlier to control aktualizr branches/revisions. I'm moving it into the main README here. It used to be in the contributions file, which I always forget even exists.

patrickvacek:
@tkfu Thanks for the tips! I knew the had to be a way, but I thought for some reason it didn't work (hence why aktualizr also has a custom TOC). Glad to be proven wrong. The one thing I'm missing now is that the TOC header is ugly. Do you know how to make it look like a normal section heading without making itself show up redudantly in the TOC?

tkfu:
Oh, nice! I wasn't exactly sure what would happen if it was blank; I thought it might just keep the default.

PR #290, opened 2018-04-06
Stop forcing Linux 4.4 on qemux86-64

lbonn:
The mentioned issue doesn't seem to exist anymore

### DIFF
--- a/classes/sota_qemux86-64.bbclass
+++ b/classes/sota_qemux86-64.bbclass
@@ -1,6 +1,3 @@
-# See https://advancedtelematic.atlassian.net/browse/PRO-2693
-PREFERRED_VERSION_linux-yocto_qemux86-64_sota = "4.4%"
-
 IMAGE_FSTYPES_remove = "wic"
 
 # U-Boot support for SOTA


lbonn:
Precision: I ran `oe-selftest -r updater`and only had failures on the rpi tests (should probably check those anyway).

patrickvacek:
> I ran oe-selftest -r updaterand only had failures on the rpi tests

What were the errors? It's possible that the test is not robust, but it's also possible that the linux-4.4 issue was because of rpi in the first place.

lbonn:
I'll relaunch a test but since the change is in `sota_qemux86-64.bbclass`, it shouldn't have an influence.

Here is the commit that introduces the version freeze: https://github.com/advancedtelematic/meta-updater/commit/a42822fe5d4351a64c4a589756f7b72c81b8f2ef

patrickvacek:
> Everything passed this time

Glad to hear it. Have you tested with an rpi? It worked for me as well with qemu, but I'd like to make sure that someone has tested booting an rpi with this.

lbonn:
@patrickvacek you really think it's necessary, even though the only change is in `sota_qemux86-64.bbclass`? We're compiling 4.9.xx kernels from meta-raspberrypi, just like before.

PR #291, opened 2018-04-06
Feat/garage sign auto version

patrickvacek:
Set GARAGE_SIGN_VERSION in local.conf to request a specific version, e.g.:
GARAGE_SIGN_VERSION = "cli-0.3.0-5-g5908997.tgz"
If unset, get the server URL out of the credentials, ask what version the server is using, and use that. If the credentials don't have tufrepo.url, the default (latest version) will be used.

Set GARAGE_SIGN_SHA256 in local.conf to specify a hash to check, e.g.:
GARAGE_SIGN_SHA256 = "94f5db1accbbbaa279f7ac5a2e618f5448adeafbe32d6cbc90c8604054ab653e"
If unset, the sha256 is not computed.

Please do not merge this until https://github.com/advancedtelematic/aktualizr/pull/684 is merged and this is updated with the correct aktualizr commit. I've created the PR prematurely in order to solicit review.

### DIFF
--- a/recipes-sota/aktualizr/aktualizr_git.bb
+++ b/recipes-sota/aktualizr/aktualizr_git.bb
@@ -22,7 +22,7 @@ SRC_URI = " \
   file://aktualizr-secondary.socket \
   file://aktualizr-serialcan.service \
   "
-SRCREV = "930d8eef6eb584686654601c056d7c9c6fca3048"
+SRCREV = "7ccfc5b4286b9a04915e74a7474a8d3451145e1c"
 BRANCH ?= "master"
 
 S = "${WORKDIR}/git"
@@ -36,6 +36,8 @@ SYSTEMD_SERVICE_${PN} = "aktualizr.service"
 SYSTEMD_SERVICE_${PN}-secondary = "aktualizr-secondary.socket"
 
 BBCLASSEXTEND =+ "native"
+
+require garage-sign-version.inc
 
 EXTRA_OECMAKE = "-DWARNING_AS_ERROR=OFF \
                  -DCMAKE_BUILD_TYPE=Release \
--- a/recipes-sota/aktualizr/aktualizr_git.bb
+++ b/recipes-sota/aktualizr/aktualizr_git.bb
@@ -37,9 +37,17 @@ SYSTEMD_SERVICE_${PN}-secondary = "aktualizr-secondary.socket"
 
 BBCLASSEXTEND =+ "native"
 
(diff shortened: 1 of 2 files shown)


recipes-sota/aktualizr/garage-sign-version.inc
```diff
@@ -0,0 +1,23 @@
...
+        return
+    import json
+    import urllib.request
+    import zipfile
+    with zipfile.ZipFile(d.getVar("SOTA_PACKED_CREDENTIALS", True), 'r') as zip_ref:
+        try:
+            with zip_ref.open('tufrepo.url', mode='r') as url_file:
+                url = url_file.read().decode() + '/health/version'
+        except (KeyError, ValueError, RuntimeError):
+            return
```
lbonn: Should we emit warnings in case of failures (both returns)?

recipes-sota/aktualizr/aktualizr_git.bb
```diff
@@ -22,8 +22,8 @@ SRC_URI = " \
   file://aktualizr-secondary.socket \
   file://aktualizr-serialcan.service \
   "
-SRCREV = "930d8eef6eb584686654601c056d7c9c6fca3048"
-BRANCH ?= "master"
+SRCREV = "2b223de0ea262ae1a057a4eeaddd2314e3329c7a"
+BRANCH ?= "feat/garage-sign-auto-version"
```
lbonn: We probably don't want to merge the branch change.

recipes-sota/aktualizr/garage-sign-version.inc
patrickvacek: The various failure cases are:
* GARAGE_SIGN_VERSION is set. Not an error.
* SOTA_PACKED_CREDENTIALS is not set. Not an error.
* credentials.zip doesn't include tufrepo.url. Not an error.
* Something gets screwed up opening/reading the zip. This will presumably also break the other recipes that use the zip.
* The server returns something other than a 200. Presumably that means the server is broken and other things will break too.

Is it worth reporting a warning for those last two cases?

recipes-sota/aktualizr/aktualizr_git.bb
patrickvacek: Agreed, still waiting on https://github.com/advancedtelematic/aktualizr/pull/684 to get merged first.

patrickvacek:
Rebased and updated to update the aktualizr commit to the latest version of the branch, although that still shouldn't get merged until https://github.com/advancedtelematic/aktualizr/pull/684 is merged and this is updated again.

patrickvacek:
Amended now that the aktualizr commits have been merged. This is ready for review now.

cajun-rat:
Works on My Machine™

PR #295, opened 2018-04-12
Retry garage-push in case of errors on the server side

rsalveti:
As done with garage-sign, make sure to retry push at least 3 times in
case of errors on the server side.


rsalveti:
This is not strictly required, but since we had network issues affecting our CI builds, I decided to have a similar retry logic in place as done with garage-sign.

patrickvacek:
After some consideration, I think it would be best to put this sort of retry logic directly into garage-push. That gives us better control over what really happens and makes testing much easier as well.

PR #296, opened 2018-04-13
Feat/backport from master

patrickvacek:
Pull in some recent changes to keep master closer to rocko.


cajun-rat:
(Rebased)

cajun-rat:
I've dropped the commit that removed the Uptane repo between build steps, because that is dangerous from a security point of view.

rsalveti:
> I've dropped the commit that removed the Uptane repo between build steps, because that is dangerous from a security point of view.

Isn't it more dangerous to let the uptane repo in the deploy folder?

We noticed this because we publish all the files that are available in the deploy/machine folder (output of the build process), and the update repo ended up being published as a consequence.

rsalveti:
Maybe change from deploy to staging instead?

patrickvacek:
Unrelated to the rest of the conversation: I suspect that there are other things in master that should get backported, but I didn't know what all was worth keeping. Once we make a sumo branch, it should presumably have everything from master and rocko.

cajun-rat:
The idea was to improve the security by persisting the Uptane repository state. Since the basic operation of garage-sign is to fetch the existing repository from the server; add the new image; sign it and re-upload it, there is clearly a risk here that the tool might put it's signature on some malicious data provided by the server.

Thinking about this a bit more I'm not convinced persisting this data actually adds anything to the security: the keys are fixed anyway because they come from credentials.zip, and targets.json is going to be fetched and validated against the targets key in both cases.

I think for now I'll add that commit back into this PR, and we can consider changing garage-sign to not require any state for the 'add to targets.json' operation.

cc @heartsucker @simao

cajun-rat:
Passed oe-selftest -r updater

PR #300, opened 2018-04-19
Check for custom target version for installed_versions

rsalveti:
Make sure to use the custom target version when creating
installed_versions if defined by the user.


OYTIS:
@rsalveti 
If I understand the use case correctly, changes also need to be made to image_types_ostree.bbclass so that garage-sign sets version as GARAGE_TARGET_VERSION. User can then control both what the device reports and what version is sent to OTA backend. Did I understand your intent correctly?

rsalveti:
> If I understand the use case correctly, changes also need to be made to image_types_ostree.bbclass so that garage-sign sets version as GARAGE_TARGET_VERSION. User can then control both what the device reports and what version is sent to OTA backend. Did I understand your intent?

That was already done as part of commit 1c48a48f0a0134e53854b4b32d558eca807df766, this patch is just to make sure the same logic also gets applied when creating installed_versions.

OYTIS:
@rsalveti Oh, I missed that, sorry. Makes sense then.

PR #304, opened 2018-04-27
Provide example aktualizr configuration recipes and documentation

patrickvacek:
I have not actually entirely removed sota.env. It still exists to handle the things that can still only be done on the command line (i.e. --config and --secondary-config for virtual secondaries).

Note that this currently doesn't work until aktualizr properly handles config directories. You can at least test that the config gets put in the right place, though.

Test with `IMAGE_INSTALL_append =  " aktualizr-examples aktualizr-example-interface aktualizr-disable-send-ip aktualizr-log-debug"`.


recipes-sota/config/disable-send-ip.bb
```diff
@@ -0,0 +1,23 @@
+SUMMARY = "Disable IP reporting in Aktualizr"
```
lbonn: Should these recipes be prefixed with `aktualizr-`? Their names are not really helpful in isolation.

patrickvacek: I went back and forth on that. That probably makes sense.

patrickvacek: Done.

patrickvacek:
I have updated this PR to now include all of the work relating to PRO-5308, including Serhiy's work to get the config working in systemd style. Please review since it would be nice to get this in before the end of the sprint.

recipes-sota/aktualizr/aktualizr-hsm-prov.bb
```diff
@@ -18,15 +18,15 @@ require environment.inc
 require credentials.inc
 
 do_install() {
-    install -d ${D}${libdir}/sota
+    install -m 0700 -d ${D}${libdir}/sota
```
lbonn: `sota/conf.d`?

patrickvacek: Good call, fixed.

PR #305, opened 2018-04-27
Fix /var/sota perms

cajun-rat:


patrickvacek:
It turns out there are a couple tests where the sleep is skipped! It's worth double-checking and fixing them all.

PR #311, opened 2018-05-15
Backport from master to rocko

lbonn:
Some of the changes we made recently in the way to select which initramfs format to use broke the qemu build, there is an additional fix for that which should eventually go back to master (should we merge rocko to master?)


patrickvacek:
When I do `git log ats/rocko..ats/master`, I'm seeing several additional commits that we should consider, such as b15a3031fe23af2fe4d5dd3d51f20f90a41f33e7, 5ce293daf207c6a01b12108c8cbb239794b160b7, ae156b865e274a5bb4bc9df92a6a88cf5afb1a03, 569e04e956bbff925ff5ebc47050913a82a810d8, and a8449cde4b81fa822e3474b88c2ee9bef0067d99. Presumably we should skip 24cdee96ef2cc3f55daf4c50e97360fc05b03de8. I'm also seeing problems with oe-selftest seeming to related to provisioning failing, although I haven't yet determined if that's something I'm doing wrong.

classes/image_types_ostree.bbclass
```diff
@@ -11,12 +11,12 @@ export OSTREE_REPO
 export OSTREE_BRANCHNAME
 export GARAGE_TARGET_NAME
 
-RAMDISK_EXT ?= ".ext4.gz"
-RAMDISK_EXT_arm ?= ".ext4.gz.u-boot"
+RAMDISK_EXT ?= ".${OSTREE_INITRAMFS_FSTYPES}"
+export GARAGE_TARGET_NAME
```
patrickvacek: This line (`export GARAGE_TARGET_NAME`) is duplicated now.

lbonn:
@patrickvacek yes you're right, I missed a bunch of commits. Probably did not fetch properly...

patrickvacek:
All of the failing tests I see have this:
```
Provisioned on server: yes
Fetched metadata: no
```
This was happening to me before and I'm not really sure what the problem is. It might just be taking its time to fetch the metadata, so maybe we need to lengthen the timers, although I already did that once recently, and perhaps it wasn't enough.

lbonn:
@patrickvacek I also got some failure but after relaunching the tests a couple of times, everything passed...

PR #317, opened 2018-05-22
Include fix for qemu in sota_qemux86-64.bbclass

lbonn:


patrickvacek:
Is this a qemu-only problem? I wish we knew if this was a persistent problem that others encounter. This has always felt like a bit of a hack.

lbonn:
@patrickvacek while investigating, I found this https://lists.yoctoproject.org/pipermail/yocto/2017-November/038876.html, so I concluded that this is a problem when `BUILD_ROM=y`, which our other platforms seem to not use (I can double check though).

We can check again with @OYTIS

lbonn:
Updated so that I won't crash if `x86_64-linux-gnu-gcc` is not present

PR #318, opened 2018-05-23
ostree: set CLEANBROKEN as upstream build system fails to clean

gizero:
Build fails when bitbake expects to call upstream clean target. Setting
CLEANBROKEN in this situations is the suggested approach to workaround build
failures i.e. when changing SRCREVs is involved.

This PR is against master branch but was tested also against pyro and should apply cleanly to rocko as well. Don't know what merging policies this repo is using and couldn't find a contributor's guide, so feel free to merge wherever applicable or tell me to prepare PR for other branches.

### DIFF
--- a/recipes-sota/ostree/ostree_git.bb
+++ b/recipes-sota/ostree/ostree_git.bb
@@ -15,6 +15,8 @@ PR = "1"
 
 S = "${WORKDIR}/git"
 
+CLEANBROKEN = "1"
+
 BBCLASSEXTEND = "native"
 
 DEPENDS += "attr libarchive glib-2.0 pkgconfig gpgme libgsystem fuse e2fsprogs gtk-doc-native curl xz"


patrickvacek:
@gizero At the moment we don't have a strict contribution policy, so making a PR against master is probably the best option. Sorry for any confusion there. We are developing on rocko for now, but hopefully moving to sumo and/or master very soon.

As to the PR itself: what about OSTree was broken that you needed to use `CLEANBROKEN`? I do recall having a problem with bumping the OSTree version at one point back when we were developing on pyro. I looked around the recipe and realized that we are using `inherit autotools-brokensep`, which I'm not sure is necessary. If we got rid of that, would that solve your problem?

patrickvacek:
@gizero Does https://github.com/advancedtelematic/meta-updater/pull/319 address the problem?

gizero:
@patrickvacek In my experience `autotools-brokensep` is to be used when the upstream build system does not support out-of-tree builds, while `CLEANBROKEN` is specifically working around broken `make clean` targets (that recent oe-core `autotools` class does automatically call). What I experienced (and this patch resolved) is that switching between different SRCREVs (I'm testing some newer revisions of the upstream ostree sources) does break the build and requires running `bitbake -c cleansstate ostree` to bring things back to normal operation. I'll give #319 a test anyway, but I suspect it's focusing on a different aspect.

patrickvacek:
@gizero My hope was that the problem of `make clean` not working well with ostree would be resolved by being able to do an out-of-tree build. Upon closer inspection, it does look like `make clean` does not actually clean up everything, although it gets close. The hack could be just to delete the build directory, but that seems like it might be more of a hack than using `CLEANBROKEN`. I don't quite get the significance of `CLEANBROKEN`, though: if it is set, then `make clean` is *not* run before configuration. How does that fix any problems?

gizero:
@patrickvacek Looks like #319 is an overall good improvement. With it in place, I can't reproduce the original issue of the broken clean target. I did a couple of SRCREV switching back and forth with and without cleaning the sstate and all looks good! I'm closing this one while upvoting #319.

patrickvacek:
@gizero Great, glad to hear it! If you see the problem again, please just get in touch again.

PR #319, opened 2018-05-23
ostree: Don't use autotools-brokensep.

patrickvacek:
The only reason we needed it was because of a relative path that I've now made absolute.

This may fix the problem at the root of https://github.com/advancedtelematic/meta-updater/pull/318.


gizero:
LGTM

patrickvacek:
This is apparently causing problems with ostree-native on shovel. `CLEANBROKEN` doesn't help.

patrickvacek:
Nevermind, the problem I was seeing seems to come from not properly cleaning old cruft before running the new tests. So if you see problems about ostree-native complaining about needing to run `make distclean` in the source directory, run something like `bitbake -c cleanall ostree ostree-native && find tmp/ -name ostree -type d -exec rm -R {} +` and then try again.

PR #321, opened 2018-05-25
Feat/find packages

patrickvacek:
This is just a rough draft for now, but if you have suggestions, I'm open to them.

### DIFF
--- a/scripts/find_packages.py
+++ b/scripts/find_packages.py
@@ -1,0 +1,87 @@
+#!/usr/bin/env python3
+
+import os.path
+import sys
+
+scripts_path = os.path.dirname(os.path.realpath(__file__))
+lib_path = os.path.abspath(scripts_path + '/../../poky/bitbake/lib')
+sys.path = sys.path + [lib_path]
+
+import bb.tinfoil
+
+
+def print_deps(tinfoil, abcd_file, rn):
+    try:
+        info = tinfoil.get_recipe_info(rn)
+    except:
+        # fails on hostperl-runtime-native, virtual/libintl-native, grep-native, virtual/libiconv-native
+        print('failing on: %s' % rn)
+        return []
+    append_files = tinfoil.get_file_appends(info.fn)
+    appends = True
+    data = tinfoil.parse_recipe_file(info.fn, appends, append_files)
+    src_uri = data.getVar('SRC_URI').split()
+    lic = data.getVar('LICENSE')
+    summary = data.getVar('SUMMARY')
+    description = data.getVar('DESCRIPTION')
+    homepage = data.getVar('HOMEPAGE')
+    srcrev = data.getVar('SRCREV')
+    branch = data.getVar('BRANCH')
+    depends = data.getVar('DEPENDS').split()
+
+    abcd_file.write('- id:\n')
+    abcd_file.write('    package_manager: "Yocto"\n')
+    abcd_file.write('    name: "%s"\n' % info.pn)
+    abcd_file.write('    version: "%s"\n' % info.pv)
+    abcd_file.write('  declared_lics:\n')
+    abcd_file.write('  - "%s"\n' % lic)
+    if summary:
+        abcd_file.write('  description: "%s"\n' % summary)
+    else:
+        abcd_file.write('  description: "%s"\n' % description)
+    abcd_file.write('  homepage_url: "%s"\n' % homepage)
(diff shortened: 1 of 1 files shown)


scripts/find_packages.py
```diff
@@ -0,0 +1,121 @@
...
+                            'aktualizr-log-debug',
+                            'libp11', # BUILD_P11 (HSM) only
+                            'dpkg', # BUILD_DEB only
+                            'systemd'] # BUILD_SYSTEMD only
+
+        for recipe in recipes_to_check:
+            depends = print_deps(tinfoil, abcd_file, recipe)
+            for dep in depends:
+                if dep not in recipes_to_check:
+                    recipes_to_check.append(dep)
```
lbonn: Didn't expect that appending to a list inside while iterating on its item would work, but looks like it does. It feels a bit magical to me so perhaps it would warrant a small comment.

patrickvacek: Is it explicitly wrong/bad/undefined to do so? Should I use a queue instead?

lbonn: Well, the internet is full of people saying it's bad but actually the official docs mention this behavior: https://docs.python.org/3/tutorial/controlflow.html#for-statements. Didn't really find a "specification" though.

patrickvacek: Since I'm not deleting anything, I can't see why this would be a problem. If I use a queue, then I end up needing two data structures, a list to hold everything and a queue to hold just what still needs to be processed. Queues apparently also depend on threading, which I don't need. Deques and other structures don't really appear to be any better. For the moment, I'm hesitant to use something else if I don't have a good reason to.

lbonn: Yes, I think it's fine, the conciseness probably out-weights the cost here. I was just suggesting adding an explanatory comment.

I did write a version using a deque and a set but it actually took some tries to get it work correctly.

patrickvacek: Done (and rebased). I also tried to do it with a queue and realized it was quite ugly.

scripts/find_packages.py
```diff
@@ -0,0 +1,124 @@
...
+    abcd_file.write('  homepage_url: "%s"\n' % homepage)
+    abcd_file.write('  source_artifact:\n')
+    repos = []
+    for src in src_uri:
+        # Strip options.
+        # TODO: ignore files with apply=false?
+        semi_pos = src.find(';')
+        if semi_pos > 0:
+            src = src[:semi_pos]
+        if src[0:7] == 'file://':
```
lbonn: All of these can be be replaced by `.startswith()`

scripts/find_packages.py
```diff
@@ -0,0 +1,124 @@
...
+            local = fetch.localpath(src)
+            abcd_file.write('  - "%s"\n' % local)
+        else:
+            abcd_file.write('  - "%s"\n' % src)
+            if src[0:7] != 'http://' and src[0:8] != 'https://' and src[0:6] != 'ftp://' and src[0:6] != 'ssh://':
+                repos.append(src)
+    if len(repos) > 1:
+        print('Multiple repos not fully supported yet. Pacakge: %s' % info.pn)
+    for repo in repos:
+        colon_pos = repo.find(':')
```
lbonn:
```
vcs_type, _, url = repo.split('://')
```

scripts/find_packages.py
```diff
@@ -0,0 +1,124 @@
...
+sys.path = sys.path + [bb_lib_path]
+
+import bb.fetch2
+import bb.tinfoil
+
+
+def print_deps(tinfoil, abcd_file, rn):
+    try:
+        info = tinfoil.get_recipe_info(rn)
+    except:
```
lbonn: Bare except are discouraged (and marked as warnings by flake8): https://www.python.org/dev/peps/pep-0008/#programming-recommendations. It can usually be replaced by `except Exception:` which, at least, doesn't catch KeyboardInterrupt and plays more nicely with python's convention with exceptions.

scripts/find_packages.py
lbonn: Or actually, it could first extract the protocol with a split (other comment) and only filter on this part afterwards.

lbonn:
Other than the python style remarks, this looks good and I've run it successfully.

patrickvacek:
@lbonn I've attempted to fix all of your concerns. Thanks for the suggestions!

PR #324, opened 2018-05-28
Fixup oe-selftest

cajun-rat:


patrickvacek:
So this ostree business is annoying, but the problem is from how it was before, not my new change. To fix it, you need to clean out any lingering builds of ostree and ostree-native. I had to do `bitbake -c cleanall ostree ostree-native` in each version of the build environment (grub, HSM, etc.). This could probably be done by sticking that line directly into oe-selftest where convenient just for one run.

cajun-rat:
This command reliably works with the revert but not before it:

    bitbake -c cleanall ostree ostree-native && oe-selftest -r updater

I think that `autotools-brokensep` is required for ostree because it doesn't support out-of-tree builds. Looking at the other recipes in OE, this is quite a common requirement, so I don't think it is anything to be worried about.

patrickvacek:
> This command reliably works with the revert but not before it:

The problem with that command is that it only cleans ostree and ostree-native for the current configuration. It will not clean it for the others (i.e. HSM, grub, etc.). This is an annoying side-effect of using oe-selftest like we do.

I tried to understand why the `brokensep` hack was required for OSTree. As best as I can tell, the only reason was because of the `autogen.sh` thing, which I fixed. I checked this manually with an independent ostree clone: it *does* actually support out-of-tree builds. There may have been a time when it did not, but that time has passed.

If we revert my fix, then we need to readdress https://github.com/advancedtelematic/meta-updater/pull/318. Since requiring `CLEANBROKEN` seems like an explicit admission that the old way is broken (which `brokensep` also acknowledges), why not just do it the "right" way?

cajun-rat:
(Summary of our offline conversation: We think this is caused by old builds that have polluted the sstate-cache, and bitbake cleanall only works against the current `MACHINE`.)

I've replaced it with a patch that just bumps `PR`. This is currently being tested.

patrickvacek:
Can we bump aktualizr to the latest version along with this PR? @EpicNuts would also like this.

PR #336, opened 2018-06-12
Ostree recipe cleanups

rsalveti:
Recipe cleanup, containing:
- fix license description
- reduce amount of local customizations
- remove need for native specific dependencies
- break up into more packages (reduce runtime dependency list when only ostree is required)


lbonn:
This is great, thanks!

PR #338, opened 2018-06-12
image_types_ostree: no need to create boot/loader folders when creating ostree repo

rsalveti:
The boot/loader folders are only required at the deploy stage, which is
already created by image_types_ota.bbclass.

### DIFF
--- a/classes/image_types_ostree.bbclass
+++ b/classes/image_types_ostree.bbclass
@@ -121,12 +121,6 @@ IMAGE_CMD_ostree () {
         cp ${SOTA_SECONDARY_ECUS} var/sota/ecus
     fi
 
-    # Creating boot directories is required for "ostree admin deploy"
-
-    mkdir -p boot/loader.0
-    mkdir -p boot/loader.1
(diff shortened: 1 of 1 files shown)


OYTIS:
@rsalveti I'm pretty sure that creating at least `boot/loader` was necessary at the time when the recipe was written (otherwise `ostree deploy` wouldn't recognize that U-boot is used). I'll retest.

rsalveti:
> @rsalveti I'm pretty sure that creating at least boot/loader was necessary at the time when the recipe was written (otherwise ostree deploy wouldn't recognize that U-boot is used). I'll retest.

ostree deploy is only called via image_types_ota.bbclass, which creates the required boot/loader folders before calling deploy.

cajun-rat:
This appears to work for me.  I can't reproduce the failure that caused this check to be added in the first place, so I suspect that creating these directories is no-longer required.

cajun-rat:
Tested manually on R-Pi + oe-selftest

PR #341, opened 2018-06-15
Compatibility with latest aktualizr

cajun-rat:
Note that the HsmTests still sometimes fail


cajun-rat:
I just bumped the version of aktualizr in this PR, and are running oe-selftest now.

patrickvacek:
Looks fine, I'm running oe-selftest in a new checkout since my other one is busy, so it'll take a bit to run.

PR #344, opened 2018-06-18
First pipeline setup for meta-updater CI

lbonn:
- check out meta-updater and read the Jenkinsfile
- check out aktualizr
- use repo to checkout everything needed for an image build
- move repo's meta-updater commit to the one that triggered CI
- build core-image-minimal

With these properties:

- keep existing workspace between run, so that builds are naturally cached. It might need manual clean intervention sometimes
- a build is triggered when a commit is pushed on aktualizr master, meta-updater master or any branch of a git tracked by repo

### DIFF
--- a/scripts/ci/Jenkinsfile
+++ b/scripts/ci/Jenkinsfile
@@ -1,0 +1,27 @@
+pipeline {
+  agent any
+  stages {
+    stage('checkout') {
+      steps {
+        checkout([$class: 'RepoScm',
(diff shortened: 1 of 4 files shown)


lbonn:
@patrickvacek  I haven't tried to make it run on PRs. But, thinking about it, it should only be a matter of in-Jenkins configuration. The problem with GitHub integration is that I always struggle to set up the right credentials for API access, there is a kind of UI bug around that it seems...

PR #345, opened 2018-06-19
Bugfix/staging dir host sumo

patrickvacek:


patrickvacek:
Alright, this should be ready to go. It includes some backported fixes from master, but not everything. It also includes the hack to work around the HSM test sysroot population issue.

PR #346, opened 2018-06-20
Test --running-mode=once

cajun-rat:


cajun-rat:
I thought that the aktualizr version was already new enough. Perhaps that isn't the case:)

Issue #342, opened 2018-06-15
Support tarballs OTA images


agners:
To integrate meta-updater with a custom installation systems we need a tarball of the updateable ostree tree. Currently  image_types_ota creates ext4 images only. The class should be extended to support tar. Maybe the generation and naming of the image formats should be hierarchical, e.g. otaimg.ext4/otaimg.tar...

cajun-rat:
It should be possible to do that, maybe as some kind of optional step, because we don't want to generate multiple images for everyone.  A hack in image_types_ota that just checked a flag and created the image if necessary would be simple, but perhaps not ideal.

falstaff84:
I would suggest to just create a new image type (otaimg.tar) and then move most of the functionality currently in IMAGE_CMD_otaimg into a common function. Then IMAGE_CMD_otaimg ends up calling this function and creating the ext4 fs, and IMAGE_CMD_otaimg.tar the same but creating the tarball in the end...

patrickvacek:
@agners Can I close this since the PR was merged and it looks like your needs have been met? Or is there something more related to this that I've overlooked?

agners:
Our needs have been met with that PR merged! You can close this issue. Thanks!

Issue #334, opened 2018-06-11
btrfs rootfs


mihaelaapetroaie:
Hello, 

Is there any easy way to get the rootfs (otaimg) as btrfs instead of ext4?

I've been trying to make some changes in `meta-updater/classes/image_types_ota.bbclass`, it looks like this now:

`MIN_BTRFS_SIZE="16384"`
               ` if  ${OTA_ROOTFS_SIZE} -lt ${MIN_BTRFS_SIZE} ] ; then`
               ` OTA_ROOTFS_SIZE=${MIN_BTRFS_SIZE}`
               ` bbwarn "Rootfs size is too small for BTRFS. Filesystem will be extended to ${size}K"`
              ` fi`
               ` # create image`
                `rm -rf ${DEPLOY_DIR_IMAGE}/${IMAGE_NAME}.otaimg`
               ` sync`
               ` dd if=/dev/zero of=${DEPLOY_DIR_IMAGE}/${IMAGE_NAME}.otaimg seek=$OTA_ROOTFS_SIZE count=$COUNT bs=1024`
                `mkfs.btrfs  ${DEPLOY_DIR_IMAGE}/${IMAGE_NAME}.otaimg -L otaroot -d ${PHYS_SYSROOT}`
                `rm -rf ${PHYS_SYSROOT}`

But I am getting an error when trying to build my image:

`ERROR: unknown profile /storage/mika/overlays/poky/build/tmp-glibc/work/raspberrypi3-oe-linux-gnueabi/remote-management-minimal/1.0-r0/ota-sysroot-MpZDO`

I am trying to build the [rocko] branch on an [Ubuntu] machine

Thank you!

lbonn:
The error you're seeing comes from the usage of `mkfs.btrfs` which differs from `mkfs.ext4` for some reason. It should be `mkfs.btrfs ${DEPLOY_DIR_IMAGE}/${IMAGE_NAME}.otaimg -L otaroot -r ${PHYS_SYSROOT}` (`-r` instead of `-d`)

However, after fixing that, another cryptic error appeared:

```
| ERROR: unable to traverse directory ...build-qemu/tmp/work/qemux86_64-poky-linux/core-image-minimal/1.0-r0/ota-sysroot-VDmyv: 1
| ERROR: error wihle filling filesystem: 1
| ERROR: attempt to start transaction over already running one
| extent buffer leak: start 4440064 len 16384
[...]

PR #347, opened 2018-06-22
Offload image creation to wic source plugin (otaimage)

Jmennius:
This allows generating sysroot images with different filesystems
transparently and removes the need to do it ourselves.
Options that control filesystem creation are now available
via standard wic options.

One thing to notice is that intermediate result in form of ${DEPLOY_DIR_IMAGE}/${IMAGE_NAME}.otaimg is not available anymore, but if you need one - you can build it from OSTREE_IMAGE_SYSROOT.

### DIFF
--- a/classes/image_types_ota.bbclass
+++ b/classes/image_types_ota.bbclass
@@ -7,42 +7,10 @@
 # boot scripts, kernel and initramfs images
 #
 
-do_image_otaimg[depends] += "e2fsprogs-native:do_populate_sysroot \
+do_image_otaimg[depends] += " \
 			${@'grub:do_populate_sysroot' if d.getVar('OSTREE_BOOTLOADER', True) == 'grub' else ''} \
 			${@'virtual/bootloader:do_deploy' if d.getVar('OSTREE_BOOTLOADER', True) == 'u-boot' else ''}"
 
-calculate_size () {
-	BASE=$1
-	SCALE=$2
-	MIN=$3
-	MAX=$4
-	EXTRA=$5
-	ALIGN=$6
-
-	SIZE=`echo "$BASE * $SCALE" | bc -l`
-	REM=`echo $SIZE | cut -d "." -f 2`
-	SIZE=`echo $SIZE | cut -d "." -f 1`
-
-	if [ -n "$REM" -o ! "$REM" -eq 0 ]; then
-		SIZE=`expr $SIZE \+ 1`
-	fi
-
-	if [ "$SIZE" -lt "$MIN" ]; then
-		SIZE=$MIN
-	fi
-
-	SIZE=`expr $SIZE \+ $EXTRA`
-	SIZE=`expr $SIZE \+ $ALIGN \- 1`
-	SIZE=`expr $SIZE \- $SIZE \% $ALIGN`
-
-	if [ -n "$MAX" ]; then
-		if [ "$SIZE" -gt "$MAX" ]; then
-			return -1
-		fi
-	fi
-	
-	echo "${SIZE}"
-}
 
 export OSTREE_OSNAME
 export OSTREE_BRANCHNAME
@@ -66,7 +34,9 @@ IMAGE_CMD_otaimg () {
 		fi
 
 
-		PHYS_SYSROOT=`mktemp -d ${WORKDIR}/ota-sysroot-XXXXX`
+		PHYS_SYSROOT=${OSTREE_IMAGE_SYSROOT}
+		rm -rf ${PHYS_SYSROOT} || true
+		mkdir ${PHYS_SYSROOT}
 
 		ostree admin --sysroot=${PHYS_SYSROOT} init-fs ${PHYS_SYSROOT}
 		ostree admin --sysroot=${PHYS_SYSROOT} os-init ${OSTREE_OSNAME}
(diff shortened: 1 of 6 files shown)


cajun-rat:
This looks like a good improvement. If it changes the names of generated files then we'll have to do a few manual tests before merging this.  We also need to check that the names are not in any documentation, and the flashing scripts we have.

== TODO 
- [ ] Check R-Pi works still (note that we'll need to bump the user facing documentation when updater-repo switches to sumo+1.
- [ ] oe-selftest -r updater passes
- [ ] run-qemu-ota works
- [ ] Minnowboard Max works

Jmennius:
It removes only what I presume is the intermediate result (`*.otaimg`), although it is indeed documented in README.
`*.otaimg` image is also used in updater selftest `test_add_package()` - I will look into fixing this.

OYTIS:
@Jmennius Thank you. I think we can just use *wic file for the same test. My only concern for now is qemu which still uses *.otaimg.

Jmennius:
I have updated (foce-pushed) with changes to README and selftest.
This specific `test_add_package()` test passed for me (although some other unrelated tests failed).

I am not sure what to do with the fact that we can't easily reuse OE infrastructure built around `image_types*` and `IMAGE_ROOTFS`, so having image or tar of a sysroot only is something that will require hacks.. that's a bit disappointing.

But qemu is able to run from full wic image, so it is basically enought to change `qemucommand.py` and `run-qemu-ota.py`? I am not experienced in running qemu, unfortunately.

cajun-rat:
Hi,

I really want to get this change in, and I think that getting the qemu and oe-selftest parts working are all that is really left now (plus a couple of manual tests).

Running a qemu image is handled by the script `meta-updater/scripts/run-qemu-ota`, which is normally run from the build directory:

    updater-repo/build-qemu$ bitbake core-image-minimal
    updater-repo/build-qemu$ ../meta-updater/scripts/run-qemu-ota --no-gui

(Then either ssh in if you have dropbear in local.conf, or connect to the serial console)

At the moment it doesn't seem to be generating the .wic by default:

    build-qemu/tmp/deploy/images/qemux86-64$ ls
    bzImage
    bzImage--4.14.48+git0+d64aec9793_97c8063d2d-r0-qemux86-64-20180720100441.bin
    bzImage-qemux86-64.bin
    core-image-minimal-qemux86-64-20180726080028.qemuboot.conf
    core-image-minimal-qemux86-64-20180726080028.rootfs.ext4
    core-image-minimal-qemux86-64-20180726080028.rootfs.manifest
    core-image-minimal-qemux86-64-20180726080028.rootfs.ostree.tar.bz2
    core-image-minimal-qemux86-64-20180726080028.rootfs.tar.bz2
    core-image-minimal-qemux86-64-20180726080028.testdata.json
    core-image-minimal-qemux86-64.ext4
    core-image-minimal-qemux86-64.manifest
    core-image-minimal-qemux86-64.qemuboot.conf
    core-image-minimal-qemux86-64.rootfs.ostree.tar.bz2
    core-image-minimal-qemux86-64.tar.bz2
    core-image-minimal-qemux86-64.testdata.json

Jmennius:
Hi,

I am now thinking that the right way to go ™️ is to actually have our libostree-managed sysroot in `IMAGE_ROOTFS`, in contrary to being deployed in some other directory (`ota-sysroot-` right now).
This will help people wanting different formats of rootfs since `IMAGE_ROOTFS` will be available to any image_type, not only wic as with this change.
It will also bring ostree/sota closer to being an `IMAGE_FEATURE`, not `DISTRO_FEATURE`.
From what I see this solution covers pretty much everything (or at least it is more complete than this change).

Basically, what I am talking about, is that we'll have a task that runs after `do_rootfs`, let's call it `do_rootfs_ostree` for now:
- do everything that `image_types_ostree` does: perform necessary changes to the `IMAGE_ROOTFS`, push to `OSTREE_REPO`
- clean the `IMAGE_ROOTFS` (this should be OK since rootfs is recreated every time anything required to build it changes)
- do what first part of `image_types_ota` does: check out the rootfs from `OSTREE_REPO` into `IMAGE_ROOTFS`

The implementation shouldn't take long, but I won't be able to work on this in near future.
Let me know in case you are willing to accept/help with such a change - I will likely start working on it earlier.

Thanks.

agners:
@Jmennius I/we are actually also reworking parts of the image creation process.

Let me back up a bit: We use our own installer to install images which requires rootfs tarballs. Therefor we created issue #342 and added packing for tar in pull request #352 (besides the currenlty supported ext4).

This works well, but as you observed too, things are not ideal currently. I started to cleanup things from the tail end (see #408 and #410). This are very incremental patches, but I think it is easier to review and avoid breaking too much (at once)...

What is pretty annoying right now is that the do_image processes creates a tarball between in `do_image_ostree` which then immediately gets extracted again in `do_otasetup`.

Your approach moving everything in $IMAGE_ROOTFS seems a bit radical to me but might work. There are some entanglement which am not sure about (e.g. the `HOME_TMP` in `do_otasetup` which needs that tar.bz2...). I guess ideally we this should be done step by step.

patrickvacek:
@Jmennius and @agners,

In light of recently merged PRs that changed how some of this works (like https://github.com/advancedtelematic/meta-updater/pull/408 and https://github.com/advancedtelematic/meta-updater/pull/410), how do you see this PR moving forward? Is this still something that you are interested in, or have we moved enough in a different direction that this is no longer useful?

If you still do see this as worth working with, please rebase it and let's discuss the pros and cons of it. Otherwise, it might be time to close this. In any event, thanks for your patience and interest!

agners:
@patrickvacek at least the ideas @Jmennius outlined in his last post go further than my changes... I think we agree that the OSTree/OTA image generation process should be made more generic.

patrickvacek:
Closing due to lack of activity and substantial changes to the code in the meantime. This PR as it stands would require substantial rebasing and the author has proposed a different strategy anyway. (On that topic, though, I don't understand why we would want to remove `IMAGE_ROOTFS` and then recreate it based on the same content.)

agners:
I guess @Jmennius suggested to remove `IMAGE_ROOTFS` since he does then the subsequent checkout from the OSTree repo. After looking more into the topic, I doubt that this is the right approach anyway, since the rootfs prepared for OSTree commit is not the same as the checked out OSTree repo... There is no value in recycling the same location.

As far as I know @Jmennius main goal was to add btrfs support. With the recent rework that should be already quite a bit easier.

patrickvacek:
We're still quite open to organizational improvements, but this particular PR seems to have fallen by the wayside. For btrfs support, we still have https://github.com/advancedtelematic/meta-updater/issues/334 open. It's not a priority for us, but it's still an interesting topic.

PR #348, opened 2018-06-25
Ci/dockerize

lbonn:
Mostly general CI improvements, makes it easier to setup a build slave

### DIFF
--- a/scripts/ci/Dockerfile.bitbake
+++ b/scripts/ci/Dockerfile.bitbake
@@ -1,0 +1,37 @@
+FROM debian:stable
+LABEL Description="Image for bitbaking"
+
+RUN sed -i 's#deb http://deb.debian.org/debian stable main#deb http://deb.debian.org/debian stable main contrib#g' /etc/apt/sources.list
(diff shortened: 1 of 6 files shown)


scripts/ci/Dockerfile.bitbake
```diff
@@ -0,0 +1,37 @@
...
+     iputils-ping \
+     iproute \
+     libpython-dev \
+     libsdl1.2-dev \
+     locales \
+     procps \
+     python \
+     python3 \
+     python3-pexpect \
+     python3-requests \
```
patrickvacek: Do we really need python3-requests? I thought we gave up on using that.

Also, requiring python, python3, and libpython-dev is pretty annoying, but if that's what we need, then so be it.

lbonn: Actually, I adapted a lot of the stuff from https://github.com/advancedtelematic/dockerfiles/blob/master/bitbakery/Dockerfile. I think you're right about python3-requests, which was needed for a brief period.

PR #352, opened 2018-06-30
Support for creating an otaimg.tar file in addition to the ext4 image

bclouser:
advancedtelematic/meta-updater#342

Added variable `BUILD_OTA_TARBALL` that when set to "1" will create a tarred output of the ota image in addition to the ext4 output. So, when set _lmp-gateway-image.otaimg_ and _lmp-gateway-image.ota.tar_ are produced from the build. 
This

### DIFF
--- a/classes/image_types_ota.bbclass
+++ b/classes/image_types_ota.bbclass
@@ -7,7 +7,11 @@
 # boot scripts, kernel and initramfs images
 #
 
-do_image_otaimg[depends] += "e2fsprogs-native:do_populate_sysroot \
+do_image_ota_ext4[depends] += "e2fsprogs-native:do_populate_sysroot \
+			${@'grub:do_populate_sysroot' if d.getVar('OSTREE_BOOTLOADER', True) == 'grub' else ''} \
+			${@'virtual/bootloader:do_deploy' if d.getVar('OSTREE_BOOTLOADER', True) == 'u-boot' else ''}"
+
+do_image_ota_tar[depends] += "e2fsprogs-native:do_populate_sysroot \
 			${@'grub:do_populate_sysroot' if d.getVar('OSTREE_BOOTLOADER', True) == 'grub' else ''} \
 			${@'virtual/bootloader:do_deploy' if d.getVar('OSTREE_BOOTLOADER', True) == 'u-boot' else ''}"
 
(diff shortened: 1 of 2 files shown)


classes/image_types_ota.bbclass
```diff
@@ -7,7 +7,11 @@
...
 #
 
-do_image_otaimg[depends] += "e2fsprogs-native:do_populate_sysroot \
+do_image_ota_ext4[depends] += "e2fsprogs-native:do_populate_sysroot \
+			${@'grub:do_populate_sysroot' if d.getVar('OSTREE_BOOTLOADER', True) == 'grub' else ''} \
+			${@'virtual/bootloader:do_deploy' if d.getVar('OSTREE_BOOTLOADER', True) == 'u-boot' else ''}"
+
+do_image_ota_tar[depends] += "e2fsprogs-native:do_populate_sysroot \
 			${@'grub:do_populate_sysroot' if d.getVar('OSTREE_BOOTLOADER', True) == 'grub' else ''} \
 			${@'virtual/bootloader:do_deploy' if d.getVar('OSTREE_BOOTLOADER', True) == 'u-boot' else ''}"
```
falstaff84: Shouldn't grub/u-boot be a dependency of the do_otasetup task?

Also, for do_image_ota_tar we probably do not need e2fsprogs... So probably no direct dependencies needed for do_image_ota_tar.

classes/image_types_ota.bbclass
```diff
@@ -51,92 +50,132 @@ export OSTREE_BOOTLOADER
...
-		if [ -z "$OSTREE_OSNAME" ]; then
-			bbfatal "OSTREE_OSNAME should be set in your local.conf"
-		fi
+export OTA_SYSROOT="${WORKDIR}/ota-sysroot"
 
-		if [ -z "$OSTREE_BRANCHNAME" ]; then
-			bbfatal "OSTREE_BRANCHNAME should be set in your local.conf"
-		fi
+## Common OTA image setup
+fakeroot do_otasetup () {
```
OYTIS: Sorry, why fakeroot?

falstaff84: Reading https://www.yoctoproject.org/docs/2.6/overview-manual/overview-manual.html#fakeroot-and-pseudo this seems correct to me, we want to do operations as root.

I think before the fakeroot flag has been set by this python script: https://github.com/openembedded/openembedded-core/blob/master/meta/classes/image.bbclass#L338

falstaff84:
Is there anything we can help to get this merged?

cajun-rat:
Sorry for the delay, I'm looking at this now. Github creates rather complex diffs, which confused me before.  I'm running the tests now, shouldn't be long.

cajun-rat:
oe-selftest -r updater passed

PR #360, opened 2018-07-18
Support initramfs in FIT

Jmennius:
Together with advancedtelematic/meta-updater-raspberrypi#25, this change moves initramfs into FIT.
Having initramfs in FIT is beneficial in several ways:
* checksumming for initramfs
* signing initramfs
* no need to change u-boot scripts or environment if initramfs is removed or added.
..and more.


Jmennius:
There is one issue with this pull, which I completely forgot about... In fact, this is an issue with initramfs script.

The issue is that `init` script from `ostree-initrd.bb` uses `pivot_root` and therefore should be on a filesystem using ramdisk (legacy initrd). So it should be on a filesystem of some kind - ext4 right now.
But OE doesn't suppose anybody would use initrd instead of initramfs, it only supports initramfs (not filesystem image, but a cpio archive), see `kernel.bbclass` and `kernel-fitimage.bbclass`.
See [1] for more information on ramfs/initramfs and ramdisk/initrd.

Basically, what needs to be done in addition to changes already present in this PR, is:
* change init script to use `switch_root` instead of `pivot_root`
* change `INITRAMFS_FSTYPES` to some cpio variant

[1] [Linux kernel: ramfs, rootfs and initramfs](https://www.kernel.org/doc/Documentation/filesystems/ramfs-rootfs-initramfs.txt)

Jmennius:
Ok, I have fixed `initramfs-ostree-image` issue (3rd commit).

This was tested on a custom platform and doesn't require changes to OE anymore.
I will test this on stock RPi3 next week. (waiting for feedback anyway!)

classes/sota.bbclass
```diff
@@ -20,13 +20,13 @@ WKS_FILE_sota ?= "sdimage-sota.wks"
 
 EXTRA_IMAGEDEPENDS_append_sota = " parted-native mtools-native dosfstools-native"
 
-OSTREE_INITRAMFS_FSTYPES ??= "${@oe.utils.ifelse(d.getVar('OSTREE_BOOTLOADER', True) == 'u-boot', 'ext4.gz.u-boot', 'ext4.gz')}"
+INITRAMFS_FSTYPES ??= "${@oe.utils.ifelse(d.getVar('OSTREE_BOOTLOADER', True) == 'u-boot', 'cpio.gz.u-boot', 'cpio.gz')}"
```
OYTIS: We're using ext4 because cpio doesn't support extended attributes which causes problems when booted system uses it.

classes/sota_raspberrypi.bbclass
```diff
@@ -2,6 +2,8 @@ RPI_USE_U_BOOT_sota = "1"
 
 KERNEL_CLASSES_append_sota = " kernel-fitimage"
 KERNEL_IMAGETYPE_sota = "fitImage"
+OSTREE_KERNEL = "${KERNEL_IMAGETYPE}-${INITRAMFS_IMAGE}-${MACHINE}.bin"
+INITRAMFS_FSTYPES = "cpio.gz"
```
OYTIS: See above about ext4 vs cpio

cajun-rat:
Am I right in thinking that we have two conflicting requirements here:

1) AGL uses Smack and xattrs, which doesn't (didn't?) work with a cpio Initramfs (@OYTIS ?)
2) FIT images require a cpio format initramfs (@Jmennius ?)

classes/image_types_ostree.bbclass
```diff
@@ -124,7 +122,14 @@ IMAGE_CMD_ostree () {
...
 
     cp ${DEPLOY_DIR_IMAGE}/${OSTREE_KERNEL} boot/vmlinuz-${checksum}
-    cp ${DEPLOY_DIR_IMAGE}/${OSTREE_INITRAMFS_IMAGE}-${MACHINE}${RAMDISK_EXT} boot/initramfs-${checksum}
+
+    if [ "${KERNEL_IMAGETYPE}" = "fitImage" ]; then
+        # this is a hack for ostree not to override init= in kernel cmdline -
+        # make it think that the initramfs is present (while it is in FIT image)
+        touch boot/initramfs-${checksum}
+    else
+        cp ${DEPLOY_DIR_IMAGE}/${INITRAMFS_IMAGE_NAME}.${INITRAMFS_FSTYPES} boot/initramfs-${checksum}
```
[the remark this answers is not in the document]
Jmennius: Will fix that, thanks!

Jmennius:
Not really a conflict, but a missing feature(s).

1. There are patches on the (kernel) mailing list which add xattrs support to cpio (format, kernel, tools). I am not sure if these were accepted. This is definitely a long-term solution.

2. Not really. FIT itself and related infrastructure doesn't care.
The only thing which only likes `cpio` is OE (see `kernel.bbclass` and `kernel-fitimage.bbclass`).
We can send a (trivial) patch to fix that. Initially, I thought that it would not be appropriate since initrd is kind of legacy (initrd == everything that requires an actual filesystem; cpio does not). But if initramfs (cpio) is not on par with initrd (ext and other FS) feature wise - it makes sense.

IMO, the best way is to have cpio support for xattrs. Need to look into it.

BTW, we can boot into ostree system w/o ramdisk, what do you think about this?

cajun-rat:
Booting without ramdisk would be a great thing to support, that would be nicer for an embedded system and solve the smack/xattrs problem at the same time.

Have you looked into how much work the no-ramdisk option is? If it is easy then we can consider using that as the route for smack/xattrs support.

cajun-rat:
Just had a chat with @OYTIS. AGL is the only user of xattrs (via smack) that I'm are aware of, and we can simply override `INITRAMFS_FSTYPES` in the AGL config to fix it up for them.  I'm not completely sure why the ramdisk needs to support xattrs when we pivot root to a filesystem that does, but neither @OYTIS or myself have had time to investigate that fully.

If/when cpio xattrs make it into mainline kernel, or when meta-updater supports booting without a ramdisk, or when the problem with smack booting from cpio goes away, then we can drop the 'ext4' initrd thing altogether. That feels like a good trade-off to get FIT images for everyone else!

cajun-rat:
I forgot the `cp ${DEPLOY_DIR_IMAGE}/${INITRAMFS_IMAGE}-${MACHINE}.${INITRAMFS_FSTYPES} boot/initramfs-${checksum}` change wasn't in this branch yet. I just backed this merge out.  Can you make that change on your branch and I'll merge it, please?

Jmennius:
I have pushed a fix to branch (although this pull request is closed - I suppose you'll need to merge manually).

Regarding AGL overriding `INITRAMFS_FSTYPES` - sounds good, but there is one caveat which I have mentioned previously - `kernel-fitimage.bbclass` won't include non-cpio image in FIT.
OE-core needs to be patches for this to work:
```diff --git a/meta/classes/kernel-fitimage.bbclass b/meta/classes/kernel-fitimage.bbclass
--- a/meta/classes/kernel-fitimage.bbclass
+++ b/meta/classes/kernel-fitimage.bbclass
@@ -377,7 +377,7 @@
 	#
 	if [ "x${ramdiskcount}" = "x1" ] ; then
 		# Find and use the first initramfs image archive type we find
-		for img in cpio.lz4 cpio.lzo cpio.lzma cpio.xz cpio.gz cpio; do
+		for img in cpio.lz4 cpio.lzo cpio.lzma cpio.xz cpio.gz cpio ext4.gz ext4; do
 			initramfs_path="${DEPLOY_DIR_IMAGE}/${INITRAMFS_IMAGE_NAME}.${img}"
 			echo "Using $initramfs_path"
 			if [ -e "${initramfs_path}" ]; then
```

So maybe we don't really want to merge this until we resolve this cleanly :(

cajun-rat:
Would it be possible to make the AGL/xattrs case work by disabling FIT images in the AGL config?

Jmennius:
Yes, but to support FIT and non-FIT simultaneously we'll also need two u-boot scripts or a bit more sophisticated script to check whether standalone initramfs is available.
advancedtelematic/meta-updater-raspberrypi#25 will only work when initramfs is not standalone (it is either not present at all or in FIT).
In general, I would try to avoid diversity.

If closing this pull is not urgent I would try either of:
1. Send a patch to oe-core (allowing ext4 initramfs in FIT)
2. Try to understand the status of cpio+xattrs, and if it is going to be accepted - we can add it as a patch.

With 1 being a temporary/fast solution, and 2 for a long run.

classes/image_types_ostree.bbclass
```diff
@@ -5,14 +5,12 @@ do_image_ostree[depends] += "ostree-native:do_populate_sysroot \
                         coreutils-native:do_populate_sysroot \
                         unzip-native:do_populate_sysroot \
                         virtual/kernel:do_deploy \
-                        ${OSTREE_INITRAMFS_IMAGE}:do_image_complete"
+"
```
cajun-rat: Are you sure this isn't required?  I got a failure when building of `cp: cannot stat '…/tmp/deploy/images/qemux86-64/initramfs-ostree-image-qemux86-64.cpio.gz': No such file or directory
`
I hadn't seen that failure before, so it might be a race condition.

Jmennius: Hmm, this shouldn't happen even if you are not using FIT.
initramfs build is always ordered before kernel deploy task (in kernel.bbclass) and on the previous line we depend on kernel deploy task.

Is the image name correct? Maybe it is altered in some way for qemu..

cajun-rat: Weird. With a clean tmp directory it reliably fails with because `initramfs-ostree-image-qemux86-64.cpio.gz` is missing. Adding `${INITRAMFS_IMAGE}:do_image_complete` fixes it.

However, the only difference between the output of `bitbake -g core-image-minimal:do_image_ostree` with and without the change is an extra direct edge from `core-image-minimal.do_image_ostree` to `initramfs-ostree-image.do_image_complete`. In theory this shouldn't matter because there is a indirect dependency between the two already. In practice it appears there is something more subtle going on.

PR #363, opened 2018-07-26
Feat/pro tech #1 rocko

patrickvacek:
Sprint release candidate.

PR #364, opened 2018-07-26
Feat/pro tech #1 sumo

patrickvacek:
Sprint release candidate.


cajun-rat:
I'm still getting hsm and implicit provisioning test failures.  Will try with the plain 'sumo' branch and see if it is new.

cajun-rat:
This seems to have broken the `updater.HsmTests` test.  It works on the current sumo but not with this patch.  I tried cleaning the openssl/openssl-native recipes, but that didn't seem to fix the problem.

lbonn:
Maybe #363 is also affected by the HSM test failure? I'm running the tests right now

cajun-rat:
Looks like the HSM failure is real.  Let's hold off merging this (and the Rocko one) until we can resolve that.

PR #366, opened 2018-08-08
aktualizr: bump to latest for key rotation and storage updates (sumo)

patrickvacek:
Work around a bug in oe-selftest with migration and aktualizr-info
interplay by forcing --allow-migrate for now.


cajun-rat:
`oe-selftest -r updater passes`

PR #369, opened 2018-08-09
Allow to use provisioning recipes without aktualizr

OYTIS:
Allows apps built on top of aktualizr reuse the provisioning recipes.

### DIFF
--- a/recipes-sota/aktualizr/aktualizr-auto-prov.bb
+++ b/recipes-sota/aktualizr/aktualizr-auto-prov.bb
@@ -5,7 +5,6 @@ SECTION = "base"
 LICENSE = "MPL-2.0"
 LIC_FILES_CHKSUM = "file://${WORKDIR}/LICENSE;md5=9741c346eef56131163e13b9db1241b3"
 DEPENDS = "aktualizr-native zip-native"
-RDEPENDS_${PN} = "aktualizr"
 PV = "1.0"
 PR = "6"
 
--- a/recipes-sota/aktualizr/aktualizr-ca-implicit-prov.bb
+++ b/recipes-sota/aktualizr/aktualizr-ca-implicit-prov.bb
@@ -10,7 +10,6 @@ LICENSE = "MPL-2.0"
 LIC_FILES_CHKSUM = "file://${WORKDIR}/LICENSE;md5=9741c346eef56131163e13b9db1241b3"
 
 DEPENDS = "aktualizr-native openssl-native"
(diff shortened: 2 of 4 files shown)


patrickvacek:
I think we need to add `aktualizr` to the `DEPENDS` of each provisioning package if it isn't in `RDEPENDS`. Last time I messed around with this and asked on the mailing list, my ultimate conclusion was that the prov recipes had to depend on both `aktualizr` and `aktualizr-native`, and we more or less fulfilled the former with the `RDEPENDS` bit.

OYTIS:
@patrickvacek Sorry, don't quite get it. `DEPENDS` is for recipes that are required to build another recipe. Provisioning recipes don't seem to depend on target aktualizr in build time.
What this PR is for is to make them not depending on `aktualizr` even in run time. Just because instead of aktualizr there can be, i.e. `sachertorte` which can use this provisioning recipes just as well.

patrickvacek:
@OYTIS No problem, this was extremely unclear to me until two or three people on the mailing list laid it out for me. The prov recipes have to depend on `aktualizr-native` to get the native version of implicit_writer. (So actually `aktualizr-auto-prov` can skip that.) The all have to depend on `aktualizr` to get the `.toml` config files. Those have to come from `aktualizr` because native recipes do not populate the sysroot.

For more detail, you can try to look through the Yocto archives, for example, here:
https://lists.yoctoproject.org/pipermail/yocto/2018-June/041552.html
(Annoyingly, not all of the conversation is in one thread, presumably because my request to join the mailing list was never accepted.)

OYTIS:
That's surprising, I was sure that `STAGING_DIR_NATIVE` is exactly sysroot for native packages. I'll try a clean build.

patrickvacek:
[Here](https://lists.yoctoproject.org/pipermail/yocto/2018-June/041476.html) is maybe a more direct quote:

> > Oddly, I do not see a package subdirectory inside the aktualizr-native
> > directory in the build tree. I do see it inside the aktualizr directory,
> > though, and it contains everything that I would expect. Is there some
> > sort of configuration of the packaging system for a native recipe that I
> > haven't done correctly?
> 
> No, what you see is expected - there's no packaging for -native recipes.

OYTIS:
@patrickvacek Yea, but the packaging is kind of a separate issue.
OK, so I couldn't reproduce the issue, but I believe it can be transient. Let's leave everything as it is and I'll just add bbappends to Sachertorte layer.

PR #377, opened 2018-08-31
Bugfix/lshw aktualizr etc rocko

patrickvacek:


OYTIS:
What was the bug?

patrickvacek:
@OYTIS technically no bug here, you're right! But arguably the lshw issue would affect rocko just as well as sumo, and the KVM mode fix is arguably a bug, and a bunch of stuff was fixed in aktualizr... I guess I used the word "bugfix" loosely.

patrickvacek:
Yo @EpicNuts, ready for you to QA this.

PR #381, opened 2018-09-03
Fix raspberrypi image generation

lbonn:
First commit fixes the image generation failure, but it still didn't boot during my test.

More work coming up soon, hopefully.


patrickvacek:
Can confirm that the image builds successfully with this.

patrickvacek:
Any reason not to rebase this on the master so that the HSM test will pass?

lbonn:
I've changed it again, it should not mess with INITRAMFS_IMAGE_NAME actually...

lbonn:
Ok, finally made it to work. No more change here, but it needed these two PRs in meta-updater-raspberrypi:

https://github.com/advancedtelematic/meta-updater-raspberrypi/pull/25
https://github.com/advancedtelematic/meta-updater-raspberrypi/pull/26

Jmennius:
@lbonn this should be caused by recent changes to master - maybe it is time to use `kernel-artifact-names.bbclass` (also introduced recently).

lbonn:
@Jmennius It's already using `kernel-artifact-names` (`${KERNEL_FIT_LINK_NAME}` comes from there). Did you think of a way to make it cleaner?

PR #382, opened 2018-09-04
Provide example for setting UPTANE target version automatically

OYTIS:
Making a PR to rocko first because the primary user for that is Spekulatius.

Turns out there is no way to set a bitbake variable from a shell task, so I was choosing between writing a hacky python task calling git via `os.popen()` and adding a hack to `image_types_ostree.bbclass` to read version from a file. The latter seemed cleaner to me.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -175,7 +175,7 @@ Every time you build an image with `SOTA_PACKED_CREDENTIALS` set, a new entry in
 1. Set `GARAGE_TARGET_VERSION` variable in your `local.conf`.
 2. Write a recipe or a bbclass to write the desired version to `${STAGING_DATADIR_NATIVE}/target_version`. An example of such bbclass can be found in `classes/target_version_example.bbclass`.
 
-Please note that [target name, target version] pairs are expected to be unique in the system. If you build a new target with the same target version as a previously build one, the old package will be overwritten on the update server. It can have unpredictable effect on devices that have this version installed, it is not guaranteed that information will be reported correctly for such devices or that you will be able to update them (we're doing our best though). The easiest way to avoid problems is to make sure that your overriding version is as unique as OSTree commit hash.
+Please note that [target name, target version] pairs are expected to be unique in the system. If you build a new target with the same target version as a previously built one, the old package will be overwritten on the update server. It can have unpredictable effect on devices that have this version installed, and it is not guaranteed that information will be reported correctly for such devices or that you will be able to update them (we're doing our best though). The easiest way to avoid problems is to make sure that your overriding version is as unique as an OSTree commit hash.
 
 == QA with oe-selftest
 
--- a/README.adoc
+++ b/README.adoc
@@ -167,6 +167,16 @@ There are a few settings that can be controlled in `local.conf` to simplify the 
 | `TOOLCHAIN_HOST_TASK_append = " nativesdk-cmake "` | Use with `bitbake -c populate_sdk core-image-minimal` to build an SDK. See the https://github.com/advancedtelematic/aktualizr#developing-against-an-openembedded-system[aktualizr repo] for more information.
 |======================
 
+=== Overriding target version
+*Warning: overriding target version is a dangerous operation, make sure you understand this section completely before doing it.*
+
+Every time you build an image with `SOTA_PACKED_CREDENTIALS` set, a new entry in your Uptane metadata is created and you can see it in the OTA Garage UI if you're using one. Normally this version will be equal to OSTree hash of your root file system. If you want it to be different though you can override is using one of two methods:
+
+1. Set `GARAGE_TARGET_VERSION` variable in your `local.conf`.
+2. Write a recipe or a bbclass to write the desired version to `${STAGING_DATADIR_NATIVE}/target_version`. An example of such bbclass can be found in `classes/target_version_example.bbclass`.
+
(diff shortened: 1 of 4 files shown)


patrickvacek:
Do we need to fix the problem with how devices show up in the UI when overwriting `GARAGE_TARGETS_VERSION` before this is delivered to anyone? @EpicNuts found that there is a mismatch in how the target is recognized on the server and UI.

patrickvacek:
@OYTIS Looks fine to me as well, oe-selftest passes except for HsmTest, which is presumably because this is missing some fixes in the latest rocko. I'd suggest either rebasing or just adding this to one of the other open PRs on rocko awaiting Ben's review. If we can gather everything we need him to test into one package, that might make everyone's lives easier when he is back in the office.

patrickvacek:
By the way, do you know what the expected behavior is for pushing two images with the same `GARAGE_TARGET_VERSION`?

OYTIS:
@patrickvacek The old version just gets replaced with the new one. I tried the following scenario:

1. Build the image with `GARAGE_TARGET_VERSION=A` and ostree hash `HASH_A`, flash it to the device.
2. Build another image with `GARAGE_TARGET_VERSION=A` and ostree hash `HASH_B`. At this point the package in the garage UI gets replaced, one can see updated ostree hash. In the device view there is still a green mark near the version `A`, although the hash apparently doesn't match. It is a security problem by the way.
3. There is no way to update device to the 'new' `A` directly, because it is already 'installed'. But one can install some other version, and then `A` again, it works fine.

patrickvacek:
> The old version just gets replaced with the new one.

That's exactly what I saw as well. Seems like a problem. It "works" for now, but how do we go about fixing this? This seems like a server-side issue unless I'm missing something that we can do to prevent it.

OYTIS:
@patrickvacek What do you think is the correct behavior?

As I can see possible solutions could be
- Give error in `garage-sign` if added target-version pair is already present.
OR
- Give error on the backend in the similar case
OR
- Accept the image, but show that the device is not running it.
OR
- Accept the image and immediately schedule it for installation on the device. This one will probably work automatically if we finally fix director targets metadata to represent device state instead of installation order.

patrickvacek:
> As I can see possible solutions could be...

I think it should be an error, so options 1 or 2 would be my preference. #3 is okay but confusing. #4 seems like a problem. Another solution: identify the problem and automatically rename the target (append `_1` or something like that). I think I'd prefer an error over that, but it's an option.

OYTIS:
@patrickvacek Fine for me, but I'm pretty sure some customers will be annoyed with the error. Another option could be `y/n` prompt, but we can't support it in bitbake.

patrickvacek:
@OYTIS Maybe throw a loud warning and tell the user that we are changing the name to something unique (i.e. by appending something like `_1`) so as to not overwrite what is on the server? Is it not an error to upload two different things with the same name? If it isn't, then we have to figure out how to disambiguate them somehow, and that seems tough.

OYTIS:
@patrickvacek I don't think that overriding a version is an error, but it's definitely risky. What we can from meta-updater is giving a warning every time user overrides target version, just because it's so easy to break something there. All the other possible checks belong to garage-sign in my opinion.
I've also filed a bug to backend/frontend for hash mismatches: https://saeljira.it.here.com/browse/OTA-1240

patrickvacek:
@OYTIS Good call on the ticket for hash mismatch. We should also maybe make a ticket to cover garage-sign or other backend checks for the target version override. So for now let's add a warning on our side and then we can call this PR ready. Should we perhaps also merge this PR with the other rocko PRs at the same time?

patrickvacek:
I forewent merging this with the big rocko PR since I still think we need the warning. But now it should probably be rebased, too. (Maybe that should be in the checklist?)

OYTIS:
@patrickvacek Added warnings, docs, and rebased. Good point on the checklist.

README.adoc
```diff
@@ -167,6 +167,16 @@ There are a few settings that can be controlled in `local.conf` to simplify the
...
+
+Every time you build an image with `SOTA_PACKED_CREDENTIALS` set, a new entry in your Uptane metadata is created and you can see it in the OTA Garage UI if you're using one. Normally this version will be equal to OSTree hash of your root file system. If you want it to be different though you can override is using one of two methods:
+
+1. Set `GARAGE_TARGET_VERSION` variable in your `local.conf`.
+2. Write a recipe or a bbclass to write the desired version to `${STAGING_DATADIR_NATIVE}/target_version`. An example of such bbclass can be found in `classes/target_version_example.bbclass`.
+
+Please note that [target name, target version] pairs are expected to be unique in the system. If you build a new target with the same target version as a previously build one, the old package will be overwritten on the update server. It can have unpredictable effect on devices that have this version installed, it is not guaranteed that information will be reported correctly for such devices or that you will be able to update them (we're doing our best though). The easiest way to avoid problems is to make sure that your overriding version is as unique as OSTree commit hash.
+
```
patrickvacek: Minor suggestions:
1. "as a previously *built* one"
2. "is as unique as *an* OSTree commit hash."
3. "that have this version installed, *and* it is not guaranteed"

patrickvacek:
oe-selftest passes.

README.adoc
OYTIS: Thank you, fixed.

EpicNuts:
Looks good to me too, the only caveat I have is that the new image is reported as 'installed' when the page is refreshed, which it most certainly isn't.
Overwriting the package to a new version, even if it's not installed on any device, doesn't trigger the automatic installation, which may or may not be a good thing, but either way that's probably a front end issue.

PR #383, opened 2018-09-05
Add SOTA_DEPLOY_CREDENTIALS variable to control whether the built image should be provisioned

OYTIS:
This PR wasn't tested at all, the testing will include going through manual provisioning which will different for each scenario. Maybe you will have some comments before the testing though.


lbonn:
Looks good so far.
Minor cosmetics though: taking a quick look at [bitbake variables](https://www.yoctoproject.org/docs/2.5/ref-manual/ref-manual.html#ref-variables-glossary), the favored way to treat boolean flags seems to be `0/1`.

OYTIS:
@lbonn You're right, thanks.

OYTIS:
OK, alternative solution. Credentials themselves are now in separate recipes. It makes the whole feature a bit easier to test (but it's still not tested).

OYTIS:
@patrickvacek You're right, I've completely missed the symmetry in the resulting recipes. Let's keep them as they are for a while (before merging this PR), maybe I'll find a more fine split for implicit provisioning and then having separate recipes will make sense, otherwise we can merge them into one.

OYTIS:
OK, I've tested everything and it seems to work in all usage scenarios which are now just
- Autoprovisioning
- Implicit provisioning with credentials in storage
- Implicit provisioning with credentials in PKCS#11 compatible device

recipes-support/softhsm-testtoken/files/createtoken.sh
```diff
@@ -13,7 +13,8 @@ fi
 mkdir -p /var/lib/softhsm/tokens
 softhsm2-util --init-token --slot 0 --label "Virtual token" --pin 1234 --so-pin 1234
 
-softhsm2-util --import /var/sota/import/pkey.pem --label "pkey" --id 02 --token 'Virtual token' --pin 1234
+openssl pkcs8 -topk8 -inform PEM -outform PEM -nocrypt -in /var/sota/import/pkey.pem -out /var/sota/import/pkey.p8
+softhsm2-util --import /var/sota/import/pkey.p8 --label "pkey" --id 02 --token 'Virtual token' --pin 1234
```
patrickvacek: What was the story here? I can't tell why this change was necessary.

recipes-sota/aktualizr/aktualizr_git.bb
```diff
@@ -11,11 +11,12 @@ DEPENDS_append_class-native = "glib-2.0-native "
 
 RDEPENDS_${PN}_class-target = "lshw "
 RDEPENDS_${PN}_append_class-target = "${@bb.utils.contains('SOTA_CLIENT_FEATURES', 'serialcan', ' slcand-start', '', d)} "
-RDEPENDS_${PN}_append_class-target = "${@bb.utils.contains('SOTA_CLIENT_FEATURES', 'hsm', ' softhsm softhsm-testtoken', '', d)}"
 
 RDEPENDS_${PN}_append_class-target = " ${PN}-tools "
 RDEPENDS_${PN}-secondary_append_class-target = " ${PN}-tools "
 
+RDEPENDS_${PN}_class-target = " openssl-bin "
```
patrickvacek: Does aktualizr require `openssl-bin`, or just the testtoken script? I added the dependency there, but didn't realize aktualizr might need it, too.

recipes-sota/aktualizr/aktualizr-auto-prov.bb
```diff
@@ -6,7 +6,8 @@ LICENSE = "MPL-2.0"
 LIC_FILES_CHKSUM = "file://${COMMON_LICENSE_DIR}/MPL-2.0;md5=815ca599c9df247a0c7f619bab123dad"
 
 DEPENDS = "aktualizr-native zip-native"
-RDEPENDS_${PN} = "aktualizr"
+RDEPENDS_${PN}_append = "${@' aktualizr-auto-prov-creds' if d.getVar('SOTA_DEPLOY_CREDENTIALS', True) == '1' else ''}"
+RDEPENDS_${PN} = " aktualizr"
```
patrickvacek: Do the other prov recipes also need the explicit `RDEPENDS` on `aktualizr`? Might not hurt just to be consistent.

README.adoc
```diff
@@ -197,3 +198,16 @@ oe-selftest --run-tests updater
 ```
 
 For more information about oe-selftest, including details about how to run individual test modules or classes, please refer to the https://wiki.yoctoproject.org/wiki/Oe-selftest[Yocto Project wiki].
+
+== Manual provisoning
+As described in <<sota-related-variables-in-localconf,SOTA-related variables in local.conf>> section you can set `SOTA_DEPLOY_CREDENTIALS` to `0` to prevent deploying credentials to the built `wic` image. In this case you get a generic image that you can use e.g. on a production line to flash a series of devices. The cost of this approach is that this image is half-backed and should be provisioned before it can connect to the backend.
```
patrickvacek: Sorry to nitpick, but it's "half-baked". :)

README.adoc
```diff
@@ -197,3 +198,16 @@ oe-selftest --run-tests updater
...
+
+Provisioning procedure depends on your provisioning recipe, i.e. the value of `SOTA_CLIENT_PROV` (equal to `aktualizr-auto-prov` by default).
+* For `aktualizr-auto-prov` put your `credentials.zip` to `/ostree/deploy/poky/var/sota/sota_provisioning_credentials.zip`
+* For `aktualizr-ca-implicit-prov`
+** put URL to the used backend (together with protocol prefix and port number) to `/var/sota/gateway.url`. If you're using HERE OTA Connect, you can find the URL in `autoprov.url` file in your credentials archive.
+** put client certificate, private key and root CA certificate (for the *server*, not for the *device*) to `/var/sota/import_client.pem`, `/var/sota/import_pkey.pem` and `/var/sota/import_root.crt` respectively.
+* For  `aktualizr-hsm-prov`
+** put URL to the used backend (together with protocol prefix and port number) to `/var/sota/gateway.url`. If you're using HERE OTA Connect, you can find the URL in `autoprov.url` file in your credentials archive.
+** put root CA certificate (for the *server*, not for the *device*) to `/var/sota/import_root.crt`.
+** put client certificate and private key to slots 1 and 2 of the PKCS#11-compatible device.
```
patrickvacek: Nice write-up. Is the `/ostree/deploy/poky` prefix really necessary for the autoprov bit? And shouldn't the import paths be `/var/sota/import/xyx` instead of `/var/sota/import_xyz`?

README.adoc
```diff
@@ -80,7 +80,8 @@ Although we have used U-Boot so far, other boot loaders can be configured work w
...
+* `SOTA_DEPLOY_CREDENTIALS` - when set to '1' (default value), deploys credentials to the built image. Override it in `local.conf` to built a generic image that can be provisioned manually after the build.
+* `SOTA_CLIENT_PROV` - which provisioning method to use. Valid options are https://github.com/advancedtelematic/aktualizr/blob/master/docs/automatic-provisioning.adoc[`aktualizr-auto-prov`], https://github.com/advancedtelematic/aktualizr/blob/master/docs/implicit-provisioning.adoc[`aktualizr-ca-implicit-prov`], and https://github.com/advancedtelematic/aktualizr/blob/master/docs/hsm-provisioning.adoc[`aktualizr-hsm-prov`]. The default is `aktualizr-auto-prov`. This can also be set to an empty string to avoid using a provisioning recipe.
```
patrickvacek: How concerned are you with us eliminating the non-CA implicit prov recipe? I suspect not many people are using it, so asking them to switch is not a big problem. Do you foresee any problems with that transition?

recipes-support/softhsm-testtoken/files/createtoken.sh
OYTIS: On master `shofthsm2-util --import` would fail on pkcs#1 keys. I'm not sure how it worked without this change, but it didn't work for me.

recipes-sota/aktualizr/aktualizr_git.bb
OYTIS: Sorry, that was just for debugging. Thanks!

recipes-sota/aktualizr/aktualizr-auto-prov.bb
OYTIS: Yes, you're right. But I'd rather delete if from here. I believe we've already had a similar discussion, and my point was that custom aktualizr-based projects might use the same provisioning recipes without pulling aktualizr itself. I can't remember if you had objections though.

README.adoc
OYTIS: Thanks!

README.adoc
OYTIS: CA implicit prov with `SOTA_DEPLOY_CREDENTIALS = '0'` is pretty much the same as the old 'plain' implicit prov (except it doesn't bind device to a specific server URL). I would inform our customers who use implicit provisioning (exactly one to my knowledge) that they should change the value of `SOTA_CLIENT_PROV` and that is mostly it.

README.adoc
OYTIS: > Is the `/ostree/deploy/poky` prefix really necessary for the autoprov bit?

Good point. I was describing how it is done with sd card mounted on the build machine, and there it is necessary. I should make it clear probably and also describe how it can be done with `scp` and friends.

> And shouldn't the import paths be /var/sota/import/xyx instead of /var/sota/import_xyz

Thanks, I forgot to update this part.

recipes-support/softhsm-testtoken/files/createtoken.sh
patrickvacek: Strange. I never had a problem. Was there something different about how you generate the keys with the implicit CA recipe? It's not a big deal, but it seems odd that we need the interim step.

recipes-sota/aktualizr/aktualizr-auto-prov.bb
patrickvacek: No objection, go ahead and delete it if you'd like.

README.adoc
patrickvacek: Sounds like a plan, should be fine. That said, it looks like oe-selftest is failing as a result. I suspect you'll need to make some changes to the Implicit, Hsm, and AktualizrTools tests.

recipes-support/softhsm-testtoken/files/createtoken.sh
OYTIS: I can't see any differences. It might be because of different openssl versions. Maybe. But it looks weird that it worked before, because we've already had exactly the same problems with Uptane keys (see below in the same file).

patrickvacek: I'd like to understand the differences, but that's something we can come back to at a later time. Not worth holding back the PR just for that.

OYTIS:
oe-selftest should be fine now. Note that aktualizr recipe is still pointing to a PR.

OYTIS:
@patrickvacek 

> One larger-scope question, though: we will still be able to use the old implicit_writer/cert_provider hacky method to provision implicitly with or without an HSM?

Sorry, missed that part. Yes, if you set
```
SOTA_CLIENT_PROV = "aktualizr-ca-implicit-prov"
SOTA_DEPLOY_CREDENTIALS = "0"
```
You should be able to manually provision the device using any method you like including `cert-provider`.

patrickvacek:
> oe-selftest should be fine now.

I'm still seeing problems with the HSM and implicit tests. It builds fine but the devices don't provision.

patrickvacek:
> you should be able to manually provision the device using any method you like including `cert-provider`.

Excellent, yes, it worked for me when I manually ran the test. I hadn't seen the `-u` flag before, and that might be at the root of why the test are failing.

OYTIS:
OK, seems to be working with the latest fixes.

patrickvacek:
@OYTIS Sorry, looks like there is a merge conflict now.

OYTIS:
@patrickvacek Thanks! Rebased. Could you retest?

PR #384, opened 2018-09-06
Ci/oe selftest

lbonn:
Looks like I forgot to merge this and left the job running on this old branch (just rebased)

### DIFF
--- a/scripts/ci/Dockerfile.bitbake
+++ b/scripts/ci/Dockerfile.bitbake
@@ -31,6 +31,11 @@ RUN apt-get update -q && apt-get install -qy \
      xterm \
      xz-utils
 
+ARG uid=1000
+ARG gid=1000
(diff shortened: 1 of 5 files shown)


scripts/ci/Dockerfile.bitbake
```diff
@@ -18,6 +18,7 @@ RUN apt-get update -q && apt-get install -qy \
      libpython-dev \
      libsdl1.2-dev \
      locales \
+     ovmf \
```
OYTIS: Cool, are we testing Minnowboard builds now?

lbonn: It's running the full oe-selftests updater tests, so yes. But it's long and fails a lot :) and only on master right now.

I'll merge when the tests I launched today will reach a reasonable point (success or near success).

PR #386, opened 2018-09-14
libp11: Fix paths after switch to openssl 1.1.

patrickvacek:
Note that the HSM test still fails due to some issue with openct that I haven't gotten to the bottom of yet.


lbonn:
Thanks!

Note that it could be a good time to bump the version: https://github.com/OpenSC/libp11/releases/tag/libp11-0.4.9

0.4.8 has this line in the changelog: "Added support for OpenSSL 1.1.1 beta" and 0.4.9 has a memory leak fix in our code path (Openssl >= 1.1).

I don't know if it would make sense to also keep supporting old openssl. Maybe if we want to ease future backports to rocko & friends...

patrickvacek:
@lbonn good suggestions. I'll look into both.

patrickvacek:
Bumped libp11 version and hopefully solved the openssl version compatibility issue. Even after trying a few fixes for openct, things still aren't working, though.

patrickvacek:
Looks like the problem is that the openssl binary isn't getting installed on the device anymore.

PR #390, opened 2018-09-14
Update `installed_versions` passing to new location

lbonn:
Will need https://github.com/advancedtelematic/aktualizr/pull/945 to be merged and aktualizr bumped (bump should be pushed later in this PR).

### DIFF
--- a/classes/image_types_ota.bbclass
+++ b/classes/image_types_ota.bbclass
@@ -103,7 +103,7 @@ fakeroot do_otasetup () {
 
 	cp -a ${IMAGE_ROOTFS}/var/sota ${OTA_SYSROOT}/ostree/deploy/${OSTREE_OSNAME}/var/ || true
 	# Create /var/sota if it doesn't exist yet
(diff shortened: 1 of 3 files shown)


patrickvacek:
We ought to be able to remove the `--allow-migrations` hack from oe-selftest, too. That will be the real test.

lbonn:
Will it actually be enough? Two executables can try to create the empty sql database at the same time.
I'll maybe start that in another PR, hopefully with a viable testing strategy.

patrickvacek:
> Will it actually be enough? Two executables can try to create the empty sql database at the same time.

It will be enough for now. Realistically there should not be two applications running that could create the database simultaneously. It could happen with `aktualizr-info`, but not if we aren't using the `--allow-migrations` flag.

oe-selftest looked fine, although I'll need to test again once you are ready. But so far you are on the right track.

lbonn:
I've re-run `updater.ManualControlTests.test_manual_running_mode_once` after the rebase which includes aktualizr's bump.

PR #396, opened 2018-09-18
Feat/import installed versions rocko

patrickvacek:
Just catching up so that we can close OTA-327.


patrickvacek:
Bumped aktualizr, fixed some license issues, and backported Anton's provisioning recipe work.

patrickvacek:
I've tested this with oe-selftest. I've done a full roundtrip update with a raspberry pi. I've verified that garage-deploy works as expected and fixes the issue with the root certificate and server URL getting updated in a problematic manner.

codders:
Looks good to me - thanks for testing this!

PR #398, opened 2018-10-02
Feat/thud compatibility

patrickvacek:


rsalveti:
Thud is now default in oe-core, so this merge request is now needed for it to be compatible with latest OE.

PR #406, opened 2018-10-10
Refactor/backport ostree upgrade rocko

patrickvacek:
Backporting ostree upgrades to rocko. Passes oe-selftest.


patrickvacek:
I've now also tested this manually end-to-end with a raspberry pi and it checks out. Any last objections before I merge this?

PR #407, opened 2018-10-12
Extend ostree support

LukaPivk:
Makes it easier to customize ostree commits


Jmennius:
@LukaPivk Looks good, but I would go further and by adding a way to provide arbitrary arguments.
Users would be able to set other options and more importantly - metadata w/o touching `image_types_ostree.bbclass`.

tkfu:
Can you also update the [readme](https://github.com/advancedtelematic/meta-updater/#sota-related-variables-in-localconf) to document the three new local.conf options you're introducing?

LukaPivk:
I signed this now

lbonn:
Thanks @LukaPivk and sorry for the delay. I've run some tests and it looks good. Just two minor bitbake style comments, aimed at making the .bbclass a bit more robust and self-documenting:

- `OSTREE_COMMIT_BODY` defined `??= ""` along with `OSTREE_COMMIT_SUBJECT`
- Using 0 and 1 for disabling and enabling `OSTREE_UPDATE_SUMMARY`: defined `??= "0"` also at the top of the file, the check changed to ` =  1` and docs updated. It would then follow the usual bitbake conventions for boolean variables

lbonn:
Great, thanks!

PR #409, opened 2018-10-17
Fix formatting in README

OYTIS:


patrickvacek:
I'd fixed this for master (https://github.com/advancedtelematic/meta-updater/pull/405) but hadn't backported yet. Looks like it is time for another round!

PR #410, opened 2018-10-18
Use IMGDEPLOYDIR

agners:
do_image tasks should not access `DEPLOY_DIR_IMAGE` directly. The `IMGDEPLOYDIR` is properly sstate managed and allows to use the symlinking feature of OpenEmbedded.

To make use some of the features the file ending need to match the image fstype name. That requires us to rename `.otaimg` to `ota-ext4`. This also makes sure that the image is named such that it can be extended to support other file systems etc.


OYTIS:
Thank you, @agners 
oe-selftest passes according to @patrickvacek. Merging.

PR #411, opened 2018-10-25
Update aktualizr example secondary config path

lbonn:
After https://github.com/advancedtelematic/aktualizr/pull/973

tested: built core-image-minimal for qemu


patrickvacek:
aktualizr fails for me with `Could not read secondary configs from "/usr/lib/sota/secondaries": not a directory`. Looks like a bug in aktualizr, sorry.

lbonn:
@patrickvacek thanks! My test was wrong because I kept `aktualizr-examples` in `IMAGES_INSTALL_append`. I believe it's only a problem with the recipe, thankfully.

lbonn:
Added a cleanup commit which depends on https://github.com/advancedtelematic/aktualizr/pull/977

PR #414, opened 2018-10-29
Refactor/sumo/rm implicit writer

patrickvacek:
Backporting recent changes to master (documentation, external secondary and implicit_writer removal, secondary configuration, libgsystem removal). Does not include changes to image_types_ota.


patrickvacek:
oe-selftest passes.

PR #418, opened 2018-10-31
Update packages removed for secondary-image.

patrickvacek:


patrickvacek:
Passes oe-selftest.

recipes-test/images/secondary-image.bb
```diff
@@ -10,9 +10,10 @@ LICENSE = "MIT"
 IMAGE_INSTALL_remove = " \
                         aktualizr \
                         aktualizr-auto-prov \
+                        aktualizr-auto-prov-creds \
                         aktualizr-ca-implicit-prov \
+                        aktualizr-ca-implicit-prov-creds \
```
lbonn: Also `aktualizr-uboot-env-rollback`?

patrickvacek: You're probably right. Fixed.

patrickvacek: `oe-selftest -r updater.SecondaryTests` passed after the fix.

Issue #412, opened 2018-10-26
Locking repo exclusive failed: Resource temporarily unavailable


agners:
We try to store two (similar) images into the same OSTree repository. The thought is that we could switch between the two using OSTree (which works surprisingly well). On our CI infrastructure however it seems that this can cause race conditions when two independent `bitbake` instances use the same `OSTREE_REPO`. Maybe the `do_image_ostree` needs some higher level locking mechanism which makes sure the task gets exclusive access to the OSTree repo?

```
ERROR: torizon-core-docker-0.2.26-r0 do_image_ostree: Function failed: do_image_ostree (log file is located at /workdir/oe/tmp/work/colibri_imx7-torizon-linux-gnueabi/torizon-core-docker/0.2.26-r0/temp/log.do_image_ostree.1720)
ERROR: Logfile of failure stored in: /workdir/oe/tmp/work/colibri_imx7-torizon-linux-gnueabi/torizon-core-docker/0.2.26-r0/temp/log.do_image_ostree.1720
Log data follows:
| DEBUG: Executing python function set_image_size
| DEBUG: 844146.000000 = 367020 * 2.300000
| DEBUG: 844146.000000 = max(844146.000000, 65536)[844146.000000] + 2
| DEBUG: 844146.000000 = int(844146.000000)
| DEBUG: 844146 = aligned(844146)
| DEBUG: returning 844146
| DEBUG: Python function set_image_size finished
| DEBUG: Executing python function extend_recipe_sysroot
[...]

PR #421, opened 2018-11-01
classes: image_types_ostree: use locking during OSTree repo update

agners:
The OSTree repository might be shared accross several individual
OSTree builders. Use bitbakes lockfiles mechanism to lock the
OSTree for exclusive use.


agners:
Actually that causes an issue when building the first time: If $OSTREE_REPO does not exist, lockfiles will create it. However, the `do_image_ostree` task uses the directory to detect whether to initialize an OSTree repo:

     if [ ! -d ${OSTREE_REPO} ]; then
         ostree --repo=${OSTREE_REPO} init --mode=archive-z2
     fi

This is not ideal anyway, I think we should use some better method to detect whether there is a OSTree repo already.

patrickvacek:
@agners Nice, I like the look of this! I'm testing it now.

Fixes #412.

OYTIS:
@agners Thanks, looks great. Do you know if this mechanism is documented somewhere?

patrickvacek:
Passes oe-selftest, I'm just checking things once more before merging.

lbonn:
@OYTIS from the mailing list thread: https://www.yoctoproject.org/docs/latest/bitbake-user-manual/bitbake-user-manual.html#variable-flags

PR #422, opened 2018-11-05
Added support to set custom commit subject and body

LM555:
Use OSTREE_COMMIT_BODY write custom text to ostree body
Use OSTREE_COMMIT_SUBJECT to write custom subject

Signed-off-by: Luka Pivk <luka.pivk@toradex.com>


patrickvacek:
@LM555 Do you just want to backport this one commit, or are there other things you think you'll want as well? I normally only bother to backport changes that are necessary for our work, but I'm certainly not opposed to backporting anything else that is useful. Anyway, this looks fine. I'm running through oe-selftest now.

LM555:
I just found this single commit to be really useful to us. Maybe it would be good to update the readme as well (2af450bc64021c2ebc2e82a59a209e56276bad4f). Thank you!

patrickvacek:
@LM555 Yeah, actually, if you could update the readme as well, that would be great.

LM555:
Added the 2. commit as well now, so the readme should now be up-to-date.

LM555:
This also added the "update summary" feature btw.

PR #423, opened 2018-11-06
Update README.adoc

xaviergully:
Update the reference to https://connect.ota.here.com


patrickvacek:
@xaviergully I see at least three other references to "ATS Garage" when I grep the repo. We presumably should change them all.

I also see a reference in `meta-updater-raspberrypi/scripts/flash-image.sh`.

xaviergully:
Yes makes sense

patrickvacek:
@xaviergully Updated to catch more references. See also https://github.com/advancedtelematic/meta-updater-raspberrypi/pull/33.

PR #425, opened 2018-11-07
Add the requirement to sign a DCO

OYTIS:
Not the best wording, mostly just testing the bot.


CONTRIBUTING.adoc
```diff
@@ -13,6 +13,8 @@ Previously, some older branches were also regularly supported, and while they sh
 
 If you are developing with meta-updater, it may be helpful to read the README and other documentation for link:README.adoc[this repo], https://github.com/advancedtelematic/aktualizr[aktualizr], and the https://github.com/advancedtelematic/updater-repo/[updater-repo], particularly the sections about development and debugging.
 
+Please sign a Developer Certificate of Origin (`--signoff` when committing) for all of your commits.
+
```
patrickvacek: Could we maybe provide a link to the docs in the aktualizr that relate to this? No need to duplicate all the text (I think), but it's worth making the intentions and requirements clear.

OYTIS: I've just changed the text to what is in aktualizr, linking seems a bit non-reliable to me.

PR #427, opened 2018-11-09
Fix/thud/newline etc

patrickvacek:


patrickvacek:
Passes oe-selftest with the new commits.

PR #428, opened 2018-11-09
Fix/sumo/newline etc

patrickvacek:
Passes oe-selftest.


patrickvacek:
Passes oe-selftest with the new commits.

PR #430, opened 2018-11-12
Fixed some issues

liuming50:


classes/sota.bbclass
```diff
@@ -16,7 +16,8 @@ IMAGE_INSTALL_append_sota = " ostree os-release ${SOTA_CLIENT} ${SOTA_CLIENT_PRO
 IMAGE_CLASSES += " image_types_ostree image_types_ota"
 
 IMAGE_FSTYPES += "${@bb.utils.contains('DISTRO_FEATURES', 'sota', 'ostreepush garagesign garagecheck ota-ext4 wic', ' ', d)}"
-IMAGE_FSTYPES += "${@bb.utils.contains('BUILD_OTA_TARBALL', '1', 'ota-tar ota-tar.xz', ' ', d)}"
+IMAGE_FSTYPES += "${@bb.utils.contains('BUILD_OSTREE_TARBALL', '1', 'ostreetar', ' ', d)}"
```
agners: Since BUILD_OSTREE_TARBALL is not set anywhere, there won't be a tar.bz2 anymore by default. This changes behavior... That might be fine, but we should be aware of the fact...

There are some self tests, and one seems to be using tar.bz2... Are the self tests still passing?
See lib/oeqa/selftest/cases/updater.py

Maybe just set BUILD_OSTREE_TARBALL to 1 in the test, or somewhere as a default for now?

classes/image_types_ostree.bbclass
```diff
@@ -145,16 +130,25 @@ IMAGE_CMD_ostree () {
 
     # Copy image manifest
     cat ${IMAGE_MANIFEST} | cut -d " " -f1,3 > usr/package.manifest
+}
 
-    cd ${WORKDIR}
+IMAGE_TYPEDEP_ostreetar = "ostreerootfs"
+IMAGE_CMD_ostreetar () {
+    ${IMAGE_CMD_TAR} -C ${OSTREE_ROOTFS} --numeric-owner --xattrs --xattrs-include='*' -cjf ${IMGDEPLOYDIR}/${IMAGE_NAME}${IMAGE_NAME_SUFFIX}${OSTREE_TAR_SUFFIX} . || [ $? -eq 1 ]
+    ln -s ${IMAGE_NAME}${IMAGE_NAME_SUFFIX}${OSTREE_TAR_SUFFIX} ${IMGDEPLOYDIR}/${IMAGE_LINK_NAME}${IMAGE_NAME_SUFFIX}${OSTREE_TAR_SUFFIX}
```
agners: One small more comment: Can we not rely on `create_symlinks`?

patrickvacek:
Could you also rebase this branch on the latest master, please?

classes/image_types_ostree.bbclass
```diff
@@ -175,7 +173,7 @@ IMAGE_CMD_ostree () {
 }
 
 IMAGE_TYPEDEP_ostreepush = "ostree"
-do_image_ostreepush[depends] += "aktualizr-native:do_populate_sysroot ca-certificates-native:do_populate_sysroot"
+do_image_ostreepush[depends] += "aktualizr-native:do_populate_sysroot ca-certificates-native:do_populate_sysroot unzip-native:do_populate_sysroot"
```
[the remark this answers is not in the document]
liuming50: Hi, Stefan:

I double checked the master branch, the tip is: 8c72fd4555dc5637316564895ec3271b40389497
[ Merge pull request #431 from albard00/master ]

in classes/image_types_ostree.bbclass:
...
IMAGE_CMD_garagesign () {
...
unzip -p ${SOTA_PACKED_CREDENTIALS} root.json targets.pub targets.sec tufrepo.url 2>&1 >/dev/null || exit 0
...
}

classes/image_types_ostree.bbclass
```diff
@@ -16,25 +9,20 @@ OSTREE_KERNEL ??= "${KERNEL_IMAGETYPE}"
...
 
-IMAGE_CMD_ostree () {
-    if [ -z "$OSTREE_REPO" ]; then
-        bbfatal "OSTREE_REPO should be set in your local.conf"
-    fi
-
-    if [ -z "$OSTREE_BRANCHNAME" ]; then
-        bbfatal "OSTREE_BRANCHNAME should be set in your local.conf"
-    fi
+export SYSTEMD_USED = "${@oe.utils.ifelse(d.getVar('VIRTUAL-RUNTIME_init_manager', True) == 'systemd', 'true', '')}"
```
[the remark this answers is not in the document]
liuming50: I guess it's needed to be exported so far since it's being referred currently in IMAGE_CMD_ostree task as a shell variable, as follows:
...
if [ -n "$SYSTEMD_USED" ]; then
...

if we intend to remove the export syntax from it, then we need change the reference to be a bitbake variable expanding syntax as ${SYSTEMD_USED}.

I guess the author might has his reason to export it, but I can not tell by myself.

//Ming Liu

classes/sota.bbclass
liuming50: Yes, I think you are absolutely correct, it will break the test if it does not set BUILD_OSTREE_TARBALL to 1, I will send a V2 to fix this. Thanks for the review.

//Ming Liu

patrickvacek:
@liuming50 Great, thanks. I ran oe-selftest and got a few failures, although I'm not really sure why. I'll wait for your fixes before I try again and dig in deeper.

classes/image_types_ostree.bbclass
liuming50: Sorry, I think I probably mis-understood your comment, did you suggest we use subimages instead of "ln -s"?

agners: No I meant using the postfunc create_symlink. But yeah, it probably would require renaming... So maybe better leave as is, at least for now...

classes/image_types_ostree.bbclass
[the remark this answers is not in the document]
liuming50: Hi, Stefan:

You probably asked if the OSTREE_ROOTFS variable needs to be exported or not. I probably mis-understood your meaning.

If so, I agreed, it does not have to be exported, will fix it in V2.

liuming50:
Changes in V2:
1 Rebase to the latest tip: 8c72fd4555dc5637316564895ec3271b40389497 [ Merge pull request #431 from albard00/master ]
2 Do not export OSTREE_ROOTFS as suggested by Stefan.
3 Add unzip-native:do_populate_sysroot dependency to do_image_garagesign instead of sdo_image_ostreepush as suggested by Stefan, sorry, I mis-undertood your comments.
4 BUILD_OSTREE_TARBALL defaults to 1, to be consistent with the original behavior, to avoid breaking things depending on the generated xxx.ostree.tar.bz2 file.

classes/image_types_ostree.bbclass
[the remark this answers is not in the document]
patrickvacek: > However, I don't think that any subprocess or script uses `SYSTEMD_USED`... So I think we do not need to export that.

We used to use export a lot more than necessary. I'm not sure why, to be honest. It is quite likely that this particular example is also not necessary to export.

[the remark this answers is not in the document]
liuming50: I am on board with Agners, it looks like not need to be exported, I have just send a V3 including a patch to remove it, please kindly help me review.

liuming50:
Changes in V3:
1 Add one more patch to avoid exporting SYSTEMD_USED.

agners:
V3 looks good to me! Also did a quick test for our use case here. Thanks Ming!

There is one more thing which I think we could optimize: `cp -a` actually makes a copy of the complete rootfs, which is time consuming and expensive. OpenEmbedded uses hardlinks in a lot of places now...

Is it safe to use hardlinks here too? I guess we are not supposed to change files if hardlinks are used, since it would change the file in the regular rootfs tree...

I am also not entirely sure how to do it properly in a shell script. Maybe it is ok to just use `cp -al`? OE often uses the Python function `oe.path.copyhardlinktree`.... `openembedded-core/meta/recipes-kernel/perf/perf.bb` uses a Python function as a prefuncs maybe that would do in this case too? 

In any case, even if we do that, I suggest to do this later/in a separate pull request.

liuming50:
That's a good suggestion, we can keep a eye on this optimizing part.

//Ming Liu

Stefan Agner <notifications@github.com> 於 2018年11月14日 週三 下午5:22寫道：

liuming50:
Hi, Patrick:

Would you please share with me the steps about how you did the autotest? I
would like to reproduce it and investigate the failures.

the best,
thank you

Patrick Vacek <notifications@github.com> 於 2018年11月15日 週四 下午2:37寫道：

agners:
@liuming50 there is CONTRIBUTING.adoc which I think should document the autotest.

liuming50:
HI, Stefan:

Thanks, will read and try the tests.

the best,
thank you

Stefan Agner <notifications@github.com> 於 2018年11月15日 週四 下午4:07寫道：

patrickvacek:
@liuming50 This could be related to the fact that in `do_image_ostree` (now `do_image_ostreerootfs`), we remove everything in `/var` except for `/var/local`. The goal is that we don't want (most of) `/var` uploaded to treehub, but we want `/var/sota` to be part of the image that we flash on the device initially. Since the dependencies have been changed, possibly something is happening out of order? Seems unlikely, because I thought the image generation happened elsewhere, but I'm just trying to imagine what is going wrong.

classes/image_types_ostree.bbclass
```diff
@@ -1,42 +1,29 @@
...
 
-    if [ -z "$OSTREE_BRANCHNAME" ]; then
-        bbfatal "OSTREE_BRANCHNAME should be set in your local.conf"
-    fi
+BUILD_OSTREE_TARBALL ??= "1"
 
-    OSTREE_ROOTFS=`mktemp -du ${WORKDIR}/ostree-root-XXXXX`
-    cp -a ${IMAGE_ROOTFS} ${OSTREE_ROOTFS}
+do_image_ostreerootfs[dirs] = "${OSTREE_ROOTFS}"
+do_image_ostreerootfs[cleandirs] = "${OSTREE_ROOTFS}"
```
patrickvacek: Will this work to have `OSTREE_ROOTFS` be created and remove with the `ostreerootfs` task if it is used again in the `ostreetar` and `ostree` tasks?

classes/image_types_ostree.bbclass
```diff
@@ -1,42 +1,29 @@
...
+do_image_ostreerootfs[depends] = "coreutils-native:do_populate_sysroot virtual/kernel:do_deploy ${INITRAMFS_IMAGE}:do_image_complete"
+IMAGE_CMD_ostreerootfs () {
+    cp -a ${IMAGE_ROOTFS}/* ${OSTREE_ROOTFS}
     chmod a+rx ${OSTREE_ROOTFS}
     sync
 
-    cd ${OSTREE_ROOTFS}
-
```
patrickvacek: If we don't do this `cd`, I don't think any of the subsequent commands will work as expected.

classes/sota.bbclass
```diff
@@ -16,7 +16,8 @@ IMAGE_INSTALL_append_sota = " ostree os-release ${SOTA_CLIENT} ${SOTA_CLIENT_PRO
...
+IMAGE_FSTYPES += "${@bb.utils.contains('BUILD_OTA_TARBALL', '1', 'ota-tar.xz', ' ', d)}"
```
patrickvacek: Why was `ota-tar` removed? If it really is redundant, then we should get rid of the that task in `classes/image_types_ota.bbclass`.

liuming50:
+do_image_ostreerootfs[cleandirs] = "${OSTREE_ROOTFS}"

> Will this work to have OSTREE_ROOTFS be created and remove with the
ostreerootfs task if it is used again in the ostreetar and ostree tasks?

It would be removed, but only in case ostreerootfs task needs re-run, in
which case, bitbake will clean that directory before it executes ostreerootfs
task, but after ostreerootfs is executed, the directory would remain until
next ostreerootfs run. And since the ostreetar, ostree are depending on
ostreerootfs, so once ostreerootfs's hash changes and a re-running is
triggered, the ostreetar, ostree would be run as well, which should be no
problem.


> -    cd ${OSTREE_ROOTFS}

> If we don't do this cd, I don't think any of the subsequent commands will
work as expected.

This line is not needed because we set "do_image_ostreerootfs[dirs] =
"${OSTREE_ROOTFS}"", this will lead bitbake execute "mkdir and cd"
${OSTREE_ROOTFS}.


+IMAGE_FSTYPES += "${@bb.utils.contains('BUILD_OTA_TARBALL', '1',
'ota-tar.xz', ' ', d)}"

> Why was ota-tar removed? If it really is redundant, then we should get
rid of the that task in classes/image_types_ota.bbclass.

because the conversion task ota-tar.xz is depending on ota-tar, and this is
set in image.bbclass, if you add "ota-tar.xz" to IMAGE_FSTYPES, then ota-tar
does not have to be there, otherwise, it's redundant, for instance, when
you wanna add "tar.bz2" to IMAGE_FSTYPES, you do not have to add "tar
tar.bz2", same reason.


I would like to verify my patches by running "oe-selftest -r updater", but
seems most of the test cases are relying on a ota server running, may I
know how can I set up a server? Is there any instructions about that? I
tried before set up ota-community-edition, I can do that again, but I am
wondering maybe there is a easier way to run the tests, maybe?

//Ming Liu




Patrick Vacek <notifications@github.com> 於 2018年11月15日 週四 下午4:58寫道：

patrickvacek:
Thanks for your explanations! Some of these features were new to me and I couldn't find documentation for them, so I wasn't sure what was happening.

> because the conversion task ota-tar.xz is depending on ota-tar, and this is set in image.bbclass, if you add "ota-tar.xz" to IMAGE_FSTYPES, then ota-tar does not have to be there

That part I still don't understand, and my apologies if I'm missing something obvious. But I don't see any connection between the `ota-tar` task and the `ota-tar.xz` object. Is there some bitbake magic making that connection behind the scenes?

> I am wondering maybe there is a easier way to run the tests, maybe?

Yes, I find the easiest way is to use our production SaaS server, [Here OTA Connect](https://app.atsgarage.com); it's reliable and free to use for demonstration/testing purposes like yours. You have to sign up for an account, but that should be fairly painless. You can use Github and Google accounts, too, at least for now.

The other option is to use the community edition, but it's a fair bit of work to get that up and running, and to be honest, I haven't used it much lately, and it is possible that oe-selftest would fail for other reasons that aren't related to your PR.

agners:
> > because the conversion task ota-tar.xz is depending on ota-tar, and this is set in image.bbclass, if you add "ota-tar.xz" to IMAGE_FSTYPES, then ota-tar does not have to be there
> 
> That part I still don't understand, and my apologies if I'm missing something obvious. But I don't see any connection between the `ota-tar` task and the `ota-tar.xz` object. Is there some bitbake magic making that connection behind the scenes?

This is OpenEmbedded magic. [image_types.bbclass](http://git.openembedded.org/openembedded-core/tree/meta/classes/image_types.bbclass#n12) treats dots special: It automatically causes `ota-tar.xz` to depend on the task which creates `ota-tar`. Also dots denote a conversions, and all conversions are available if the regular OpenEmbedded task naming is followed. This works since we use `IMAGE_CMD_xx` task which follow the image type name. Dashes also get a special treatment, so `IMAGE_CMD_ota-tar` gets converted to `do_image_ota_tar` in the end.

There is quite a bit magic around image types, and I found it most helpful to just read the `image_types.bbclass`.

patrickvacek:
> This is OpenEmbedded magic.

Good to know, thanks for clearing that up for me!

liuming50:
> Yes, I find the easiest way is to use our production SaaS server, Here
OTA Connect <https://app.atsgarage.com/>; it's reliable and free to use for
demonstration/testing purposes like yours. You have to sign up for an
account, but that should be fairly painless. You can use Github and Google
accounts, too, at least for now.

> The other option is to use the community edition, but it's a fair bit of
work to get that up and running, and to be honest, I haven't used it much
lately, and it is possible that oe-selftest would fail for other reasons
that aren't related to your PR

Thanks for the instructions, will try to do the test with the server set up.

//Ming Liu




Patrick Vacek <notifications@github.com> 於 2018年11月16日 週五 上午10:52寫道：

classes/image_types_ota.bbclass
```diff
@@ -96,19 +96,14 @@ fakeroot do_otasetup () {
 
 	ostree admin --sysroot=${OTA_SYSROOT} deploy ${kargs_list} --os=${OSTREE_OSNAME} ${ostree_target_hash}
 
-	# Copy deployment /home and /var/sota to sysroot
-	HOME_TMP=`mktemp -d ${WORKDIR}/home-tmp-XXXXX`
-
-	tar --xattrs --xattrs-include='*' -C ${HOME_TMP} -xf ${DEPLOY_DIR_IMAGE}/${IMAGE_LINK_NAME}.rootfs.ostree.tar.bz2 ./usr/homedirs ./var/local || true
-
-	cp -a ${IMAGE_ROOTFS}/var/sota ${OTA_SYSROOT}/ostree/deploy/${OSTREE_OSNAME}/var/ || true
+	cp -a ${OSTREE_ROOTFS}/var/sota ${OTA_SYSROOT}/ostree/deploy/${OSTREE_OSNAME}/var/ || true
```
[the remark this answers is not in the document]
patrickvacek: Ah-ha, that might be the problem, good call. If the AutoProv test works, the others probably will, too.

liuming50:
Hi, Stefan:

Thanks for finding that error, you are absolutely right, I also found that
issue today, and I am working on the V4 patch, actually it turns out to
have more refactoring of the image_types_ostree.bbclass and
image_types_ota.bbclass, will send a V4 soon.

//Ming Liu

Patrick Vacek <notifications@github.com> 於 2018年11月19日 週一 下午5:49寫道：

agners:
@liuming50 just encountered another error:

```
NOTE: recipe core-image-minimal-1.0-r0: task do_image_ostreepush: Started                                                                  
NOTE: recipe core-image-minimal-1.0-r0: task do_otasetup: Started
ERROR: core-image-minimal-1.0-r0 do_image_ostreetar: Function failed: do_image_ostreetar (log file is located at /home/ags/ats-oe/build/tmp/work/qemux86_64-poky-linux/core-image-minimal/1.0-r0/temp/log.do_image_ostreetar.57669)                                                    
ERROR: Logfile of failure stored in: /home/ags/ats-oe/build/tmp/work/qemux86_64-poky-linux/core-image-minimal/1.0-r0/temp/log.do_image_ostreetar.57669
Log data follows:
| DEBUG: Executing python function set_image_size
| DEBUG: 112522.800000 = 86556 * 1.300000
| DEBUG: 182154.800000 = max(112522.800000, 8192)[112522.800000] + 1
| DEBUG: 182155.000000 = int(182154.800000)
| DEBUG: 182155 = aligned(182155)
| DEBUG: returning 182155
| DEBUG: Python function set_image_size finished
| DEBUG: Executing shell function do_image_ostreetar
| ln: failed to create symbolic link '/home/ags/ats-oe/build/tmp/work/qemux86_64-poky-linux/core-image-minimal/1.0-r0/deploy-core-image-minimal-image-complete/core-image-minimal-qemux86-64.rootfs.ostree.tar.bz2': File exists                                                       
| WARNING: /home/ags/ats-oe/build/tmp/work/qemux86_64-poky-linux/core-image-minimal/1.0-r0/temp/run.do_image_ostreetar.57669:1 exit 1 from 'ln -s core-image-minimal-qemux86-64-20181119192749.rootfs.ostree.tar.bz2 /home/ags/ats-oe/build/tmp/work/qemux86_64-poky-linux/core-image-minimal/1.0-r0/deploy-core-image-minimal-image-complete/core-image-minimal-qemux86-64.rootfs.ostree.tar.bz2'                                 
| ERROR: Function failed: do_image_ostreetar (log file is located at /home/ags/ats-oe/build/tmp/work/qemux86_64-poky-linux/core-image-minimal/1.0-r0/temp/log.do_image_ostreetar.57669)
NOTE: recipe core-image-minimal-1.0-r0: task do_image_ostreetar: Failed                                                                    
ERROR: Task (/home/ags/ats-oe/poky/meta/recipes-core/images/core-image-minimal.bb:do_image_ostreetar) failed with exit code '1'            
NOTE: recipe core-image-minimal-1.0-r0: task do_image_ostreepush: Succeeded                                                                
NOTE: recipe core-image-minimal-1.0-r0: task do_otasetup: Succeeded
NOTE: Tasks Summary: Attempted 3612 tasks of which 3607 didn't need to be rerun and 1 failed.                                              

Summary: 1 task failed:
  /home/ags/ats-oe/poky/meta/recipes-core/images/core-image-minimal.bb:do_image_ostreetar                                                  
Summary: There was 1 ERROR message shown, returning a non-zero exit code.
```

I guess that needs a ln -s**f** in `IMAGE_CMD_ostreetar`.

liuming50:
Hi, Stefan:

Thanks for testing, I have made a V4 patch series, there will be more
refactoring in it, probably can fix this problem as well, I am now running
"oe-selftest -r updater", will send the V4 once it passes the selftest.

//Ming Liu

Stefan Agner <notifications@github.com> 於 2018年11月19日 週一 下午8:41寫道：

liuming50:
HI, Patrick/Stefan:

I have just sent a V4 patch, please kindly help me review, I am now doing
the oe-selftest with this new patch series, will let you know the results.

//Ming Liu

Ming Liu <liu.ming50@gmail.com> 於 2018年11月19日 週一 下午9:52寫道：

agners:
Just looked through it, the main change to the earlier versions seems to be that you are using now tar as a conversion type. I was thinking about that in the past too, but was a bit hesitant: There is `IMAGE_CMD_tar` as well... How does OE behaves if there is an image type as well as a conversion type?


E.g. we have now: `ota.tar.xz`. Is the first part always the base type, so `IMAGE_CMD_ota` must exist, and then every other extension is a conversion?

My fear is that OE picks up our tar conversion if you try to create a regular image, or vice-versa.

liuming50:
Hi, Stefan:

Good questions!

> How does OE behaves if there is an image type as well as a conversion
type?

OE can distinguish them well since the conversion tar command is:
CONVERSION_CMD_tar, its content will be appended to the basetype image
task, like do_image_ostree, do_image_ota, while, the other tar commond is:
IMAGE_CMD_tar, it will be handled by bitbake by adding a do_image_tar task
before do_image_complete after do_image, so there is no conflicting between
them.

I have verified the following build is OK, IMAGE_FSTYPES=" tar.bz2 ext4
ostreepush garagesign garagecheck ota-ext4 wic ostree.tar.bz2", in which:

tar.bz2: will create a tarball against IMAGE_ROOTFS.
ostree.tar.bz2: will create a tarball against OSTREE_ROOTFS or OTA_SYSROOT
depending on the basetype task is do_image_ostree or do_image_ota.


> E.g. we have now: ota.tar.xz. Is the first part always the base type, so
IMAGE_CMD_ota must exist, and then every other extension is a conversion?

Yes, exactly, now ota is a basetype, so it must exist, that's why I changed
the task do_image_otasetup -> do_image_ota.

the best,
thank you

Stefan Agner <notifications@github.com> 於 2018年11月20日 週二 上午10:07寫道：

agners:
Ok, thanks for clarification and verification!

Some small things I'd prefer you would address:

> "Move the dependency do_image_wic -> do_image_ota_ext4 to image_types_ota.bbclass, since it's a image specific dependency."

This change is completely independent from what I understand. Can you make it a separate commit?


> "Remove ota-tar from IMAGE_FSTYPES, no need to set it explicitly, it is redundant code."

This part of the commit message seems not entirely accurate anymore.

liuming50:
Hi, Stefan:

I have sent a V5 with the modification as you suggested, please kindly help me review it.

//Ming Liu

agners:
Perfect thanks!

Just tested here, also with `BUILD_OTA_TARBALL` set. This looks good from my side!

liuming50:
Please hold on reviewing this patch series, it needs to be reworked since
Stefan had found a CONVERSION_CMD_tar and IMAGE_CMD_tar conflict, it
happened when both tar.xz and ota.tar.xz was set in IMAGE_FSTYPES. I am
working on it, will send a V5.

the best,
thank you

Stefan Agner <notifications@github.com> 於 2018年11月20日 週二 下午1:39寫道：

liuming50:
Changes in V5:

1 Add a sanity check to ensure the OVERRIDES do not include "ota ostree", which is leading naming pollution with meta-updater layer.

Now it's ready to be reviewed, the error Stefan reported, is not actually caused by this patch series (my original analysis is wrong), it's actually caused by some other layers are setting 'ota' as a OVERRIDES.

agners:
Oh I see, the conversion of the `do_otasetup` to a proper `IMAGE_TYPE` named `ota` together with the use of `ota` in `DISTROOVERRIDES` killed it...

Yeah I guess especially `ota` could be used by other layers too, so its a nice touch to actually check OVERRIDES!

The changes looks good to me, thanks!

classes/image_types_ota.bbclass
```diff
@@ -43,18 +42,14 @@ calculate_size () {
...
-export GARAGE_TARGET_NAME
-
-export OTA_SYSROOT="${WORKDIR}/ota-sysroot"
-
-## Common OTA image setup
-fakeroot do_otasetup () {
-	
+OTA_SYSROOT = "${WORKDIR}/ota-sysroot"
+OTA_IMAGE_ROOTFS_task-image-ota = "${OTA_SYSROOT}"
+IMAGE_TYPEDEP_ota = "ostree"
```
agners: Still found another one :-)

I think this should depend on `ostreecommit`.

For two reasons: First, we need the ostree native binary (which is a depends of do_image_ostreecommit), and second because we actually pull from the local repository to create the initial OTA OSTree.

liuming50:
> Still found another one :-)

> I think this should depend on ostreecommit.

> For two reasons: First, we need the ostree native binary (which is a
depends of do_image_ostreecommit), and second because we actually pull from
the local repository to create the initial OTA OSTree.

Arha, you are absolutely right! Thanks for finding out this error, will
send a V6 soon.

//Ming Liu




Stefan Agner <notifications@github.com> 於 2018年11月21日 週三 下午5:20寫道：

liuming50:
Changes in V6:
1 do_image_ota should depend on ostreecommit instead of ostree.

liuming50:
Changes in V7:

1 Stefan found another issue, that is: when ota.tar.xz or ostree.tar.bz2 is being depended by other image types by setting IMAGE_TYPEDEP_, then ota or ostree would be treated as a intermittent type and there should be a .ota or .ostree generated beforehand and deleted later on, so we hack this by touching a .ota or .ostree file in CONVERSION_CMD_tar command.

agners:
Thanks Ming, tested v7 and now works fine also with our layers too!

patrickvacek:
Passes oe-selftest. I still need to do some more manual checks but it's looking good.

liuming50:
Thanks for the testing, Patrick.

//Ming Liu

Patrick Vacek <notifications@github.com> 於 2018年11月22日 週四 下午1:59寫道：

patrickvacek:
I've reviewed this more thoroughly and I think it looks quite good. There is only one minor issue I still found: the various fatal variable checks can be cleaned up a bit. The only actually bad thing is that `OSTREE_BRANCHNAME` is used by `IMAGE_CMD_ostree` but itsn't checked for non-emptiness until `IMAGE_CMD_ostreecommit`. The other part is that `IMAGE_CMD_ota` checks `OSTREE_REPO` and `OSTREE_BRANCHNAME`, but since it depends on `IMAGE_CMD_ostreecommit`, those two checks are actually redundant.

And unfortunately you will need to rebase again because we've added another variable in the meantime. Should be a simple fix, though.

I've tested a full round-trip with qemu and everything looked fine, so we are very close to getting this done.

liuming50:
HI, Patrick:

OK, will send a V8 with the fixes for these minor issues, and rebase it.

//Ming Liu

Patrick Vacek <notifications@github.com> 於 2018年11月23日 週五 上午10:12寫道：

agners:
Looks good to me, and seems to build our use case fine.

liuming50:
Hi, Patrick/Stefan:

Thanks for the review.

//Ming Liu

Patrick Vacek <notifications@github.com> 於 2018年11月26日 週一 下午4:49寫道：

agners:
@patrickvacek I see that the thud branch has been created. We actually target thud currently, is there a chance to get this change backported?

patrickvacek:
@agners Yes, sure! I was actually planning on doing some backporting later this week. I've been debating how much to port from master to thud, but since they are still fairly similar, it should be easy enough to take almost everything since we made the split.

agners:
@patrickvacek cool, thanks! Just let me or @liuming50 know if we can do anything which helps the processes.

patrickvacek:
Cool, thanks. I think it should be easy. Of course, what I think and what the reality is are often different, but it'll probably be fine. :)

patrickvacek:
@liuming50 It appears that this change has made the ota-ext4 output balloon from about 400 MB to somewhere around 50-100 GB. Can you think of a reason why that would happen?

liuming50:
Hi, Patrick:

Will investigate it.

the best,
thank you

Patrick Vacek <notifications@github.com> 於 2018年11月27日 週二 下午1:41寫道：

liuming50:
HI, Patrick:

Unfortunately, I could not reproduce this issue on my machine, I got the
ota-ext4 image around 400M, but I guess the root cause is somehow the
OTA_SYSROOT's size is being calculated wrongly, as follows:

```
IMAGE_CMD_ota-ext4 () {
        # Calculate image type
        OTA_ROOTFS_SIZE=$(calculate_size `du -ks $OTA_SYSROOT | cut -f 1`
"${IMAGE_OVERHEAD_FACTOR}" "${IMAGE_ROOTFS_SIZE}" "${IMAGE_ROOTFS_MAXSIZE}"
`expr ${IMAGE_ROOTFS_EXTRA_SPACE}` "${IMAGE_ROOTFS_ALIGNMENT}")
```

could you please try "bitbake -v -v -v core-image-minimal", then you can
get the output of all shell tasks, then you can check why the size is so
huge from the output.

the best,
thank you


Ming Liu <liu.ming50@gmail.com> 於 2018年11月27日 週二 下午2:52寫道：

lbonn:
Hello @liuming50, it seems that @OYTIS found the root problem which indeed lays in this part of the code: #438. The wrong bitbake variable usage has been here forever but recent changes seem to have made it worse.

Thanks for your help! We'll test and confirm that it completely solves the issue.

liuming50:
Arha, that's good catching! Sorry, my fault, I did not test the ota-ext4
size.

//Ming Liu

lbonn <notifications@github.com> 於 2018年11月28日 週三 上午10:09寫道：

patrickvacek:
> We actually target thud currently, is there a chance to get this change backported?

@agners FYI I've created a PR for a thud backport (actually a merge) including these changes (and the subsequent fixes): https://github.com/advancedtelematic/meta-updater/pull/443

PR #434, opened 2018-11-19
selftest: introduce QEMU_USE_KVM

agners:
The qemucommand.py script uses kvm-ok to determine whether KVM
is available. However, kvm-ok is very Ubuntu specific and not
readily available on Fedora.

Use QEMU_USE_KVM variable which is also used in OpenEmbedded
selftests. The variable must contain True to enable KVM for
machines containing x86 in its name, or contain a list of
machines. For meta-updater this makes sure KVM is used for all
tests:
QEMU_USE_KVM = "intel-corei7-64 qemux86-64"

### DIFF
--- a/lib/oeqa/selftest/cases/updater.py
+++ b/lib/oeqa/selftest/cases/updater.py
@@ -655,7 +655,13 @@ def qemu_launch(efi=False, machine=None, imagename=None):
     args.dir = 'tmp/deploy/images'
     args.efi = efi
     args.machine = machine
-    args.kvm = None  # Autodetect
+    qemu_use_kvm = get_bb_var("QEMU_USE_KVM")
+    if qemu_use_kvm and \
+       (qemu_use_kvm == 'True' and 'x86' in machine or \
+        get_bb_var('MACHINE') in qemu_use_kvm.split()):
+        args.kvm = True
+    else:
+        args.kvm = None  # Autodetect
     args.no_gui = True
     args.gdb = False
     args.pcap = None


agners:
We use a Fedora server, hence I had to debug some meta-updater selftest issues.

Better than the change above would probably be to use OEs Qemu automation, e.g. `meta/lib/oeqa/targetcontrol.py`... But this would be a rather involved change.

Besides this change I also had to use "/usr/share/edk2/ovmf/OVMF_CODE.fd" as location for OVMF.fd `scripts/qemucommand.py`.

From what I can tell OpenEmbedded uses OVMF built by OE (see `meta/lib/oeqa/selftest/cases/efibootpartition.py`). So this would be probably also easier to solve when moving to OE Qemu automation code...

patrickvacek:
Thanks! The KVM fix seems fine. I'm testing it now to make sure it works okay for us.

About OE qemu automation, I believe that once upon a time there was a good reason why we didn't use it, but I'm not sure I ever knew the reason, and it is increasingly less likely to still be valid today. It probably would be ideal if we could make use of what is already there and well supported. I'll try to keep that in mind and if I get a chance I'll try to figure out why we took a different path back when we started this.

cajun-rat:
> About OE qemu automation, I believe that once upon a time there was a good reason why we didn't use it, but I'm not sure I ever knew the reason, and it is increasingly less likely to still be valid today. It probably would be ideal if we could make use of what is already there and well supported. I'll try to keep that in mind and if I get a chance I'll try to figure out why we took a different path back when we started this.

I can explain this one :). The OE qemu automation didn't have an easy way to start from the U-Boot or UEFI bootloader. Their design (at least at the time) used qemu's built-in bootloader (`-kernel` and friends), and so there isn't (wasn't?) a way to get OE to start the boot process at U-Boot.

Now that we are tracking the latest Yocto master, it might be possible to work with OE upstream to add the suitable hooks to OE to make this possible.

agners:
At least UEFI/OVMF seems to be there nowadays, introduced with this commit:
https://patchwork.openembedded.org/patch/135757/

U-Boot support seems not to be readily available.

cajun-rat:
> At least UEFI/OVMF seems to be there nowadays, introduced with this commit:
> https://patchwork.openembedded.org/patch/135757/

Interesting--thank for for the pointer to that!

lib/oeqa/selftest/cases/updater.py
```diff
@@ -655,7 +655,14 @@ def qemu_launch(efi=False, machine=None, imagename=None):
     args.dir = 'tmp/deploy/images'
     args.efi = efi
     args.machine = machine
-    args.kvm = None  # Autodetect
+    get_bb_var('DEPLOY_DIR_IMAGE')
```
patrickvacek: What is the benefit of getting `DEPLOY_DIR_IMAGE` but not using it?

patrickvacek:
@cajun-rat thanks for the historical context and @agners thanks for finding that patch! It would be worth moving in that direction if all of our boot needs can be met.

lib/oeqa/selftest/cases/updater.py
agners: Oh I think that is a copy paste issue. Will update the merge request.

PR #435, opened 2018-11-21
Allow setup custom hardware_id

patriotyk:
I was able to build image with custom hardware id.
Image has been successfully provisioned and appeared in the web admin panel.
Also Image successfully has been updated to a new version of image.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -87,6 +87,7 @@ Although we have used U-Boot so far, other boot loaders can be configured work w
 * `SOTA_CLIENT_PROV` - which provisioning method to use. Valid options are https://github.com/advancedtelematic/aktualizr/blob/master/docs/automatic-provisioning.adoc[`aktualizr-auto-prov`], https://github.com/advancedtelematic/aktualizr/blob/master/docs/implicit-provisioning.adoc[`aktualizr-ca-implicit-prov`], and https://github.com/advancedtelematic/aktualizr/blob/master/docs/hsm-provisioning.adoc[`aktualizr-hsm-prov`]. The default is `aktualizr-auto-prov`. This can also be set to an empty string to avoid using a provisioning recipe.
 * `SOTA_CLIENT_FEATURES` - extensions to aktualizr. The only valid options are `hsm` (to build with HSM support) and `secondary-network` (to set up a simulated 'in-vehicle' network with support for a primary node with a DHCP server and a secondary node with a DHCP client).
 * `SOTA_SECONDARY_CONFIG_DIR` - a directory containing JSON configuration files for virtual secondaries on the host. These will be installed into `/etc/sota/ecus` on the device and automatically provided to aktualizr.
+* `SOTA_HARDWARE_ID` - a custom hardware id variable. If not set, - MACHINE variable is used.
 
 == Usage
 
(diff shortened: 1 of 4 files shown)


README.adoc
```diff
@@ -87,6 +87,8 @@ Although we have used U-Boot so far, other boot loaders can be configured work w
...
 * `SOTA_CLIENT_FEATURES` - extensions to aktualizr. The only valid options are `hsm` (to build with HSM support) and `secondary-network` (to set up a simulated 'in-vehicle' network with support for a primary node with a DHCP server and a secondary node with a DHCP client).
 * `SOTA_SECONDARY_CONFIG_DIR` - a directory containing JSON configuration files for virtual secondaries on the host. These will be installed into `/etc/sota/ecus` on the device and automatically provided to aktualizr.
+* `SOTA_SECONDARY_CONFIG_DIR` - a directory containing JSON configuration files for virtual secondaries on the host. These will be installed into `/etc/sota/ecus` on the device and automatically provided to aktualizr.
+* `HARDWARE_ID` - a custom hardware id variable. If not set, - MACHINE varable is used.
```
patrickvacek: Looks like `SOTA_SECONDARY_CONFIG_DIR` got duplicated somehow.

classes/sota.bbclass
```diff
@@ -29,7 +30,7 @@ INITRAMFS_FSTYPES ??= "${@oe.utils.ifelse(d.getVar('OSTREE_BOOTLOADER', True) ==
 
 # Please redefine OSTREE_REPO in order to have a persistent OSTree repo
 OSTREE_REPO ?= "${DEPLOY_DIR_IMAGE}/ostree_repo"
-OSTREE_BRANCHNAME ?= "${MACHINE}"
+OSTREE_BRANCHNAME ?= "${HARDWARE_ID}"
```
patrickvacek: I think `SOTA_HARDWARE_ID` might be slightly better. I'm debating whether `PRIMARY` is a good word to include as well (or instead of `SOTA`), but that seems like overkill and might not even be accurate; we can also bitbake secondaries.

recipes-sota/aktualizr/aktualizr_git.bb
```diff
@@ -70,6 +70,8 @@ do_install_append () {
     install -m 0700 -d ${D}${libdir}/sota/conf.d
     install -m 0700 -d ${D}${sysconfdir}/sota/conf.d
 
+    echo "[uptane]\nprimary_ecu_hardware_id = ${HARDWARE_ID}\n" > ${D}${libdir}/sota/conf.d/40-hardware-id.toml
+
```
patrickvacek: Originally I was thinking it would be best if we did this echo in a separate recipe, like one of the ones in the `recipes-sota/config` directory, but actually, we could just keep it here. But if we keep it here, it should be guarded to check that `HARDWARE_ID` is set and nonempty.

README.adoc
```diff
@@ -87,6 +87,7 @@ Although we have used U-Boot so far, other boot loaders can be configured work w
 * `SOTA_CLIENT_PROV` - which provisioning method to use. Valid options are https://github.com/advancedtelematic/aktualizr/blob/master/docs/automatic-provisioning.adoc[`aktualizr-auto-prov`], https://github.com/advancedtelematic/aktualizr/blob/master/docs/implicit-provisioning.adoc[`aktualizr-ca-implicit-prov`], and https://github.com/advancedtelematic/aktualizr/blob/master/docs/hsm-provisioning.adoc[`aktualizr-hsm-prov`]. The default is `aktualizr-auto-prov`. This can also be set to an empty string to avoid using a provisioning recipe.
 * `SOTA_CLIENT_FEATURES` - extensions to aktualizr. The only valid options are `hsm` (to build with HSM support) and `secondary-network` (to set up a simulated 'in-vehicle' network with support for a primary node with a DHCP server and a secondary node with a DHCP client).
 * `SOTA_SECONDARY_CONFIG_DIR` - a directory containing JSON configuration files for virtual secondaries on the host. These will be installed into `/etc/sota/ecus` on the device and automatically provided to aktualizr.
+* `SOTA_HARDWARE_ID` - a custom hardware id variable. If not set, - MACHINE varable is used.
```
xaviergully: Typo - MACHINE variable

PR #436, opened 2018-11-26
oe-selftest: rpi-basic-image is deprecated. Use core-image-basic.

patrickvacek:


lib/oeqa/selftest/cases/updater.py
```diff
@@ -285,17 +285,17 @@ def tearDownLocal(self):
 
     def test_rpi(self):
         logger = logging.getLogger("selftest")
-        logger.info('Running bitbake to build rpi-basic-image')
+        logger.info('Running bitbake to build core-image-base')
         self.append_config('SOTA_CLIENT_PROV = "aktualizr-auto-prov"')
-        bitbake('rpi-basic-image')
+        bitbake('core-image-base')
```
lbonn: `core-image-minimal` is sufficient in our case (tried it) and that's already what we use for qemux86-64.

patrickvacek: Not only sufficient but apparently 1/20th the size!

patrickvacek:
Tested relevant oe-selftest case.

PR #437, opened 2018-11-27
classes/image_types_ostree.bbclass: add device tree to boot/devicetree

LukaPivk:
Include device trees into boot/devicetree-*-checksum. Device trees with
this name format get picked up by ostree and copied into boot/.
Ostree also generates entry inside uEnv.txt called fdt_file pointing to
this device tree.

Normaly one would include only one device tree. If KERNEL_DEVICETREE
includes more then one device tree this will add all the device trees
but ostree will still pick only one device tree and include it into
boot.

We are working on modifications to ostree so it will pickup all
the device trees.

Signed-off-by: Luka Pivk <luka.pivk@toradex.com>


README.adoc
```diff
@@ -88,7 +88,7 @@ Although we have used U-Boot so far, other boot loaders can be configured work w
 * `SOTA_CLIENT_FEATURES` - extensions to aktualizr. The only valid options are `hsm` (to build with HSM support) and `secondary-network` (to set up a simulated 'in-vehicle' network with support for a primary node with a DHCP server and a secondary node with a DHCP client).
 * `SOTA_SECONDARY_CONFIG_DIR` - a directory containing JSON configuration files for virtual secondaries on the host. These will be installed into `/etc/sota/ecus` on the device and automatically provided to aktualizr.
 * `SOTA_HARDWARE_ID` - a custom hardware ID that will be written to the aktualizr config. Defaults to MACHINE if not set.
-
+* `OSTREE_DEPLOY_DEVICETREE` - Set this to '1' to include devicetree(s) to boot
```
patrickvacek: I know this is trivial, but we actually need the empty line before the `Usage` section header, otherwise it gets parsed incorrectly.

LukaPivk:
> Looks great, thank you. We were using FIT images to solve this problem, but it is good to give the user a choice.
> It will also require support in bootloader scripts, that we haven't yet standardized unfortunately. Did you already test it under some environment?

I tested on our setup with u-boot. With single device tree included in this format -> devicetree-{filename}-checksum ostree will be aware that this is device tree ( same as its aware of kernel ). When ostree will pull from repository and this file is present it will add fdt_file entry into uEnv.txt in case of u-boot. U-Boot can then use this variable to load correct device tree. 

Now for multiple device trees its a bit trickier. Part of solution is to add all device trees in correct format. Here comes the issue, currently ostree only pics one device tree even if there are multiple ones present.  I am in contact with Javier Martinez who wrote that part of ostree to find out best solution. Solution I currently have works but I am not yet happy with it.

This merge request solved using 1 device tree that is already supported and also is future proof for when I finish ostree implementation.

patrickvacek:
oe-selftest passes.

LukaPivk:
Any update on this?

patrickvacek:
@LukaPivk The problem is that we don't have a good way to test this. @OYTIS has some ideas but they would take some work to set up, so it may take some time for us to make it happen.

OYTIS:
@LukaPivk I'm happy to confirm that this PR doesn't break anything in the FIT case and boots on Raspberry Pi (overlays don't work yet naturally). I don't think we'll want to make Raspberry Pi support separate initramfs and dtbs though, since device tree overlays are pretty critical there. Is there any progress with supporting overlays in OSTree?
Do you have any plans to make your work on meta-updater on Toradex boards public by the way? No pressure, just curious =)

LukaPivk:
> @LukaPivk I'm happy to confirm that this PR doesn't break anything in the FIT case and boots on Raspberry Pi with separate kernel, ramdisk and dtbs (overlays don't work yet naturally). I don't think we'll want to make Raspberry Pi support separate initramfs and dtbs though, since device tree overlays are pretty critical there. Is there any progress with supporting overlays in OSTree?
> Do you have any plans to make your work on meta-updater on Toradex boards public by the way? No pressure, just curious =)

Yes we will do everything opensource.

in fact you can already find some info here -> https://labs.toradex.com/projects/torizon

and even some docs here https://github.com/toradex/toradex-torizon-manifest
There is always lots of things in internal testing or research phase so not everything we currently work on is published already. We only try to push stuff we know we are going to use sooner or later.

OYTIS:
Looks great so far, thank you!

PR #438, opened 2018-11-27
Fix bloated OTA image

OYTIS:
Curly braces are necessary around bitbake variables


lbonn:
Wow! What a catch...

agners:
That makes sense, since we removed the export in "meta: drop redundant export syntax", the variable only exists as a bitbake variable. It is the only location where `OTA_SYSROOT` got used without curly braces.

Looks good to me!

PR #439, opened 2018-11-28
image_types_ostree: use OSTree repo location for lockfiles

agners:
OSTree does not allow concurrent access to the repository. If
two bitbake instances use the same OSTree repository, then they
might access it simultanously. The lockfile to prevent such
simultaneous access should really be in the same location where
the OSTree repository is stored.

Revert back to the location inside the OSTree repository.

Fixes: e97975d6113c ("image_types_ostree/ota.bbclass: refactor ostree task")

### DIFF
--- a/classes/image_types_ostree.bbclass
+++ b/classes/image_types_ostree.bbclass
@@ -129,7 +129,7 @@ IMAGE_CMD_ostree () {
 
 IMAGE_TYPEDEP_ostreecommit = "ostree"
 do_image_ostreecommit[depends] += "ostree-native:do_populate_sysroot"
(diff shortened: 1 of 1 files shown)


agners:
@liuming50 can you have a quick look at that? I am not sure what your rational was when you changed the location...

patrickvacek:
This did catch my eye before but I didn't think much of it. @agners I think you are probably correct. We've got a bit of a backlog to test now but we will get around to it soon.

liuming50:
Sorry, my fault. I did not think about the 2 bitbake instance could commit to a repo, I just thought it's would be better to put a lockfile into ${WORKDIR}

I agree with this fix.

patrickvacek:
@agners Would you mind rebasing?

agners:
@patrickvacek no problem, done!

PR #440, opened 2018-11-28
Use usrmerge

agners:
This makes use of the OpenEmbedded feature usrmerge, which implements usrmove at recipe build/package time. We build with usrmerge DISTRO_FEATURE since quite some time, and have not seen a downside here.

### DIFF
--- a/classes/image_types_ostree.bbclass
+++ b/classes/image_types_ostree.bbclass
@@ -1,4 +1,5 @@
 # OSTree deployment
+inherit distro_features_check
 
 OSTREE_KERNEL ??= "${KERNEL_IMAGETYPE}"
 OSTREE_ROOTFS ??= "${WORKDIR}/ostree-rootfs"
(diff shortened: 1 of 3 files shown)


agners:
As far as I understand, it really make sure that nothing is installed in `/bin` etc by defining the paths correctly on a very low level: https://patchwork.openembedded.org/patch/140830/

With `usrmerge` OE also checks that no package is doing anything wrong using QA tests.

Some packages also behave special, e.g. the systemd recipe passes the information to the systemd build system (uses `-Dsplit-usr=true`). This has quite some consequences throughout systemd: (see https://github.com/systemd/systemd/search?utf8=%E2%9C%93&q=HAVE_SPLIT_USR&type=).

Since we used usrmerge since start, I cannot really tell how it exactly differ in practise...

But it seems to me to be really the right thing to do (tm) :-)

agners:
See also: http://lists.openembedded.org/pipermail/openembedded-devel/2017-January/110758.html

ricardosalveti:
> Since we used usrmerge since start, I cannot really tell how it exactly differ in practise...
> 
> But it seems to me to be really the right thing to do (tm) :-)

One other thing I like about this is that the original rootfs gets closer to the ostree one, as the build system will then be responsible for making sure the paths are all correct.

OYTIS:
@agners Thank you, testing.
I'm a bit concerned about potential problems integrating with external layers that might not respect usrmerge discipline (e.g. hard-coding installation paths instead of using bitbake variables) though.

ricardosalveti:
> I'm a bit concerned about potential problems integrating with external layers that might not respect usrmerge discipline (e.g. hard-coding installation paths instead of using bitbake variables) though.

The OE QA checks will warn the user, but manual changes might indeed be required.

agners:
@OYTIS Since the base directories are redefined and usually passed to the configure step, it should just work for most build systems., but as @ricardosalveti mentions, QA will report it:
https://github.com/openembedded/openembedded-core/blob/master/meta/classes/insane.bbclass#L815

So far I haven't encountered a package, but we also did not add a lot. I guess we just have to fix wherever issues crops up, but i expect that at least core layers are already well tested. After all, this is proposed for `master` branch...

patrickvacek:
Passes oe-selftest.

PR #443, opened 2018-11-29
Feat/thud/hardware id etc

patrickvacek:
Instead of cherry-picking a few dozen commits, I just pulled master into thud. Since thud hasn't diverged far from master yet, this is by far easier.

Passes oe-selftest.


agners:
Thanks for the backport! This builds fine for me in our thud setup.

PR #446, opened 2018-12-12
Make INITRAMFS_FSTYPES an override

OYTIS:
It is how the other SOTA-specific variables are treated.

### DIFF
--- a/classes/sota.bbclass
+++ b/classes/sota.bbclass
@@ -21,7 +21,7 @@ WKS_FILE_sota ?= "sdimage-sota.wks"
 
 EXTRA_IMAGEDEPENDS_append_sota = " parted-native mtools-native dosfstools-native"
 
-INITRAMFS_FSTYPES ??= "${@oe.utils.ifelse(d.getVar('OSTREE_BOOTLOADER', True) == 'u-boot', 'cpio.gz.u-boot', 'cpio.gz')}"
+INITRAMFS_FSTYPES_sota ??= "${@oe.utils.ifelse(d.getVar('OSTREE_BOOTLOADER', True) == 'u-boot', 'cpio.gz.u-boot', 'cpio.gz')}"
 
 # Please redefine OSTREE_REPO in order to have a persistent OSTree repo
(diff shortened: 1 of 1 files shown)


patrickvacek:
What's the advantage? Less risk for messing up other unrelated builds?

Jmennius:
Hi @OYTIS,

Are you sure this works as expected (did you test w/o sota override)?
In my experience overrides don't work with weak assignments correctly (and it's actually broken everywhere..).

OYTIS:
@Jmennius 
Yes, overrides have higher priority than normal assignment. So in this case `_sota =` > `_sota ?=` > `_sota ??=` > `=` > `?=` > `??=`. So we can override this variable with another a stronger assignment with `_sota` override in, for instance, a board-related class.

Jmennius:
@OYTIS, this is exactly what I am concerned about.
The end result of this is that you can only override this kind of 'weak conditional assignment' with an additional override. This is extremely inconvenient.

For example, let's take a `SOTA_MACHINE` variable.
When you implement your machine (in Yocto/OE terms) based on RPi, you create a machine config file `mymachine.conf`, in which:
- you usually include a base board config (e.g. RPi) to inherit its configuration
- extend `MACHINEOVERRIDES` to include `rpi`
- and in the case of meta-updater integration, you set `SOTA_MACHINE` to `MACHINE` in your distro configuration (if you are not doing any tricky name mappings)
- alternatively, you can set `SOTA_MACHINE` to a value of `MACHINE` in your machine configuration, but this is no good

Having this style of override makes you do `SOTA_MACHINE_mymachine = "${MACHINE}"` in your distro config, which is quite bad, don't you agree?

Regarding the PR itself,
I think this should be fixed in OE to have weak default value assignment, and the meta-updater to use ordinary default assignment.
Alternatively, we can still use plain assignment in meta-updater and other layers should be able to override this assignment by ordering their assignment after meta-updater one.
IMO, using overrides that way is, to say the least, very inconvenient.

P.S. I also don't think this is properly documented (please point me if you know).

OYTIS:
@Jmennius @patrickvacek 
OK, I believe this can be closed, probably consistency is not a good enough reason to break things.

PR #447, opened 2018-12-14
Configure oe-selftests to run on specific node label

kmanikandan-kmk:


patrickvacek:
So which machines have the bitbake label? Can I take it this just changes the default which machines this runs on by default in Jenkins?

kmanikandan-kmk:
> So which machines have the bitbake label? Can I take it this just changes the default which machines this runs on by default in Jenkins?

bagger node has label bitbake. So, these tests will run only in bagger now.

Issue #449, opened 2019-01-03
Checksum not calculated for initramfs and devicetree


LukaPivk:
Hi,

We ware reading ostree documentation and came across this statement 

>  If an initramfs or devicetree exist, the checksum must include all of the kernel, initramfs and devicetree contents
On this website https://ostree.readthedocs.io/en/latest/manual/deployment/

I see that image ostree only calculates kernel checksum and not device tree and initramfs. 
Is there a reason for that? This ostree change and meta-updater was not updated or is it simply a bug?
If its simply a bug I will create a pull request but I wanted to know if there is some reason behind it.

Thanks

patrickvacek:
@LukaPivk no reason that I am aware of. It looks like a bug to me. I think we either overlooked that line or it was added after we implemented the checksum and we never went back and updated it.

Good catch!

LukaPivk:
ok good. I will fix it test it and create pull request.
Thanks

PR #454, opened 2019-01-10
classes/image_types_ostree.bbclass: fix checksum calculation

LukaPivk:
Adds initramfs and/or devicetree into checksum calculation.

According to ostree documentation if devicetree and/or initramfs are
included into ostree, checksum should also include them.

This would fix issue #449

### DIFF
--- a/classes/image_types_ostree.bbclass
+++ b/classes/image_types_ostree.bbclass
@@ -104,25 +104,26 @@ IMAGE_CMD_ostree () {
         ln -sf var/roothome root
     fi
 
-    checksum=`sha256sum ${DEPLOY_DIR_IMAGE}/${OSTREE_KERNEL} | cut -f 1 -d " "`
-
-    cp ${DEPLOY_DIR_IMAGE}/${OSTREE_KERNEL} boot/vmlinuz-${checksum}
-
     if [ "${KERNEL_IMAGETYPE}" = "fitImage" ]; then
         # this is a hack for ostree not to override init= in kernel cmdline -
         # make it think that the initramfs is present (while it is in FIT image)
+        # since initramfs is fake file, it does not need to be included in checksum
+        checksum=$(sha256sum ${DEPLOY_DIR_IMAGE}/${OSTREE_KERNEL} | cut -f 1 -d " ")
         touch boot/initramfs-${checksum}
     else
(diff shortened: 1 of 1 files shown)


classes/image_types_ostree.bbclass
```diff
@@ -104,26 +104,30 @@ IMAGE_CMD_ostree () {
         ln -sf var/roothome root
     fi
 
-    checksum=`sha256sum ${DEPLOY_DIR_IMAGE}/${OSTREE_KERNEL} | cut -f 1 -d " "`
-
-    cp ${DEPLOY_DIR_IMAGE}/${OSTREE_KERNEL} boot/vmlinuz-${checksum}
+    FILES_TO_CHECKSUM="${DEPLOY_DIR_IMAGE}/${OSTREE_KERNEL} ${INITRAMFS_IMAGE}-${MACHINE}.${INITRAMFS_FSTYPES} ${KAKUS}"
```
patrickvacek: Sorry, what is `KAKUS`? I can't find it defined anywhere.

classes/image_types_ostree.bbclass
```diff
@@ -104,26 +104,30 @@ IMAGE_CMD_ostree () {
...
+    # calculate checksum of kernel and initramfs. In case of fitimage we ignore error becuase initramfs is not there
+    checksum=$(cat ${FILES_TO_CHECKSUM} 2>/dev/null  | sha256sum | cut -f 1 -d " ")
```
patrickvacek: Would it make sense to separate the checksum calculation out into three separate cases? If we already have two cases, and one of them is actually two cases in which an error is expected in one of them, it seems safer to just calculate it separately for all three:
1. fitImage case where there is just a kernel
2. "else" case where there is kernel and initramfs (no error ignoring necessary)
3. "else" case with devicetree

classes/image_types_ostree.bbclass
LukaPivk: sorry about this one. Its leftover from testing.

classes/image_types_ostree.bbclass
patrickvacek: I would agree that fewer variations are better, but since all three cases are different, I'd rather have three separate cases. In particular, I don't like that we ignore potential errors in the non-FIT case.

LukaPivk: ok I will fix it as soon as I am back to pc

PR #462, opened 2019-01-28
aktualizr: bump to latest 1cad6d10286ade64b24021ca0e23de0d3b64f520

lbonn:
- OStree installations are considered complete after a reboot under the new version
- image downloads are not parallelized anymore
- various enhancements and bug fixes

oe-selftests are currently running


patrickvacek:
Should be fine if oe-selftest is happy, but I swear that I broke something that I was expecting to have to fix in meta-updater. Unfortunately that was before the new year and I've forgotten what that was.

lbonn:
Ok got an anonymous tip: the broken bit is oe-selftests still using the `--running-mode=` syntax.

PR #463, opened 2019-01-29
Test/morty oe selftest

patrickvacek:
A few tests are still failing:
* updater.SecondaryTests
* updater.HsmTests
* updater.ImplProvTests

I don't know why the secondary tests are failing, but I'm not too concerned; that code is old and possibly obsolete. The HSM and Implicit tests are failing because of sysroot/path problems with ssh. It works when manually testing.

### DIFF
--- a/lib/oeqa/selftest/qemucommand.py
+++ b/lib/oeqa/selftest/qemucommand.py
@@ -1,0 +1,1 @@
+../../../scripts/qemucommand.py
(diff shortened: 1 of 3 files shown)


lbonn:
Maybe it's time to merge that and rework on top of it later if needed.

patrickvacek:
> Maybe it's time to merge that and rework on top of it later if needed.

Fine with me! This doesn't affect anything but the tests (except for one minor fix), so I think we can safely merge it without further testing if you want to approve it.

PR #464, opened 2019-01-29
Feat/rocko latest aktualizr

lbonn:
Backport from https://github.com/advancedtelematic/meta-updater/pull/462


lbonn:
Note for testing: a full cycle update will now need a reboot of the device to show as completed on the UI

PR #468, opened 2019-01-29
sota.bbclass: fix a variable assignment regression

liuming50:
A regression was introduced by commit 9dcfcdb9:
[ classes, images: Use standard variables for initramfs ]

it replaced OSTREE_INITRAMFS_IMAGE with INITRAMFS_FSTYPES which is
fine, but the '??=' should be changed to "?=" as well, or else it will
not take effect since INITRAMFS_FSTYPES's already set in bitbake.conf
with '?='.

### DIFF
--- a/recipes-sota/aktualizr/aktualizr_git.bb
+++ b/recipes-sota/aktualizr/aktualizr_git.bb
@@ -5,16 +5,11 @@ SECTION = "base"
 LICENSE = "MPL-2.0"
 LIC_FILES_CHKSUM = "file://${S}/LICENSE;md5=9741c346eef56131163e13b9db1241b3"
 
-DEPENDS = "boost curl openssl libarchive libsodium asn1c-native sqlite3 "
-DEPENDS_append_class-target = "ostree ${@bb.utils.contains('SOTA_CLIENT_FEATURES', 'hsm', ' libp11', '', d)} "
-DEPENDS_append_class-native = "glib-2.0-native "
+require garage-sign-version.inc
 
-RDEPENDS_${PN}_class-target = "lshw "
-RDEPENDS_${PN}_append_class-target = "${@bb.utils.contains('SOTA_CLIENT_FEATURES', 'serialcan', ' slcand-start', '', d)} "
-RDEPENDS_${PN}_append_class-target = " ${@bb.utils.contains('SOTA_CLIENT_FEATURES', 'ubootenv', ' u-boot-fw-utils aktualizr-uboot-env-rollback', '', d)} "
-
-RDEPENDS_${PN}_append_class-target = " ${PN}-tools "
-RDEPENDS_${PN}-secondary_append_class-target = " ${PN}-tools "
+DEPENDS = "boost curl openssl libarchive libsodium sqlite3 asn1c asn1c-native"
+RDEPENDS_${PN}_class-target = "${PN}-tools lshw"
+RDEPENDS_${PN}-secondary_class-target = "${PN}-tools"
 
 PV = "1.0+git${SRCPV}"
 PR = "7"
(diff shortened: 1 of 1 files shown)


agners:
Agreed, fixes what has been broken back then.

U-Boot actually can handle non-u-bootified initramfs, but that depends on the exact U-Boot configuration (if `CONFIG_SUPPORT_RAW_INITRD` is used). But thanks to `?=` a distribution can still opt to use a different like this:

```
INITRAMFS_FSTYPES = "cpio.gz"
```

patrickvacek:
This looks fine and oe-selftest passed except for the grub test. I'm trying to determine if that is related or not.

liuming50:
I have pushed another patch, to introduce PACKAGECONFIG to aktualizr, please kindly help me review.

agners:
@liuming50 you really should not push other features on the same branch, it gets messy here on this Github pull request.

Create feature branches (e.g. `git checkout -b introduce-packageconfig`) and create separate pull requests.

liuming50:
Arha, got this, thanks for the tips, will do that.

the best,
thank you

Stefan Agner <notifications@github.com> 於 2019年1月30日 週三 下午11:14寫道：

recipes-sota/aktualizr/aktualizr_git.bb
```diff
@@ -32,29 +27,34 @@ BRANCH ?= "master"
...
-                                      -DBUILD_SYSTEMD=OFF \
-                                      -DGARAGE_SIGN_VERSION=${GARAGE_SIGN_VERSION} \
-                                      -DGARAGE_SIGN_SHA256=${GARAGE_SIGN_SHA256}"
+EXTRA_OECMAKE = "-DCMAKE_BUILD_TYPE=Release -DAKTUALIZR_VERSION=${PV}"
+
+GARAGE_SIGN_OPS = "${@ '-DGARAGE_SIGN_VERSION=%s' % d.getVar('GARAGE_SIGN_VERSION') if d.getVar('GARAGE_SIGN_VERSION') is not None else ''} \
+                   ${@ '-DGARAGE_SIGN_SHA256=%s' % d.getVar('GARAGE_SIGN_SHA256') if d.getVar('GARAGE_SIGN_SHA256') is not None else ''} \
+                  "
+
+PACKAGECONFIG ?= "ostree ${@bb.utils.filter('DISTRO_FEATURES', 'systemd', d)} ${@bb.utils.filter('SOTA_CLIENT_FEATURES', 'hsm serialcan ubootenv', d)}"
```
agners: Using bb.utils.filter for `SOTA_CLIENT_FEATURES` is a nice touch here, I like that!

recipes-sota/aktualizr/aktualizr_git.bb
```diff
@@ -32,29 +27,34 @@ BRANCH ?= "master"
...
+PACKAGECONFIG[deb] = "-DBUILD_DEB=ON,-DBUILD_DEB=OFF,dpkg,"
+PACKAGECONFIG[hsm] = "-DBUILD_P11=ON,-DBUILD_P11=OFF,libp11,"
+PACKAGECONFIG[sota-tools] = "-DBUILD_SOTA_TOOLS=ON ${GARAGE_SIGN_OPS},-DBUILD_SOTA_TOOLS=OFF,glib-2.0,"
+PACKAGECONFIG[opcua] = "-DBUILD_OPCUA=ON,-DBUILD_OPCUA=OFF,boost,"
+PACKAGECONFIG[isotp] = "-DBUILD_ISOTP=ON,-DBUILD_ISOTP=OFF,"
+PACKAGECONFIG[systemd] = "-DBUILD_SYSTEMD=ON,-DBUILD_SYSTEMD=OFF,systemd,"
+PACKAGECONFIG[load-tests] = "-DBUILD_LOAD_TESTS=ON,-DBUILD_LOAD_TESTS=OFF,"
+PACKAGECONFIG[fault-injection] = "-DFAULT_INJECTION=ON,-DFAULT_INJECTION=OFF,libfiu"
+PACKAGECONFIG[serialcan] = ",,,slcand-start"
+PACKAGECONFIG[ubootenv] = ",,,u-boot-fw-utils aktualizr-uboot-env-rollback"
```
agners: I wonder whether really having all of them as PACKAGECONFIG makes sense in reality (e.g. warning-as-error, pedantic-warnings or isotp, are those really things we want to have configurable through PACKAGECONFIG?). But I guess they don't hurt...

recipes-sota/aktualizr/aktualizr_git.bb
```diff
@@ -97,12 +97,11 @@ do_install_append_class-native () {
     install -m 0644 ${B}/src/sota_tools/garage-sign/lib/* ${D}${libdir}
 }
 
-PACKAGES =+ " ${PN}-examples ${PN}-host-tools ${PN}-tools ${PN}-secondary "
+PACKAGES =+ "${PN}-examples ${PN}-host-tools ${PN}-tools ${PN}-secondary"
 
 FILES_${PN} = " \
                 ${bindir}/aktualizr \
                 ${bindir}/aktualizr-info \
-                ${bindir}/aktualizr-check-discovery \
```
agners: Is this no longer needed? Maybe a separate commit, seems a rather unrelated change.

recipes-sota/aktualizr/aktualizr_git.bb
liuming50: Yes, they are not necessary, but I was trying to make the PACKAGECONFIG aligned with the options() in CMakeFIle.txt of aktualizr source code, :)

agners: Ok I see. I do not have a strong opinion on whether to add all or only the necessary ones.

@patrickvacek what is your thought on this?

ricardosalveti: One bad side effect of having so many package config options is that DISTRO_FEATURES can end up having way too many options that might end up being specific to one single recipe (or generating unexpected side effects on other recipes that might be using the same string for a different purpose).

patrickvacek:
I would also prefer to see the pkgconfig change as a separate PR. The original commit looked fine, but I'm not sure I understand the pkgconfig benefits yet.

liuming50:
Hi,

I will close this pull request and resend two pull requests with feature branches, sorry for the inconvenience.

PR #471, opened 2019-02-03
Rocko aktualizr pkgconfig

liuming50:


lbonn:
Thanks for taking the trouble of opening the PRs to stable branches.

In general, there is no need to do it as we usually bundle this kind of changes in larger backport PRs. We open PRs for all branches to fast-track a more urgent fix (like bumping aktualizr's version recently) but in this case it could wait as it is mostly a refactor.

But no big deal, we can look into merging them after the inclusion in master.

lbonn:
Ok I see, thanks for the explanation. Then it makes good sense.

liuming50:
Changes in patch set V2:
1 Dropped some unused pkgconfig options.

PR #472, opened 2019-02-03
Sumo aktualizr pkgconfig

liuming50:


liuming50:
Changes in patch set V2:
1 Dropped some unused pkgconfig options.

PR #476, opened 2019-02-11
Bump aktualizr (master)

lbonn:
To 2aa9d93ccea09ec327789eecf858de561ef632da, which changed the installation result reporting format


lbonn:
Ok, prod is ready now

PR #477, opened 2019-02-11
Bump aktualizr (rocko)

lbonn:
To 2aa9d93ccea09ec327789eecf858de561ef632da, which changed the installation result reporting format

### DIFF
--- a/recipes-sota/aktualizr/aktualizr_git.bb
+++ b/recipes-sota/aktualizr/aktualizr_git.bb
@@ -27,7 +27,7 @@ SRC_URI = " \
   file://aktualizr-serialcan.service \
   "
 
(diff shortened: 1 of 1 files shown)


EpicNuts:
@lbonn It looks to be doing what we want it to do - I'm seeing the events in the UI when updating the secondaries. This should co-incide with an update to the documentation though, right? It's still using /storage/ for the virtsec.jsons. Post here when that's good to go and we can get this merged :)

lbonn:
Thanks! I'd like to wait for our backend changes to end up in main prod before merging that, otherwise we'll have some compatibility problems.

EpicNuts:
It hasn't changed since I tested it the last few days, right?
LGTM

PR #478, opened 2019-02-12
wic:plugins:otaimage.py: fix a potential issue

liuming50:
do_image_wic task is a standalone task that depending on
do_image_ota_ext4, so it's possible that do_image_wic runs
(taskhash contaminated) but do_image_ota_ext4 does not (taskhash not
contaminated), in which case, the otaimage would be in DEPLOY_DIR_IMAGE
instead of in IMGDEPLOYDIR, so we need check both of them.

Also, the logger.error is not supposed to raise a error, it just
prints out a error message, so we should use WicError.

And another typo is it checks IMGDEPLOYDIR but reports DEPLOY_DIR_IMAGE
not exist, also fix that.

### DIFF
--- a/scripts/lib/wic/plugins/source/otaimage.py
+++ b/scripts/lib/wic/plugins/source/otaimage.py
@@ -19,6 +19,7 @@ import logging
 import os
 import sys
 
+from wic import WicError
 from wic.plugins.source.rawcopy import RawCopyPlugin
 from wic.misc import get_bitbake_var
 
@@ -32,6 +33,18 @@ class OTAImagePlugin(RawCopyPlugin):
     name = 'otaimage'
 
     @classmethod
+    def _get_src_file(cls, image_dir_var):
+        """
+        Get OTA image file from image directory variable.
+        """
+        image_dir = get_bitbake_var(image_dir_var)
+        if not image_dir:
+            raise WicError("Couldn't find %s, exiting" % image_dir_var)
+
+        image_file = image_dir + "/" + get_bitbake_var("IMAGE_LINK_NAME") + ".ota-ext4"
+        return image_file if os.path.exists(image_file) else ""
+
+    @classmethod
     def do_prepare_partition(cls, part, source_params, cr, cr_workdir,
                              oe_builddir, bootimg_dir, kernel_dir,
                              rootfs_dir, native_sysroot):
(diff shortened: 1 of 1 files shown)


patrickvacek:
oe-selftest passed, but I've never had to touch that code before and I haven't booted a minnowboard in ages, so I'd like to either give that a try myself or get @OYTIS's stamp of approval.

OYTIS:
@liuming50 
Any idea how to reproduce the error scenario being fixed by this PR?

liuming50:
Hi, Patrick:

You probably could try the following to reproduce it:

1 INHERIT += "rm_work"
2 bitbake image-name
3 bitbake image-name -c do_image_wic -f

patrickvacek:
@liuming50 I've taken another look at this. I think it looks fine, and certainly fixing the error and typo are good, but I still don't understand how to reproduce the circumstances in which the do_image_wic task would run without do_image_ota_ext4 running first. I've tried to reproduce it with the steps you suggested and some playing around, and I haven't been able to simulate the error. How did you come across this issue? Thanks for your patience with this.

liuming50:
Hi, Patrick:

Indeed, this issue is not easy to be reproduced, and I think I gave you a incomplete reproduce steps last time, sorry for that, but technically, it could be reproduced by this:
```
1 INHERIT += "rm_work"
2 bitbake image-name
3 bitbake image-name -c clean
4 bitbake image-name
5 bitbake image-name -c do_image_wic -f
```
The key point is in steps 3 and 4, which will delete the ${IMAGEDEPLOY} directory and then install the artifacts to ${DEPLOY_DIR_IMAGE} from sstate, in which case, the do_image_wic would not be able to look for artifacts in ${IMAGEDEPLOY}.

patrickvacek:
@liuming50 Thanks for the more detailed steps. However, even with those, I can't reproduce the problem. Step 5 still causes `IMGDEPLOYDIR` to get repopulated. I do at least see that that directory is empty after step 4.

Interestingly, though, after step 5, the values repopulated in `IMGDEPLOYDIR` are newer than the values in `DEPLOY_DIR_IMAGE`. Your logic does the right thing in choosing `IMGDEPLOYDIR` first, so even if I can't reproduce the problem, I can tell that if it were to occur, your patch would cover that case. Hence, I think this is good to go. Thanks for your patience!

PR #481, opened 2019-02-14
Backport/rocko/garage sign lock etc

patrickvacek:
Backporting most of the changes to master from the last three months that haven't already been backported, with some exceptions:

1. Did not include the large refactoring of image_types_ostree. Maybe later if we need to.
2. Did not include most of the devicetree and FIT changes.
3. Kept the libp11 patch to deal with openssl, since rocko still uses openssl 1.0.2.
4. Did not include the usrmerge support, since ostree-initrd complained.
5. Did not include the softhsm upgrade, as there was a compilation error.
6. Did not include the reproducible build support, as that doesn't appear to be supported in rocko.

Somehow the custom ostree commit subject/body/summary changes didn't make it to rocko before, so I've also added them here.

Also did some reindentation of image_types_ota to make comparison with thud/master easier.


patrickvacek:
Passed oe-selftest.

patrickvacek:
Rebased on latest rocko. Still passes oe-selftest.

PR #489, opened 2019-02-21
image_types_ostree: Add a unique ref to fix simultaneous bitbaking.

patrickvacek:
To enable simultaneous bitbaking of two images with the same branch name, create
a new ref in the OSTree repo using the basename of the image.

As @OYTIS has pointed out, this probably won't solve every problem that might
come up with simultaneous bitbaking, such as if the two images use different
bootloaders. However, it does solve the immediate problem that comes up with our
primary-image + secondary-image example.


patrickvacek:
Passes oe-selftest.

PR #490, opened 2019-02-21
aktualizr-auto-prov-creds: remove API Gateway URL if it exists.

patrickvacek:
It will only be used by garage-sign.

See https://github.com/advancedtelematic/ota-plus-server/pull/1913.


patrickvacek:
Passes oe-selftest and has been manually tested to verify that the URL gets removed if present and nothing happens if it is not present.

PR #491, opened 2019-02-25
Add universal uEnv recipe

OYTIS:

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -70,9 +70,20 @@ If your board isn't supported yet, you can add board integration code yourself. 
 1.  Make the board boot into http://www.denx.de/wiki/U-Boot[U-Boot]
 2.  Make U-boot import variables from /boot/loader/uEnv.txt and load the kernel with initramfs and kernel command line arguments according to what is set in this file.
 
-You may take a look into https://github.com/advancedtelematic/meta-updater-minnowboard[Minnowboard] or https://github.com/advancedtelematic/meta-updater-raspberrypi[Raspberry Pi] integration layers for examples.
+There is a flexible universal U-boot script provided by meta-updater to do OSTree-enabled boot, `ota-u-boot-script`. It is not being deployed by default, so you should enable it in your machine configuration. This script can be configured with following variables:
 
-Although we have used U-Boot so far, other boot loaders can be configured work with OSTree as well.
+* `SOTA_BOOTLOADER_BOOTCOMMAND` - command U-Boot will use to load the kernel. Normally is either `bootm` or `bootz`.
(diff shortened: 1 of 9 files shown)


OYTIS:
It is an attempt to make a platform-independent uEnv.txt script supporting rollbacks and FIT images. The primary purpose here is that instead of pointing people integrating this layer with their boards to https://github.com/advancedtelematic/meta-updater-raspberrypi provide some framework that could be adjusted to different boards.

It's hard to tell right now how universal it really is, because the only U-Boot board we have in master is Raspberry Pi, but I hope it will not be hard to make some adjustments if problems arise.

I've tested it in the default mode, that is using FIT and rollback functionality, and with some adjustments without FIT as well (a few other variables need to be changed in sota_raspberrypi.bbclass, like the type of kernel image for instance). I didn't test the rollback functionality itself (yet). It would be also nice to extend it to support Qemu as well, the blocker here is that there is no boot partition in Qemu case now, `otaimg` gets booted directly.

Also it looks like the selftest for Grub is failing again, because I had to make `INITRAMFS_IMAGETYPES` an overlay again. I'll try to look into that.

OYTIS:
Update: rollback works.

OYTIS:
Configuration with rollback disabled works as well (apart from the rollback functionality of course). So only the Grub test issue is left.

agners:
Can we not just mandate using distro boot? Distro boot passes a lot of variables to the boot script which then can be reused.

ricardosalveti:
I also find quite weird to have defaults for values that are really platform specific, and most are already provided by the bootloader (and distro boot).

Shouldn't we just rely on what gets provided by the bootloader instead?

OYTIS:
@ricardosalveti @agners 

Thank you for your comments and sorry for the delay. As far as I remember, the FIT image for raspberrypi couldn't fit at `kernel_addr_r` and I had to relocate it to `ramdisk_addr_r`. Therefore I had that idea that ostree-related variables should be decoupled from distro boot variables.

Maybe that was a transient bug and is not really the case any more, I'll try to test that ASAP and if possible then would be happy to remove these extra variables.

OYTIS:
@ricardosalveti @agners @patrickvacek 

Sorry for the delay. Just rechecked - there's really a problem with loading FIT image to `kernel_addr_r` in that it overlaps with unpacked kernel. `ramdisk_addr_r` works well. So `SOTA_BOOTLOADER_*_ADDR` are apparently needed. One improvement I can suggest is to set them by default to distro boot counterparts.

lbonn:
I'll just close this for now as it hasn't been active for a long time and we still haven't agreed on how/whether to do this.

Feel free to reopen if needed.

OYTIS:
Nooooo!
Well, it wasn't a huge success anyway.

PR #493, opened 2019-02-26
Add meta-python layer to all platforms

lbonn:
Needed for meta-oe since this change:

http://git.openembedded.org/meta-openembedded/commit/?id=987d28c27e70ba12edf2b99745c13e7525206f4a

Note that existing build directories will need to be modified manually!


lbonn:
Thanks!

patrickvacek:
Added some reformatting and a fix for oe-selftest, which I'm running now, although I expect it to be fine. Also FYI @Raigi and @EpicNuts.

PR #497, opened 2019-02-26
Backport/thud/auto reboot etc

patrickvacek:
* Bump to latest aktualizr.
* Add recipe for automatic reboot after installation.
* More robust fix for simultaneous bitbaking.
* Remove API Gateway URL if present before putting credentials on a device.

Note that simply merging master into thud is no longer possible because of the meta-python change, which we don't want in thud.


patrickvacek:
Passes oe-selftest except for an unrelated rpi error I keep seeing on bagger...

PR #503, opened 2019-03-04
fit-conf: se PACKAGE_ARCH to ${MACHINE_ARCH}

liuming50:
fit-conf should be a machine specific package instead of being a arch
specific package.


patrickvacek:
This looks fine, by the way, but I haven't been able to test it yet. It's in the queue!

PR #504, opened 2019-03-05
Remove connman

lbonn:
connman is an annoying dependency which is not present in the base poky distribution.

The first step is to remove it from default poky-sota.conf and add it back on the various machines that we have.

The second step is to replace it with a simple systemd-networkd configuration on qemux86-64. We can migrate other platforms later.

One notable change is that users of the layer without systemd won't have a default network configuration on qemux86-64 but:

* we don't test anything without systemd
* users are expected to know better about their network configuration requirements than us

One even less intrusive alternative would be to add the `IMAGE_INSTALL_append` in the default local.conf generated by envsetup.sh. But since I wanted to do it on a per-machine basis for now, it requires a bit more changes.

**oe-selftest ongoing, I'll try to push the changes for raspberrypi as well if everything works fine**

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -74,6 +74,8 @@ You may take a look into https://github.com/advancedtelematic/meta-updater-minno
 
 Although we have used U-Boot so far, other boot loaders can be configured work with OSTree as well.
 
+Your images will also need network connectivity to be able to reach an actual OTA backend. Our 'poky-sota' distribution does not mandate or install a default network manager but our supported platforms use the `virtual/network-configuration` recipe, which can be used as a starting example.
+
 == SOTA-related variables in local.conf
 
 * `OSTREE_REPO` - path to your OSTree repository. Defaults to `$\{DEPLOY_DIR_IMAGE}/ostree_repo`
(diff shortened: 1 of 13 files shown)


lbonn:
Added bonus: about 17MB (18%!) less on a qemux86-64 rootfs (measured on uncompressed core-image-minimal-qemux86-64.tar.bz2).

lbonn:
Tweaked a bit to define virtual/network-configuration, provided by connman and the new recipe.
It's added by default in our platforms but not on poky-sota distro.

lbonn:
Ok it's become a bit more complicated but works on my rpi. `Driver=smsc95xx` might be a better matching condition, will look into it.

lbonn:
Hm they changed the network controller on raspberrypi3 B+ (also supported under the raspberrypi3 machine), so the driver matching won't work

lbonn:
Disregard, the last stroke-through comment: several drivers can be listed for device matching.

I've tested it on qemu and raspberrypi3, works as expected!
Also added a small documentation bit for integrators.

PR #505, opened 2019-03-09
aktualizr: use echo -e when using escape sequences

agners:
Some shell require the -e argument when using escape sequences
in echo. This has been observed when building images on a
Fedora machine, where the configuration files ended up including
"\n" sequences instead of newlines.


lbonn:
Looks like it causes problems outside of Fedora. I got that installed on a qemu device:

```
-e [provision]
primary_ecu_hardware_id = qemux86-64
```

The preferred method seems to be to use `printf`, as done in occasions in poky recipes. It usually deviates less from posix.

Will open a quick PR asap.

PR #506, opened 2019-03-11
Use 'printf' instead of 'echo -e'

lbonn:
Should be less surprising

Rework of https://github.com/advancedtelematic/meta-updater/pull/505 which didn't quite work on my environment.


agners:
Hm, I did see that `echo -e` is used in other instances in oe-core too, but maybe with quotes? (e.g. `echo -e "test\n"`). I am currently travelling so I can't check right now.

lbonn:
To be frank I'm not really sure why it behaved like that as I would have expected it to work as well. But printf seems safe.

PR #507, opened 2019-03-12
aktualizr resource control

lbonn:
I've done the logic inside the main aktualizr recipe (as opposed to, another .bb), so that it's easily configurable with `RESOURCE_xxx_pn-aktulizr`, which I find quite handy.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -228,7 +228,7 @@ sudo apt install ovmf
 5. Run oe-selftest:
 +
 ```
-oe-selftest --run-tests updater
+oe-selftest -r updater_native updater_qemux86_64 updater_minnowboard updater_raspberrypi
 ```
 
 For more information about oe-selftest, including details about how to run individual test modules or classes, please refer to the https://wiki.yoctoproject.org/wiki/Oe-selftest[Yocto Project wiki].
--- a/README.adoc
+++ b/README.adoc
@@ -95,6 +95,7 @@ Your images will also need network connectivity to be able to reach an actual OT
 * `SOTA_MAIN_DTB` - base device tree to use with the kernel. Used together with FIT images. You can change it, and the device tree will also be changed after the update.
 * `SOTA_DT_OVERLAYS` - whitespace-separated list of used device tree overlays for FIT image. This list is OSTree-updateable as well.
 * `SOTA_EXTRA_CONF_FRAGS` - extra https://lxr.missinglinkelectronics.com/uboot/doc/uImage.FIT/overlay-fdt-boot.txt[configuration fragments] for FIT image.
+* `RESOURCE_xxx_pn-aktualizr` - controls maximum resource usage of the aktualizr service, when `aktualizr-resource-control` is installed on the image. See <<aktualizr service resource control>> for details.
 
 == Usage
 
@@ -159,6 +160,19 @@ Second, you can write recipes to install additional config files with customized
 * link:recipes-sota/config/aktualizr-log-debug.bb[aktualizr-log-debug.bb] sets the log level of aktualizr to 0 (trace). The default is 2 (info). This recipe is intended for development and debugging purposes.
 
 To use these recipes, you will need to add them to your image with a line such as `IMAGE_INSTALL_append = " aktualizr-log-debug "` in your `local.conf`.
+
+=== aktualizr service resource control
+
+With systemd based images, it is possible to set resource policies for the aktualizr service. The main use case is to provide a safeguard against resource exhaustion during an unforeseen failure scenario.
+
+To enable it, install `aktualizr-resource-control` on the target image and optionally override the default resource limits set in link:recipes-sota/aktualizr/aktualizr_git.bb[aktualizr_git.bb], from your `local.conf`.
+
+For example:
+
+....
+IMAGE_INSTALL_append += " aktualizr-resource-control "
+RESOURCE_CPU_WEIGHT_pn-aktualizr = "50"
+....
 
 == Development configuration
 
--- a/lib/oeqa/selftest/cases/updater_qemux86_64.py
+++ b/lib/oeqa/selftest/cases/updater_qemux86_64.py
@@ -424,4 +424,49 @@ class PrimaryTests(OESelftestTestCase):
         self.assertEqual(retcode, 0, "Unable to run aktualizr --help")
         self.assertEqual(stderr, b'', 'Error: ' + stderr.decode())
 
+
+class ResourceControlTests(OESelftestTestCase):
+    def setUpLocal(self):
+        layer = "meta-updater-qemux86-64"
+        result = runCmd('bitbake-layers show-layers')
+        if re.search(layer, result.output) is None:
+            # Assume the directory layout for finding other layers. We could also
+            # make assumptions by using 'show-layers', but either way, if the
+            # layers we need aren't where we expect them, we are out of luck.
+            path = os.path.abspath(os.path.dirname(__file__))
+            metadir = path + "/../../../../../"
+            self.meta_qemu = metadir + layer
+            runCmd('bitbake-layers add-layer "%s"' % self.meta_qemu)
+        else:
+            self.meta_qemu = None
+        self.append_config('MACHINE = "qemux86-64"')
+        self.append_config('SOTA_CLIENT_PROV = " aktualizr-auto-prov "')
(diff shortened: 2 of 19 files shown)


patrickvacek:
Seems cool! I still need to try it out, though.

recipes-sota/aktualizr/files/10-resource-control.conf
```diff
@@ -0,0 +1,6 @@
+[Service]
+CPUAccouting=true
```
lbonn: oops made a typo here

patrickvacek:
Passes oe-selftest and appears to work when I manually tested it. However, as discussed offline:
1. It's hard to know when any why these limits kick in.
2. This looks like a great candidate for an oe-selftest case.

lbonn:
I've rebased and had to tweak it slightly. Namely fixing a typo, and removing the `RRECOMMENDS`, as it's actually make it on by default and was causing problems with ptest.

lbonn:
Unfortunately, I haven't found any doc pertaining to the kill reason of a service but would love to see it if it exists.
As for the oe-selftest, are you advocating for adding it in this PR.

To illustrate a bit, on a running qemu image:

```
root@qemux86-64:/sysroot/home/root# systemctl status aktualizr
● aktualizr.service - Aktualizr SOTA Client
   Loaded: loaded (/usr/lib/systemd/system/aktualizr.service; enabled; vendor preset: enabled)
  Drop-In: /usr/lib/systemd/system/aktualizr.service.d
           └─10-resource-control.conf
   Active: active (running) since Wed 2019-03-13 15:14:26 UTC; 10min ago
 Main PID: 204 (aktualizr)
    Tasks: 4 (limit: 1175)
   Memory: 16.0M (high: 100.0M max: 789.1M)
      CPU: 3.967s
   CGroup: /system.slice/aktualizr.service
           └─204 /usr/bin/aktualizr
```

patrickvacek:
>  As for the oe-selftest, are you advocating for adding it in this PR?

Yeah, I don't think it should be hard to write. Any feature we write into meta-updater that is testable with oe-selftest should be tested. We should have done that for the custom hardware ID logic, for example, but I didn't think about it at the time.

We should probably even list these actions somewhere, possibly in aktualizr's actions.md if that's not too annoying. But I'll let that topic go for now!

lbonn:
Yes makes total sense. Ptest could also be added as well but it's maybe failing right now.

I also had the project of splitting our selftests by target machines so that it's a bit easier to run a useful subset while saving a lot of disk space and cpu time.

patrickvacek:
> Ptest could also be added as well but it's maybe failing right now.

Agreed, but I'd say that's a task for another day. :)

> I also had the project of splitting our selftests by target machines so that it's a bit easier to run a useful subset while saving a lot of disk space and cpu time.

What do you envision there? Is the current grouping insufficient?

lbonn:
> > I also had the project of splitting our selftests by target machines so that it's a bit easier to run a useful subset while saving a lot of disk space and cpu time.
> 
> What do you envision there? Is the current grouping insufficient?

Pretty much IMO. I think it would make sense to single-out three/four CI jobs for the tests we have right now:

- "updater" tests running on qemux86-64
- "updater" tests running on emulated minnowboard
- "updater" rpi build test
- (future) qemux86-64 ptest

With current oe-selftests, it's a bit difficult.

It's looking good on paper:

```
-R SKIPS [SKIPS ...], --skip-tests SKIPS [SKIPS ...]
                        Run all (unhidden) tests except the ones specified.
                        Format should be <module>[.<class>[.<test_method>]]
-r RUN_TESTS [RUN_TESTS ...], --run-tests RUN_TESTS [RUN_TESTS ...]
                        Select what tests to run (modules, classes or test
                        methods). Format should be:
                        <module>.<class>.<test_method>
```

But in practice, it is quite useless:

```
$ oe-selftest -r updater -R RpiTests
oe-selftest: error: argument -R/--skip-tests: not allowed with argument -r/--run-tests
...
```

So, to run all updater tests without raspberrypi, we need to have a big brittle list of test groups listed with `-r`.

If we had two .py files updater-qemu.py and updater-rpi.py, it would solve this problem.

patrickvacek:
> Pretty much IMO. I think it would make sense to single-out three/four CI jobs for the tests we have right now

I understand now, thanks. Makes sense to me. The reason for the grouping was based mostly around convenience of development, limited by my knowledge of python/bitbake/oe-selftest/qemu/etc. Multiple files might be easier than deep-diving into all of that, though. Up to you!

lbonn:
Added an oe-seltest + made the split.

I need to re-run the full suite after the split.

README.adoc
```diff
@@ -228,7 +228,7 @@ sudo apt install ovmf
 5. Run oe-selftest:
 +
 ```
-oe-selftest --run-tests updater
+oe-selftest -r updater_native -r updater_qemux86_64 -r updater_minnowboard -r updater_raspberrypi
```
patrickvacek: I don't think you have to specify `-r` each time.

lib/oeqa/selftest/cases/updater_qemux86_64.py
```diff
@@ -606,99 +425,43 @@ def test_aktualizr_present(self):
...
+    def test_aktualizr_resource_control(self):
+        print('Checking aktualizr was killed')
+        stdout, stderr, retcode = self.qemu_command('systemctl --no-pager show --property=ExecMainStatus aktualizr')
+        self.assertIn(b'ExecMainStatus=9', stdout, 'Aktualizr was not killed')
+
+        self.qemu_command('systemctl --runtime set-property aktualizr MemoryMax=')
+        self.qemu_command('systemctl restart aktualizr')
 
+        stdout, stderr, retcode = self.qemu_command('systemctl --no-pager show --property=ExecMainStatus aktualizr')
+        self.assertIn(b'ExecMainStatus=0', stdout, 'Aktualizr did not restart')
```
patrickvacek: Cool, that's great! Do you think it's worth testing all the properties? One is probably enough to prove that the process works, I'm just wondering.

lbonn:
I had problem with running oe-selftests because of the rng initialization taking a minute and blocking aktualizr startup. But I've pushed a fix in the last commit and they seem to run mostly fine, but my repo sync is not fresh from today.

Let's see if I broke anything with my checkout, then we should be able to test that with fresh commits.

lib/oeqa/selftest/cases/updater_qemux86_64.py
lbonn: You can see all the available properties, by just running `systemctl show aktualizr`. Their meaning is not always immediately obvious but ExecMainStatus seems to be enough for this purpose.

README.adoc
lbonn: ok I'll try it

lib/oeqa/selftest/cases/updater_qemux86_64.py
patrickvacek: I was just wondering if it made sense to set each of the properties that you made a variable for and confirming that systemd is interpreting them correctly.

patrickvacek:
Not the fault of your PR, but is the autoprov test checking aktualizr-info twice? That seems like an accidental artifact of some refactoring from a while back. But it might be worth cleaning up if I'm reading it right.

lib/oeqa/selftest/cases/updater_qemux86_64.py
patrickvacek: I ran oe-selftest and it passed. Looks good to me, but I have one request: in addition to `RESOURCE_MEMORY_MAX_pn-aktualizr`, set the other two variables that you've hooked up (probably to generous/sane values), and then do the `systemctl --no-pager show --property=ResourceMemoryMax aktualizr` check for all three variables to check that the values match. I'd like to make sure the parsing of the configuration works as we expect. These things are easy to break without noticing it otherwise.

lbonn: Ok I've added that in the last push (getting all the properties at once)

patrickvacek: Perfect, thanks!

lbonn:
> Not the fault of your PR, but is the autoprov test checking aktualizr-info twice? That seems like an accidental artifact of some refactoring from a while back. But it might be worth cleaning up if I'm reading it right.

Pushed one commit to address that.

PR #509, opened 2019-04-03
layer.conf: Add warrior to compatible release series

liuming50:


liuming50:
Hi, Patrick:

I would like to suggest we keep 'thud' for a while since OE just bump to
warrior, firstly, because we do support both thud and warrior,  secondly,
some users of meta-updater master branch, they might still use the thud
branch of OE and other meta- layers in their repo XML files, so they do not
have to bump OE to warrior to build with meta-updater master branch.


Patrick Vacek <notifications@github.com> 於 2019年4月4日 週四 上午9:18寫道：

PR #368, opened 2018-08-09
Fit initramfs support

cajun-rat:

PR #511, opened 2019-04-17
Fix non-sota build issue in rocko

cminyard:
Well, I guess testing non-sota builds is not something that's done very much :-).  You get an error when you do this.  This patch seems to fix the issue, though I'm not 100% sure it's the right fix.

### DIFF
--- a/classes/sota.bbclass
+++ b/classes/sota.bbclass
@@ -23,8 +23,6 @@ WKS_FILE_sota ?= "sdimage-sota.wks"
 
 EXTRA_IMAGEDEPENDS_append_sota = " parted-native mtools-native dosfstools-native"
 
-OSTREE_INITRAMFS_FSTYPES ??= "${@oe.utils.ifelse(d.getVar('OSTREE_BOOTLOADER', True) == 'u-boot', 'ext4.gz.u-boot', 'ext4.gz')}"
-
 # Please redefine OSTREE_REPO in order to have a persistent OSTree repo
 export OSTREE_REPO ?= "${DEPLOY_DIR_IMAGE}/ostree_repo"
 export OSTREE_BRANCHNAME ?= "${SOTA_HARDWARE_ID}"
--- a/recipes-core/images/initramfs-ostree-image.bb
+++ b/recipes-core/images/initramfs-ostree-image.bb
@@ -13,6 +13,8 @@ IMAGE_LINGUAS = ""
 
 LICENSE = "MIT"
 
+OSTREE_INITRAMFS_FSTYPES ??= "${@oe.utils.ifelse(d.getVar('OSTREE_BOOTLOADER', True) == 'u-boot', 'ext4.gz.u-boot', 'ext4.gz')}"
+
 IMAGE_FSTYPES = "${OSTREE_INITRAMFS_FSTYPES}"
 
 inherit core-image


cminyard:
On Thu, Apr 18, 2019 at 07:10:50AM +0000, Patrick Vacek wrote:
> patrickvacek commented on this pull request.
> 
> This might work, but then we've largely rendered `OSTREE_INITRAMFS_FSTYPES` redundant, so we should go further. In fact, on master (and thud), this was already done by another contributor in https://github.com/advancedtelematic/meta-updater/commit/9dcfcdb9aa2da833a3b86b87fbbb22b60e5d7e3c (part of https://github.com/advancedtelematic/meta-updater/pull/368). I'd considered backporting that whole PR into rocko and sumo, but because of the complexity and differences in the branches, it seemed like too much work. But if this is important for you, I'd rather pursue that and try to keep the branches as similar to one another as possible. Is that something you'd be willing to try?

Sure, that's not a problem.

> 
> On another note, what is your motivation to use the meta-updater layer with the `sota` feature? It is true we don't test that, because the only reason we have the layer is to use the `sota` feature!

We provide a general-purpose distro, some customer may use this, most will not.

We still haven't decided on whether to use this, I'm just "kicking the tires" and reporting things as I find them.

-corey

patrickvacek:
@cminyard Sounds good on both accounts. I appreciate you reaching out while looking around. If you have questions about the backport (or anything else), let us know.

patrickvacek:
@cminyard Is this still something you're interested in? On second thought, we probably don't want to backport #368 as that causes breaking changes to existing devices. We've also just moved rocko (and sumo) to the backburner; thud is now the oldest release branch we're actively maintaining. If you'd still like to see this in rocko (and sumo), we can test it and take in under consideration.

cminyard:
On Fri, Sep 06, 2019 at 05:27:17AM -0700, Patrick Vacek wrote:
> @cminyard Is this still something you're interested in? On second thought, we probably don't want to backport #368 as that causes breaking changes to existing devices. We've also just moved rocko (and sumo) to the backburner; thud is now the oldest release branch we're actively maintaining. If you'd still like to see this in rocko (and sumo), we can test it and take in under consideration.

I would like this fixed, if possible.  Plus others that might integrate
the layers might run into the same issue, so I would think it's
something you want, too.

-corey

patrickvacek:
After looking into this again, I don't think this is actually an appropriate fix. `image_types_ostree.bbclass` also depends on `OSTREE_INITRAMFS_FSTYPES`, so if you remove it from `sota.bbclass`, it won't know about it. This problem is also present (but slightly more complicated) in master/zeus/warrior/thud, although we got rid of `OSTREE_INITRAMFS_FSTYPES` and just use `INITRAMFS_FSTYPES` directly.

It is still interesting to me know why you are interested in meta-updater but without the `sota` feature. What are you trying to do? Maybe if I understand your use case better we can find a good solution. In the meantime, can you just set `OSTREE_INITRAMFS_FSTYPES` in your local.conf or elsewhere?

In the meantime since you made this PR, we now have a CI, so if you have suggestions, we can try them out. However, as it is now, I would expect it to fail.

patrickvacek:
Closing due to inactivity and unsuitability of the currently proposed fix. However, if this becomes relevant again, please don't hesitate to reopen this or create a new PR!

PR #512, opened 2019-04-23
aktualizr: Bump to latest for recovery fixes.

patrickvacek:


patrickvacek:
BTW I'm still waiting on oe-selftest; I think the problems I'm seeing are specific to the machines I'm running the tests on, but I'd like to be more confident in that before merging. oe-selftest succeeding with latest master also depends on https://github.com/advancedtelematic/meta-updater-raspberrypi/pull/51.

patrickvacek:
Finally got oe-selftest to pass on my laptop by being careful with which versions of the various layers I used. @lbonn are you okay with the last commit I added?

lib/oeqa/selftest/cases/updater_qemux86_64.py
```diff
@@ -395,7 +395,7 @@ def setUpLocal(self):
         self.append_config('IMAGE_INSTALL_append += " aktualizr-resource-control "')
         self.append_config('RESOURCE_CPU_WEIGHT_pn-aktualizr = "1000"')
         self.append_config('RESOURCE_MEMORY_HIGH_pn-aktualizr = "50M"')
-        self.append_config('RESOURCE_MEMORY_MAX_pn-aktualizr = "1M"')
+        #self.append_config('RESOURCE_MEMORY_MAX_pn-aktualizr = "1M"')
```
lbonn: Is it intended?

patrickvacek: D'oh! Nope. Good catch. I'm glad I asked you to re-review. I was double-checking that it failed as expected. Anyway, fixed now.

PR #515, opened 2019-05-03
Fix/sumo/backport

patrickvacek:


recipes-sota/aktualizr/files/libcrypto_valgrind_supp.patch
```diff
@@ -0,0 +1,25 @@
+diff --git a/tests/aktualizr.supp b/tests/aktualizr.supp
+index 51aae5c5..2b65f5c0 100644
+--- a/tests/aktualizr.supp
++++ b/tests/aktualizr.supp
+@@ -90,3 +90,19 @@
+    fun:ostree_sysroot_simple_write_deployment
+    ...
+ }
++{
++   libcrypto.so.1.0.2 false positives (see e.g. https://github.com/scikit-hep/pyjet/blob/master/etc/valgrind-python.supp)
```
lbonn: Maybe link to something more explanatory, like https://github.com/curl/curl/issues/1117 or http://openssl.6102.n7.nabble.com/PURIFY-flag-in-md-rand-c-and-a-strongly-worded-comment-td41674.html.

Also, I don't think there will be an Openssl 1.0.3,  `obj:/usr/lib/libcrypto.so.1.0.*` shouldn't hurt

lib/oeqa/selftest/cases/updater_qemux86_64_ptest.py
```diff
@@ -0,0 +1,52 @@
...
+
+    def qemu_command(self, command, timeout=60):
+        return qemu_send_command(self.qemu.ssh_port, command, timeout=timeout)
+
+    def test_run_ptests(self):
+        # simulate a login shell, so that /usr/sbin is in $PATH (from /etc/profile)
+        stdout, stderr, retcode = self.qemu_command('sh -l -c ptest-runner', timeout=None)
+        output = stdout.decode()
+        print(output)
+        self.assertEqual(retcode, 0)
```
lbonn: The log not showing might be cause by this assert breaking. I'll try to see if moving it helps

lbonn: Hm actually it looks like a timeout triggered by ptest-runner when I tried it. And ctest first dumps its output in `/tmp/aktualizr-ptest.log.tmp` before renaming it to its final name, presumably at the end.

So, if ctest is killed, it would still keep the old name.

patrickvacek: Yeah, I've been seeing that timeout, too. Somehow it isn't a problem on master/thud.

lbonn: In my case, the only test that doesn't terminate is `test_fetcher`.

lbonn:
```
Traceback (most recent call last):
  File "tests/sota_tools/treehub_server.py", line 14, in <module>
    from pathlib import Path
ImportError: No module named 'pathlib'
```

That's a bit annoying that the test doesn't terminate after that... But at least it should be fixable.

lbonn:
#518 has two commits to help with ptest: failure of test_fetcher and fix for my buggy "More robust aktualizr-ptest printing of errors"

patrickvacek:
Finally ready with this. Passes oe-selftest but not ptest, which Laurent and I agreed is not importantly for now anyway.

PR #518, opened 2019-05-07
Make aktualizr-ptest depend on python3-misc

lbonn:
Some of our tests use pathlib which ends up in there...

Seems to be a problem on sumo at least, don't know how we haven't noticed on master... Still running tests to confirm!


lbonn:
Ok seems to fix test_fetcher. Also, my #516 PR was buggy sorry, here is a fix.

patrickvacek:
Looks fine. Have you run oe-selftest? I just started a run, although it looks like it should be fine.

lbonn:
I haven't run the test in full but verified that it fixed test_fetcher + printing of ptest output.

PR #519, opened 2019-05-13
OTA-2606: Renaming provisioning methods

merltron:
Updating links to documentation with proper terminology
rather than aktualiz repo topics (aktualizr repo also now has a mirror of
the same topic).


README.adoc
```diff
@@ -88,7 +88,7 @@ Your images will also need network connectivity to be able to reach an actual OT
...
 * `SOTA_DEPLOY_CREDENTIALS` - when set to '1' (default value), deploys credentials to the built image. Override it in `local.conf` to built a generic image that can be provisioned manually after the build.
-* `SOTA_CLIENT_PROV` - which provisioning method to use. Valid options are https://github.com/advancedtelematic/aktualizr/blob/master/docs/automatic-provisioning.adoc[`aktualizr-auto-prov`], https://github.com/advancedtelematic/aktualizr/blob/master/docs/implicit-provisioning.adoc[`aktualizr-ca-implicit-prov`], and https://github.com/advancedtelematic/aktualizr/blob/master/docs/hsm-provisioning.adoc[`aktualizr-hsm-prov`]. The default is `aktualizr-auto-prov`. This can also be set to an empty string to avoid using a provisioning recipe.
+* `SOTA_CLIENT_PROV` - which provisioning method to use. Valid options are [`aktualizr-auto-prov`], [`aktualizr-ca-implicit-prov`], and [`aktualizr-hsm-prov`]. For more information on these provisioning methods, see the https://docs.ota.here.com/client-config/client-provisioning-methods.html[OTA Connect documentation]. The default is `aktualizr-auto-prov`. This can also be set to an empty string to avoid using a provisioning recipe.
```
patrickvacek: What's the point of keeping the brackets if there is no link anymore? I think it makes more sense to fix the things in aktualizr first and then come back and fix the links here.

There are also a lot of code changes to take care of as part of this rename. Do you want me to work on that, and would you prefer to keep that as a separate branch/PR?

xaviergully: @patrickvacek, could we get rid of '_auto_' and '_implicit_' word for the configuration options ?

patrickvacek: Yes, that is the plan. However, it isn't quite as easy as just changing it, we have to update tests and make sure everything still works. That first requires updating aktualizr and then coming back to this here.

merltron: Sorry the brackets were an oversight. Fixed now. Other issues to be handled in separate PR:

merltron:
@patrickvacek I removed the brackets - is another review necessary?

patrickvacek:
> Configuration options should be updated as well

Yes, that will come in a separate PR when I or someone else gets around to it. That requires testing; this does not.

PR #521, opened 2019-05-17
aktualizr-polling-interval: new config fragment.

patrickvacek:
Can be used for testing purposes after we increase the default polling interval in aktualizr. We no longer recommend anything less than an hour for production use cases, but it's still convenient to poll more frequently while testing.

FYI @alexhumphreys and @koshelev. This should probably get merged and backported (and the quickstart guide updated) before we merge https://github.com/advancedtelematic/aktualizr/pull/1212.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -96,6 +96,7 @@ Your images will also need network connectivity to be able to reach an actual OT
 * `SOTA_DT_OVERLAYS` - whitespace-separated list of used device tree overlays for FIT image. This list is OSTree-updateable as well.
 * `SOTA_EXTRA_CONF_FRAGS` - extra https://lxr.missinglinkelectronics.com/uboot/doc/uImage.FIT/overlay-fdt-boot.txt[configuration fragments] for FIT image.
 * `RESOURCE_xxx_pn-aktualizr` - controls maximum resource usage of the aktualizr service, when `aktualizr-resource-control` is installed on the image. See <<aktualizr service resource control>> for details.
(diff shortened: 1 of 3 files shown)


recipes-sota/config/files/60-test-polling-interval.toml
```diff
@@ -0,0 +1,2 @@
+[uptane]
+polling_sec = 30
```
lbonn: Wouldn't we want the actual value to be parametrizable, like here https://github.com/advancedtelematic/meta-updater/blob/master/recipes-sota/aktualizr/files/10-resource-control.conf?

patrickvacek: Not a bad idea. Maybe 30 should be the default and then have a variable that the user can control?

lbonn: Yes, that's basically what I had in mind, we have the same mechanism there https://github.com/advancedtelematic/meta-updater/blob/550ef5c1c32be58cca91f80afbedc23d5e9b79b6/recipes-sota/aktualizr/aktualizr_git.bb#L65

patrickvacek: @lbonn Done, I think. What do you think now?

patrickvacek:
Passed oe-selftest, merging. Now time to talk about docs with @christianhuggerATS and whoever else is interested.

PR #522, opened 2019-05-21
sota.bbclass: move sota to DISTROOVERRIDES

liuming50:
Move sota to DISTROOVERRIDES from OVERRIDES, it should be a distro
overrides.

This change also let 'sota' to be in front of 'forcevariable' in
OVERRIDES, since 'forcevariable' should always be the last overrides.


patrickvacek:
Seems fine, but I'm still trying to test it, and master has been so unreliable lately that this is taking some time. Thanks for the PR and for being patient!

PR #523, opened 2019-05-21
Log persistence

lbonn:
Here is a sample configuration to make logs persistent between reboots. They are stored in systemd's binary format in /var/log/journal.

I don't quite know where to advertise this, it's not directly related to meta-updater's functionality and is something that we'd expect image writers to configure themselves according to their need. So for now, it just sits there as a sample.


patrickvacek:
Why not enable it by default?

lbonn:
Well, these are embedded devices, we don't know how much space customers want to dedicate to logs or even if they should avoid logging as much as possible to not wear flash.
Having it on by default could surprise users and removing a recipe from an image is a bit uglier than adding it.

patrickvacek:
> we don't know how much space customers want to dedicate to logs or even if they should avoid logging as much as possible to not wear flash.
> Having it on by default could surprise users and removing a recipe from an image is a bit uglier than adding it.

True. But then I still think something is missing. Ideally, this would always be enabled during development and testing and then optional during production. Some users will probably never want it because of the space concerns, but we (in our support role) want all users to always have it on to make it easier to help debug problems encountered anywhere in the process (development, factory, or field). So maybe it should be specifically encouraged in the docs (both here and the official docs)?

I wonder if it makes sense to have some sort of default dev/test image or configuration separate from a recommended/reference production image/config.

lbonn:
> I wonder if it makes sense to have some sort of default dev/test image or configuration separate from a recommended/reference production image/config.

I like the idea. One objection would be that it would, once again, increase the space of things that we need to test.

patrickvacek:
> One objection would be that it would, once again, increase the space of things that we need to test.

True. Maybe for now we should just suggest very early in the README that users add this and aktualizr-log-debug to their images during development and testing. And maybe do the same in the official docs, too.

lbonn:
New local.conf now contains more sample configuration + explanatory comments. We still might want to have it mentioned in other docs, I'm just not sure which.

conf/local.conf.sample.append
```diff
@@ -0,0 +1,30 @@
...
+IMAGE_INSTALL_append += " systemd-journald-persistent"
+
+# Set resource limits for aktualizr service
+#
+# Uncomment these lines to change the default parameters.
+#
+#RESOURCE_CPU_WEIGHT_pn-aktualizr = "100"
+#RESOURCE_MEMORY_HIGH_pn-aktualizr = "100M"
+#RESOURCE_MEMORY_MAX_pn-aktualizr = "80%"
+IMAGE_INSTALL_append += " aktualizr-resource-control"
```
patrickvacek: Does it make sense to keep `aktualizr-resource-control` uncommented even if none of the variables are used?

Also, does it makes sense to have a commented-out line for adding `aktualizr-log-debug`, perhaps in the same area as the log persistence?

lbonn: The commented lines just reflect the defaults that are in aktualizr_git.bb, that's just a convenience for easily changing these values.

Agree for adding aktualizr-log-debug.

PR #524, opened 2019-05-28
Refactor/ota 2798/rename provisioning

patrickvacek:
Depends on https://github.com/advancedtelematic/aktualizr/pull/1220.


patrickvacek:
Passes oe-selftest except for the `test_aktualizr_fullostree` ptest as expected. Ready for review.

classes/sota.bbclass
```diff
@@ -7,6 +7,17 @@ SOTA_CLIENT_PROV ??= "aktualizr-shared-prov"
 SOTA_DEPLOY_CREDENTIALS ?= "1"
 SOTA_HARDWARE_ID ??= "${MACHINE}"
 
+# Translate old provisioning recipe names into the new versions.
```
lbonn: Oh great! Can it also show an advisory warning "name has changed, please be aware..."?

patrickvacek: Yes! Just need to figure out how to do it in python... otherwise I can try redoing this bit in bash, since I think `bbwarn` does the trick.

patrickvacek: Finally got around to doing it. It does print a ton of identical warnings, though, but I can't figure out a good way to avoid that.

lbonn: Haha nice. Duplicate warnings are probably fine.

PR #525, opened 2019-05-28
aktualizr: bump garage sign to version 0.7.0-3-gf5ba640.

patrickvacek:
Not urgent, just seemed like it was time to do it before we do the next round of backporting.


patrickvacek:
Passes oe-selftest BTW.

PR #526, opened 2019-06-05
Refactor/thud/renaming secondaries etc

patrickvacek:


patrickvacek:
Passes oe-selftest except for the ostree test (a known issue).

PR #529, opened 2019-06-06
Add support for raspberrypi3-64

lbonn:
Also needs change in meta-updater-raspberrypi, forthcoming!

Currently building on master for a final test, as it was initially developed on rocko.


lbonn:
Actually having problem building on master right now unfortunately. We could indeed add oe-selftests as well later.

lbonn:
Boots with the last changes!

PR #530, opened 2019-06-13
OTA-2541: Primary and Secondary network configuration in case of RPi

mike-sul:
Signed-off-by: Mike Sul <ext-mykhaylo.sul@here.com>


recipes-test/demo-network-config/network-config.inc
```diff
@@ -0,0 +1,19 @@
+CONF_TYPE ?= "${@ 'multihomed' if d.getVar('MACHINE') == 'raspberrypi3' and d.getVar('IMAGE_TYPE', '') == 'primary' else 'static'}"
```
patrickvacek: Why the empty string for the second parameter of `d.getVar('IMAGE_TYPE', '')`?

recipes-test/demo-network-config/files/26-multihomed-client.network
```diff
@@ -0,0 +1,9 @@
+[Match]
+Name=@IFNAME@
+
+[Network]
+Description=Multihomed network. DHCP-assigned IP for Primary<->Backend. Statically assigned IP for Primary<->Secondary
```
patrickvacek: How is multihoming actually being used here? I'm not sure I'm following it.

recipes-test/demo-network-config/network-config.inc
mike-sul: The empty string is used as a default value of 'IMAGE_TYPE' in case if it's not defined.  If the default value is not specified (d.getVar('IMAGE_TYPE')) then an error will be yielded if 'IMAGE_TYPE' is not defined.

recipes-test/demo-network-config/files/26-multihomed-client.network
mike-sul: Two IP addresses are assigned two a single RPi Ethernet NIC. One is statically defined (10.0.3.1) and  aimed for communication with secondaries, the second is obtained from DHCP server running on a local router that has an access to Inet thus it allows Primary to communicate with the server.

recipes-test/demo-network-config/network-config.inc
mike-sul: I've just replaced d.getVar('IMAGE_TYPE', '') with d.getVar('IMAGE_TYPE') as it doesn't bring any value here.

recipes-test/demo-network-config/files/26-multihomed-client.network
patrickvacek: Thanks, I'm starting to see it more clearly. But in the RPi case, if you only have one ethernet connection and you aren't using wifi, how do you access multiple networks? Are you plugging the primary and secondary RPis both into the same router or switch so that they can talk to each other through that without leaving the local network?

mike-sul: Yes, I plug them into the same switch that works just on data link layer (layer 2) so it doesn't impact IP level anyhow which allows my to configure as many as I want IP networks  for devices connected to it. I'll prepare some doc depicting networking topology for this setup later on.

mike-sul:
Updated this PR to support WiFi on Primary in order to communicate with the backend via it. Otherwise (no wifi) RPi's single Ethernet NIC is used for communication with secondaries and the backend, an internal network (10.0.3.0/8) is used for the first and an external network (DHCP-managed) is used for the second.

This PR has dependency on https://github.com/advancedtelematic/meta-updater-raspberrypi/pull/53

mike-sul:
Perhaps, we should run the oe-selftest on regular basis on some powerful server since it takes huge amount of time and CPU to run it on a local host (it still running on my machine) and bagger doesn't do work well recently as it seems becoming too 'crowded'.

patrickvacek:
> Perhaps, we should run the oe-selftest on regular basis on some powerful server since it takes huge amount of time and CPU to run it on a local host (it still running on my machine) and bagger doesn't do work well recently as it seems becoming too 'crowded'.

This is an ongoing topic of discussion. We are considering a new build server and we are also preparing to work on getting oe-selftest to actually run in CI soon. I'm surprised that bagger is too crowded, but that's why we have four build machines. :)

PR #532, opened 2019-06-25
Feat/ota 2666/garage tools mutual tls

patrickvacek:
Depends on ~~https://github.com/advancedtelematic/aktualizr/pull/1243~~ https://github.com/advancedtelematic/aktualizr/pull/1288.


patrickvacek:
Confirmed via manual testing that this fixes the bug with the `403 Forbidden` from garage-push.

PR #533, opened 2019-07-01
OTA-2488: Remove SOTA_SECONDARY_CONFIG_DIR configuration variable 

mike-sul:
Alignment with changes in this PR https://github.com/advancedtelematic/aktualizr/pull/1246

Signed-off-by: Mike Sul <ext-mykhaylo.sul@here.com>


mike-sul:
@patrickvacek @lbonn Let's decide what to do with it.

README.adoc
```diff
@@ -104,7 +104,7 @@ Your images will also need network connectivity to be able to reach an actual OT
...
 * `SOTA_CLIENT_FEATURES` - extensions to aktualizr. The only valid options are `hsm` (to build with HSM support) and `secondary-network` (to set up a simulated 'in-vehicle' network with support for a primary node with a DHCP server and a secondary node with a DHCP client).
-* `SOTA_SECONDARY_CONFIG_DIR` - a directory containing JSON configuration files for virtual secondaries on the host. These will be installed into `/etc/sota/ecus` on the device and automatically provided to aktualizr.
+* `SOTA_SECONDARY_CONFIG` - a file containing JSON configuration for secondaries. It will be installed into `/etc/sota/ecus` on the device and automatically provided to aktualizr. See link:https://github.com/advancedtelematic/aktualizr/blob/master/docs/posix-secondaries-bitbaking.adoc[here] for more details.
```
patrickvacek: Is the bit about /var/sota/ecus accurate? And has aktualizr been bumped already such that the new method is the only way forward now? If yes to both, then this is fine. But how does the config file get on the device? I've forgotten.

mike-sul: Yes, I am sure about `/etc/sota/ecus` - this is a folder that the secondaries configuration file is put in.
No, I haven't updated the reference to the desired version of aktualizr at meta-udpater, will do it.

>But how does the config file get on the device?

If `SOTA_SECONDARY_CONFIG` is specified in the local.conf then the corresponding file will be installed into /etc/sota/ecus  folder and /usr/lib/sota/conf.d/30-secondary-config.toml will reference to it.
If `SOTA_SECONDARY_CONFIG` is not specified then it will be auto-generated taking into account other configuration variables, see here for details on them  https://github.com/advancedtelematic/aktualizr/blob/master/docs/posix-secondaries-bitbaking.adoc

mike-sul: Bumped the corresponding version of aktualizr.

patrickvacek:
Looks good. Have you run oe-selftest? Since we have to bump aktualizr, it's more than just docs and we gotta play it safe. I can get a run started.

mike-sul:
> Looks good. Have you run oe-selftest? Since we have to bump aktualizr, it's more than just docs and we gotta play it safe. I can get a run started.

Yes, I have run them, but I suppose, I need to rerun them against the newly bumped version of aktalizr. So, please, run them, I will run some subset of them.

PR #534, opened 2019-07-02
Ci/gitlab core image minimal

lbonn:
Looks like we have what we need right now.

It also works with other branches than master: https://github.com/advancedtelematic/meta-updater/pull/535, but we'll have to synchronize pipeline definitions and scripts there.


lbonn:
Closed in favor of #537

PR #535, opened 2019-07-03
CI on rocko

lbonn:


lbonn:
I've used another way: include pipelines and test scripts from the master branch so that we don't have to back-port every ci enhancements to all release branches.

lbonn:
Passes oe-selftest ;). CI has been re-triggered by mistake, looking into that...

lbonn:
> Cool! Officially, this should get QA'd, although it actually doesn't affect anything except tests and CI.

There are changes to `envsetup.sh` and the qemu run scripts, so maybe it's still worth checking that an image can be built and run.

Raigi:
> > Cool! Officially, this should get QA'd, although it actually doesn't affect anything except tests and CI.
> 
> There are changes to `envsetup.sh` and the qemu run scripts, so maybe it's still worth checking that an image can be built and run.
> 
> No need to run oe-selftest obviously, though.

Ran bitbake for qemu, which was successful + device was created and provisioned successfully on server.

PR #536, opened 2019-07-04
Use libp11 from meta-oe upstream

ricardosalveti:
Remove local recipe (older than the one in meta-oe) and fix the missing opensc dependency (pkcs11-tool) by making it part of the softhsm-testoken rdepends.

libp11 itself shouldn't rdepend on opensc as it can be used without it.


patrickvacek:
Nice, this looks good. I'm testing it now.

PR #537, opened 2019-07-05
Ci/oe selftest

lbonn:
Based on #534.

### DIFF
--- a/lib/oeqa/selftest/cases/testutils.py
+++ b/lib/oeqa/selftest/cases/testutils.py
@@ -69,6 +69,16 @@ def qemu_send_command(port, command, timeout=60):
     return stdout, stderr, s2.returncode
 
 
+def metadir():
+    # Assume the directory layout for finding other layers. We could also
+    # make assumptions by using 'show-layers', but either way, if the
+    # layers we need aren't where we expect them, we are out of luck.
+    path = os.path.abspath(os.path.dirname(__file__))
+    metadir = path + "/../../../../../"
+
+    return metadir
+
+
 def akt_native_run(testInst, cmd, **kwargs):
     # run a command supplied by aktualizr-native and checks that:
     # - the executable exists
--- a/lib/oeqa/selftest/cases/testutils.py
+++ b/lib/oeqa/selftest/cases/testutils.py
@@ -1,4 +1,5 @@
 import os
+import oe.path
 import logging
 import re
 import subprocess
@@ -72,20 +73,16 @@ def akt_native_run(testInst, cmd, **kwargs):
     # run a command supplied by aktualizr-native and checks that:
     # - the executable exists
     # - the command runs without error
-    # NOTE: the base test class must have built aktualizr-native (in
-    # setUpClass, for example)
-    bb_vars = get_bb_vars(['SYSROOT_DESTDIR', 'base_prefix', 'libdir', 'bindir'],
-                          'aktualizr-native')
-    sysroot = bb_vars['SYSROOT_DESTDIR'] + bb_vars['base_prefix']
-    sysrootbin = bb_vars['SYSROOT_DESTDIR'] + bb_vars['bindir']
-    libdir = bb_vars['libdir']
+    #
+    # Requirements in base test class (setUpClass for example):
+    #   bitbake aktualizr-native
+    #   bitbake build-sysroots -c build_native_sysroot
+    #
+    # (technique found in poky/meta/lib/oeqa/selftest/cases/package.py)
+    bb_vars = get_bb_vars(['STAGING_DIR', 'BUILD_ARCH'])
+    sysroot = oe.path.join(bb_vars['STAGING_DIR'], bb_vars['BUILD_ARCH'])
 
-    program, *_ = cmd.split(' ')
-    p = '{}/{}'.format(sysrootbin, program)
-    testInst.assertTrue(os.path.isfile(p), msg="No {} found ({})".format(program, p))
-    env = dict(os.environ)
-    env['LD_LIBRARY_PATH'] = libdir
-    result = runCmd(cmd, env=env, native_sysroot=sysroot, ignore_status=True, **kwargs)
+    result = runCmd(cmd, native_sysroot=sysroot, ignore_status=True, **kwargs)
     testInst.assertEqual(result.status, 0, "Status not equal to 0. output: %s" % result.output)
 
 
--- a/lib/oeqa/selftest/cases/testutils.py
+++ b/lib/oeqa/selftest/cases/testutils.py
@@ -34,6 +34,7 @@ def qemu_boot_image(imagename, **kwargs):
     args.dir = 'tmp/deploy/images'
     args.efi = kwargs.get('efi', False)
     args.machine = kwargs.get('machine', None)
+    args.mem = kwargs.get('mem', '128M')
     qemu_use_kvm = get_bb_var("QEMU_USE_KVM")
     if qemu_use_kvm and \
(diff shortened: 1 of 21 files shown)


lbonn:
Ok it's been quite chaotic, I'll squash all of that at some point.
But it seems that it's getting close to the end.

lib/oeqa/selftest/cases/testutils.py
```diff
@@ -60,32 +61,38 @@ def qemu_bake_image(imagename):
...
-    sysroot = bb_vars['SYSROOT_DESTDIR'] + bb_vars['base_prefix']
-    sysrootbin = bb_vars['SYSROOT_DESTDIR'] + bb_vars['bindir']
-    libdir = bb_vars['libdir']
-
-    program, *_ = cmd.split(' ')
-    p = '{}/{}'.format(sysrootbin, program)
-    testInst.assertTrue(os.path.isfile(p), msg="No {} found ({})".format(program, p))
-    env = dict(os.environ)
-    env['LD_LIBRARY_PATH'] = libdir
-    result = runCmd(cmd, env=env, native_sysroot=sysroot, ignore_status=True, **kwargs)
```
patrickvacek: Are you sure it's safe to remove the LD_LIBRARY_PATH bit? I seem to recall that was necessary for some obscure reason. If only I'd left a comment...

lib/oeqa/selftest/cases/updater_qemux86_64.py
```diff
@@ -39,20 +41,18 @@ def setUpClass(cls):
...
+        bitbake('aktualizr-native aktualizr-device-prov')
+        bitbake('build-sysroots -c build_native_sysroot')
 
     def test_cert_provider_help(self):
         akt_native_run(self, 'aktualizr-cert-provider --help')
 
     def test_cert_provider_local_output(self):
-        logger = logging.getLogger("selftest")
-        logger.info('Running bitbake to build aktualizr-device-prov')
-        bitbake('aktualizr-device-prov')
```
patrickvacek: So are you expecting bitbake to have been run first? Probably okay, but causes problems if you try to run this test in isolation or in an unexpected sequence.

scripts/ci/Dockerfile.bitbake
```diff
@@ -1,42 +1,55 @@
...
      texinfo \
      unzip \
      wget \
      xterm \
      xz-utils
 
-ARG uid=1000
-ARG gid=1000
+ARG uid=4321
+ARG gid=4321
```
patrickvacek: What's the deal with this special user?

lib/oeqa/selftest/cases/testutils.py
```diff
@@ -60,32 +61,38 @@ def qemu_bake_image(imagename):
...
 
 
+def metadir():
+    # Assume the directory layout for finding other layers. We could also
+    # make assumptions by using 'show-layers', but either way, if the
+    # layers we need aren't where we expect them, we are out of luck.
+    path = os.path.abspath(os.path.dirname(__file__))
+    metadir = path + "/../../../../../"
+
+    return metadir
```
patrickvacek: Good call to pull this out.

scripts/ci/Dockerfile.bitbake
lbonn: That was the uid/gid of Jenkins and the permissions of the bitbake cache matches that.

lib/oeqa/selftest/cases/updater_qemux86_64.py
lbonn: I moved it to `setUpClass()`, so I guess it should always happen, no?

lib/oeqa/selftest/cases/testutils.py
lbonn: I actually had problem with .so before this change. It looks like we had to do that because the library path was actually outside of the sysroot. With `run_cmd(..., native_sysroot=...)` and the bitbake trick from there https://git.yoctoproject.org/cgit/cgit.cgi/poky/tree/meta/lib/oeqa/selftest/cases/package.py?id=2a6094ca8c4aa5c00636df77669e5bca6ad52bcf#n29, it seems that it works with less confusion.

Might need a double check on rocko and others though.

lib/oeqa/selftest/cases/updater_qemux86_64.py
patrickvacek: Oops, yes of course. Sorry, using github on a phone is difficult.

lib/oeqa/selftest/cases/testutils.py
patrickvacek: Cool, no objection if it works!

lbonn:
Did some cleanup:

- added comments about oe-selftests skipping ostreepush
- added comments on the checkout step
- added comments on the kvm trick
- squashed some similar commits

lbonn:
Ok, I've made some more changes. It should have enough features to be used on other yocto layers + aktualizr and pin some versions just in case.

lib/oeqa/selftest/cases/updater_qemux86_64.py
```diff
@@ -263,18 +263,17 @@ def test_provisioning(self):
...
         # Verify that HSM is able to initialize.
-        ran_ok = False
         for delay in [5, 5, 5, 5, 10]:
             sleep(delay)
             p11_out, p11_err, p11_ret = self.qemu_command(pkcs11_command)
             hsm_out, hsm_err, hsm_ret = self.qemu_command(softhsm2_command)
-            if p11_ret == 0 and hsm_ret == 0 and hsm_err == b'':
-                ran_ok = True
+            if (p11_ret == 0 and hsm_ret == 0 and hsm_err == b'' and
+                    b'X.509 cert' in p11_out and b'present token' in p11_err):
```
patrickvacek: Good call.

scripts/ci/gitlab/docker.yml
```diff
@@ -20,6 +22,8 @@ Docker Setup (local):
     - docker push "$CHECKOUT_IMAGE"
 
 Docker Setup (simple):
+  # intended to be un on other branches and repos: just pulls the last master
```
patrickvacek: "un on"?

lib/oeqa/selftest/cases/updater_qemux86_64.py
lbonn: Yes, I think it was the cause of some random failures. Haven't seen any since then.

scripts/ci/gitlab/docker.yml
lbonn: thanks!

lbonn:
Ok the scripts have been changed significantly again and use [the extends feature from gitlab](https://docs.gitlab.com/ee/ci/yaml/#extends) instead of less-powerful yaml anchors.

I'll change the [rocko example PR](https://github.com/advancedtelematic/meta-updater/pull/535) to reflect these changes and start the integration with aktualizr for ptest.

patrickvacek:
Looks cool, but I only see the qemux86-64 tests running. Should the others run too?

lbonn:
I've only enabled the qemu tests for now. As we have two build machines, minnowboard and raspberrypi would lengthen the tests some more (maybe double the time) and they tend to break even less often than the rest.

Maybe as nightly jobs?

patrickvacek:
> Maybe as nightly jobs?

That's fine for me, we can punt on that for the moment. But I didn't see the native tests, either.

lbonn:
Ah my bad sorry, that's indeed a mistake. Thanks!

lbonn:
ptest is far from being green but I think this is finally ready now

PR #539, opened 2019-07-16
Allow customization of data in targets.json

doanac:
This allows a user to define their own script to be executed after
a target has been added to the repository but not yet signed.

Signed-off-by: Andy Doan <andy@foundries.io>


doanac:
@patrickvacek - its to enable us to add docker-apps entries into the custom data for a target before its pushed to the server. I thought a little about trying to do that in garage-sign itself, but I think this is pretty workflow specific.

PR #540, opened 2019-07-17
Ci/ptest fixes

lbonn:
Fixes for ptest on master.

One successful run here: https://main.gitlab.in.here.com/olp/edge/ota/connect/client/meta-updater/-/jobs/206534


lib/oeqa/selftest/cases/updater_qemux86_64_ptest.py
```diff
@@ -21,7 +21,7 @@ def setUpLocal(self):
         self.append_config('PTEST_ENABLED_pn-aktualizr = "1"')
         self.append_config('IMAGE_INSTALL_append += "aktualizr-ptest ptest-runner "')
         self.append_config('IMAGE_FSTYPES_remove = "ostreepush garagesign garagecheck"')
-        self.qemu, self.s = qemu_launch(machine='qemux86-64', mem="256M")
+        self.qemu, self.s = qemu_launch(machine='qemux86-64', mem="768M")
```
patrickvacek: Seems like an oddly specific number. What's the deal?

lbonn: 256M seems to not be enough, some tests are killed and don't pass.

I tried to use something that's a bit less than a full GB since we've witnessed problems on CI where qemu had trouble allocating that much adjacent memory. There might be another solution for that but in the meantime, I'm trying to keep the foot print as low as possible.

lbonn:
Ah I forgot that the aktualizr part needs to be merged first

lbonn:
Ok, now uses aktualizr from master

PR #545, opened 2019-07-18
update oe-selftests to use wait-until-provisioned feature

Zee314159:
Update oe-selftests to use wait-until-provisioned feature

### DIFF
--- a/lib/oeqa/selftest/cases/updater_qemux86_64.py
+++ b/lib/oeqa/selftest/cases/updater_qemux86_64.py
@@ -150,7 +150,7 @@ class ManualControlTests(OESelftestTestCase):
 
         stdout, stderr, retcode = self.qemu_command('aktualizr once')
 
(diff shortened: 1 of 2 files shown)


lib/oeqa/selftest/cases/updater_qemux86_64.py
```diff
@@ -150,7 +150,7 @@ def test_manual_run_mode_once(self):
 
         stdout, stderr, retcode = self.qemu_command('aktualizr once')
 
-        stdout, stderr, retcode = self.qemu_command('aktualizr-info')
+        stdout, stderr, retcode = self.qemu_command('aktualizr-info --wait-until-provisioned')
```
patrickvacek: This particular function actually shouldn't have that, since the calls are made sequentially here. If `aktualizr once` didn't succeed in provisioning the device, something went wrong and there's nothing more to wait for.

lbonn:
Closed for now for #553

PR #547, opened 2019-07-18
Backport/rocko/secondaries and ci improvement

mike-sul:
oe-selftest hasn't been executed yet


lbonn:
I suspect that the oe-selftest failure on qemu has something to do with the libp11 and softhsm changes.

The raspberrypi thing looks like a backport problem as the boot process is quite different on rocko and master (ongoing problem that we're having).

patrickvacek:
That last commit (0978ca9) seems like a misnomer; I don't see openssl-bin getting removed, and I'm not sure that's a good idea anyway.

lbonn:
Yes, maybe it would just be wiser to keep our custom recipe on old branches. I don't what the state of it is in rocko.

mike-sul:
openssl-bin has been actually removed. I removed it at first, then added, then removed the whole commit which includes 'opensc openssl-bin' and then added a commit that adds opensc as it's actually is needed while openssl-bin is not (tests passed on bagger).

patrickvacek:
Oh that's right, I think `openssl-bin` wasn't a thing on rocko. That's one of the lovely things that changed at some point.

lbonn:
By using upstream's p11 recipe on rocko, we're moving from 0.4.9 (current version or our custom recipe) to 0.4.7. I'm not sure that's what we want in the end...

mike-sul:
The tests have passed.
Do you think it makes sense to do manual verification for RPi, build RPi image and see if it works.

patrickvacek:
>  Do you think it makes sense to do manual verification for RPi, build RPi image and see if it works.

That's not a bad idea. Typically, QA will do that as part of their job for PRs to rocko, but it never hurts to do it yourself first, especially since there are several relevant changes. We normally ask QA to take a look after another developer has reviewed and approved.

patrickvacek:
This looks good but I think I agree with Laurent about keeping our version of the p11 recipe in rocko, since it is newer than the meta-oe version. (This may be true for sumo/thud as well, I haven't checked those.)

mike-sul:
> This looks good but I think I agree with Laurent about keeping our version of the p11 recipe in rocko, since it is newer than the meta-oe version. (This may be true for sumo/thud as well, I haven't checked those.)

Please, see if it's what is required (I removed corresponding commit during rebasing).

lbonn:
Don't know why it's not a problem here, but I would suggest also cherry-picking this #556 as it's necessary to fix the test on sumo and thud.

lbonn:
Other than that, everything looks good except the 'Remove explicit dependency' which is not needed since we're using our libp11 recipe again.

mike-sul:
@lbonn @patrickvacek Please, verify if it has the desired set of commits/cherrypicks.

lbonn:
Ok it looks good, it's leaving out rpi64 but we can rework on that later.

mike-sul:
> Ok it looks good, it's leaving out rpi64 but we can rework on that later.

yes, it didn't work out of box on rocko.

patrickvacek:
BTW @Raigi this should be tested in tandem with https://github.com/advancedtelematic/meta-updater-raspberrypi/pull/54.

Raigi:
> BTW @Raigi this should be tested in tandem with [advancedtelematic/meta-updater-raspberrypi#54](https://github.com/advancedtelematic/meta-updater-raspberrypi/pull/54).

Tests:
1. create qemu devices with ip secondaries OK
2. create shared credentials qemu devices, install updates OK
3. create device credentials qemu devices, install updates OK
4. create device cred raspberry device, install updates OK
5. create device with ip secondaries specifying the SOTA_SECONDARY_CONFIG OK
6. update a device from aktualizr fce5854ff1 version to aktualizr 9c592cf9d8 version OK

PR #551, opened 2019-07-18
Fix PIN_LIST in checkout-oe.sh

lbonn:
It didn't work if the manifest already had a revision specified


lbonn:
Thanks!

For a quick explanation, `-d` deletes an attribute and `-i` inserts one. There is also `-u` which updates one but I think it doesn't work when the attribute is not present before.

Hence, the most robust way is to delete it (wont' do anything if not present) and insert it again.

What happened before is that `revision` was specified twice and it made repo trip.

PR #542, opened 2019-07-17
Ci on thud

lbonn:

PR #541, opened 2019-07-17
 CI on sumo

lbonn:

PR #556, opened 2019-07-19
Give more memory for minnowboard selftest

lbonn:
Problem was spotted on sumo and thud #541 and #542 but it should also appear here, in all likelihood. I'll launch a pipeline with minnowboard here soon.


lbonn:
Hm looks like the build failed. I'll merge anyway because this commit helped on other branches and we've already merged there. The build issue will be considered later.

PR #558, opened 2019-07-22
TI SDK 05.03 support

eu-smirnov:


patrickvacek:
Looks fine, thanks. Does it makes sense to put this in master as well?

lbonn:
I think the TI support is based on rocko (and thud for the recent one?). So, probably the necessary layers won't work with master.

@eu-smirnov can you confirm?

eu-smirnov:
Yes, this was tested on rocko only, so I wouldn't put it in master for now.

PR #560, opened 2019-07-23
Output site.conf during CI configure step

lbonn:
Was deemed useful there: https://github.com/advancedtelematic/aktualizr/pull/1262


lbonn:
Now contains two other unrelated things:

- fixing gitlab
- removing references to Jenkins (uninstalled today)

PR #565, opened 2019-07-24
sota: add support for freedom-u540

ricardosalveti:
fitImage used by default.

To be used together with https://github.com/ricardosalveti/meta-updater-riscv.


patrickvacek:
When I run `source meta-updater/scripts/envsetup.sh freedom-u540`, I get `cat: <base-dir>/meta-updater/scripts/../../meta-updater/conf/include/bblayers/sota_freedom-u540.inc: No such file or directory`. Not critical, and I know that in our case that is tightly coupled with how the [updater-repo](https://github.com/advancedtelematic/updater-repo/) sets up the environment, but worth considering.

ricardosalveti:
Yeah, sorry, didn't get this one as I'm setting it up differently, will take a look.

ricardosalveti:
Added the bblayers include, but updater-repo also requires changes, will follow that up separately.

classes/sota_freedom-u540.bbclass
```diff
@@ -0,0 +1,18 @@
...
+KERNEL_IMAGETYPE_sota = "fitImage"
+KERNEL_IMAGETYPES_remove_sota = "uImage"
+OSTREE_KERNEL = "${KERNEL_IMAGETYPE}-${INITRAMFS_IMAGE}-${MACHINE}-${MACHINE}"
+
+OSTREE_BOOTLOADER ?= "u-boot"
+INITRAMFS_FSTYPES = "cpio.gz"
+PREFERRED_PROVIDER_virtual/bootloader_sota ?= "u-boot"
+
+IMAGE_BOOT_FILES_sota ?= "fw_payload.bin boot.scr uEnv.txt"
+KERNEL_DEVICETREE_sota ?= "sifive/${RISCV_SBI_FDT}"
```
alistair23: Can we add `IMAGE_FSTYPES_remove = "wic.gz"` in here so we don't need it in the meta-updater-riscv layer?

classes/sota_freedom-u540.bbclass
```diff
@@ -0,0 +1,18 @@
...
+
+UBOOT_DTB_LOADADDRESS = "0x82200000"
+
+OSTREE_KERNEL_ARGS_sota ?= "earlycon=sbi console=ttySIF0 ramdisk_size=16384 root=/dev/ram0 rw rootfstype=ext4 rootwait rootdelay=2 ostree_root=/dev/mmcblk0p3"
+
+WKS_FILE_sota = "freedom-u540-opensbi-sota.wks"
```
alistair23: Can we add this wic file to meta-riscv as well?

classes/sota_freedom-u540.bbclass
```diff
@@ -0,0 +1,18 @@
...
+OSTREE_KERNEL = "${KERNEL_IMAGETYPE}-${INITRAMFS_IMAGE}-${MACHINE}-${MACHINE}"
+
+OSTREE_BOOTLOADER ?= "u-boot"
+INITRAMFS_FSTYPES = "cpio.gz"
+PREFERRED_PROVIDER_virtual/bootloader_sota ?= "u-boot"
+
+IMAGE_BOOT_FILES_sota ?= "fw_payload.bin boot.scr uEnv.txt"
+KERNEL_DEVICETREE_sota ?= "sifive/${RISCV_SBI_FDT}"
+
+UBOOT_DTB_LOADADDRESS = "0x82200000"
```
alistair23: This line shouldn't be required now either

classes/sota_freedom-u540.bbclass
alistair23: Argh, that doesn't work if you are using a fitImage

ricardosalveti: So, there is actually no need to remove wic.gz, and I tested with it as well (flashed the wic.gz image into my sdcard). Which error are you getting with it?

classes/sota_freedom-u540.bbclass
ricardosalveti: Sure, let me propose it there.

classes/sota_freedom-u540.bbclass
ricardosalveti: Yup, will refresh, thanks!

classes/sota_freedom-u540.bbclass
alistair23: I see this error:

> ERROR: Task do_image_wic in /scratch/alistair/yocto/oe-master/meta/recipes-core/images/core-image-minimal.bb depends upon non-existent task IMAGE_CMD_ota-ext4 in /scratch/alistair/yocto/oe-master/meta/recipes-core/images/core-image-minimal.bb
> ERROR: Command execution failed: 1

I suspect you don't see if as you have the meta-updater-riscv layer which contains this line:
`IMAGE_FSTYPES_remove = "wic wic.gz wic.bmap ext4 ext4.gz"`
in
`recipes-core/images/initramfs-ostree-image.bbappend`.

I am currently not using they layer and was hoping to contain everything in the meta-riscv and meta-updater layer.

ricardosalveti: Right, that is another pending fix that I will propose to meta-updater directly, but the issue is that the initramfs recipe is being contaminated by IMAGE_FSTYPES (which is also populated by the board config).

alistair23: Yep, I think your bbappend can just be put in the main recipe. That should fix the issue.

ricardosalveti: There is no need for initramfs wic/ext4 images, which is why the error.

classes/sota_freedom-u540.bbclass
```diff
@@ -0,0 +1,16 @@
...
+KERNEL_CLASSES_sota = " kernel-fitimage "
+KERNEL_IMAGETYPE_sota = "fitImage"
+KERNEL_IMAGETYPES_remove_sota = "uImage"
+OSTREE_KERNEL = "${KERNEL_IMAGETYPE}-${INITRAMFS_IMAGE}-${MACHINE}-${MACHINE}"
+
+OSTREE_BOOTLOADER ?= "u-boot"
+INITRAMFS_FSTYPES = "cpio.gz"
+PREFERRED_PROVIDER_virtual/bootloader_sota ?= "u-boot"
+
+IMAGE_BOOT_FILES_sota ?= "fw_payload.bin boot.scr uEnv.txt"
```
alistair23: Could we change this to:
```
IMAGE_BOOT_FILES_append = " uEnv.txt ${OSTREE_KERNEL}"
```
That way we don't change what is already there.

ricardosalveti: The kernel is available as part of the rootfs and not from the boot partition, which is why we have boot.scr and uEnv.txt, so that would only work if not really using ostree.

ricardosalveti: So in this case the only pieces required at the boot partition is fw_payload.bin boot.scr uEnv.txt (in really only fw_payload.bin, but you would need to change the default u-boot load script).

alistair23: Ok. can it still be an append?

ricardosalveti: Append will cause the kernel images to be available in the boot partition, and they are not used/required, so unsure why.

alistair23: My thinking was to keep the /boot partition generic, but you are right. This is fine

patrickvacek:
My build from last night succeeded, but I saw one warning:
```
WARNING: opensbi-0.4-r0 do_package: Manifest /home/patrick/Code/updater_repo_master/build-freedom/tmp/sstate-control/manifest-x86_64_x86_64-nativesdk-initramfs-ostree-image.packagedata not found in freedom_u540 riscv64 allarch x86_64_x86_64-nativesdk (variant '')?
```

Is that anything to be concerned about from your perspective?

In any case, this looks cool and I think we can safely merge this soon. Thanks!

PR #566, opened 2019-07-24
initramfs-ostree-image: unset EXTRA_IMAGEDEPENDS to avoid extra dependencies

ricardosalveti:
It is common for machine configurations to set EXTRA_IMAGEDEPENDS, which
can end up causing circular dependencies when building the initramfs
image (they are usually only required for the actual rootfs image).


ricardosalveti:
@alistair23 this change + https://github.com/riscv/meta-riscv/pull/163 should remove the need for https://github.com/ricardosalveti/meta-updater-riscv/blob/master/recipes-core/images/initramfs-ostree-image.bbappend

PR #568, opened 2019-07-26
feat/OTA-3155/Dont-patch-credentials

Zee314159:
Draft code.

### DIFF
--- a/recipes-sota/aktualizr/aktualizr-shared-prov-creds.bb
+++ b/recipes-sota/aktualizr/aktualizr-shared-prov-creds.bb
@@ -14,14 +14,17 @@ require credentials.inc
 do_install() {
     if [ -n "${SOTA_PACKED_CREDENTIALS}" ]; then
         install -m 0700 -d ${D}${localstatedir}/sota
-        cp "${SOTA_PACKED_CREDENTIALS}" ${D}${localstatedir}/sota/sota_provisioning_credentials.zip
-        # Device should not be able to push data to treehub
-        zip -d ${D}${localstatedir}/sota/sota_provisioning_credentials.zip treehub.json
-        # Device has no use for the API Gateway. Remove if present. See:
-        # https://github.com/advancedtelematic/ota-plus-server/pull/1913/
-        if unzip -l ${D}${localstatedir}/sota/sota_provisioning_credentials.zip api_gateway.url > /dev/null; then
-            zip -d ${D}${localstatedir}/sota/sota_provisioning_credentials.zip api_gateway.url 
-        fi
+        # root.json contains the root metadata for bootstrapping the Uptane metadata verification process.
+        # autoprov.url has the URL to the device gateway on the server, which is where we send most of our requests.
+        # autoprov_credentials.p12 contains the shared provisioning credentials.
+        for var in root.json autoprov.url autoprov_credentials.p12; do
+            if unzip -l "${SOTA_PACKED_CREDENTIALS}" $var > /dev/null; then
+                unzip "${SOTA_PACKED_CREDENTIALS}" $var -d ${T}
+                zip -mj -q ${D}${localstatedir}/sota/sota_provisioning_credentials.zip ${T}/$var
+            else
+                bbwarn "$var is missing from credentials.zip"
+            fi
+        done
     fi
 }
 


patrickvacek:
Oh! As to the whys:
1. `autoprov_credentials.p12` contains the shared provisioning credentials.
2. `autoprov.url` has the URL to the device gateway on the server, which is where we send most of our requests.
3. `root.json` contains the [root metadata](https://uptane.github.io/uptane-standard/uptane-standard.html#rfc.section.5.1.1) for bootstrapping the Uptane metadata verification process.

Also note that the code mentions a "server_ca.pem". That is only used in device credential provisioning for the [OTA Community Edition](https://github.com/advancedtelematic/ota-community-edition/blob/2341d590eda6cdefe7beeca5ab22489567ce5520/scripts/start.sh#L156) and this recipe is only for shared credential provisioning, so you can ignore that file here.

Zee314159:
@patrickvacek Thank you!

recipes-sota/aktualizr/aktualizr-shared-prov-creds.bb
```diff
@@ -9,18 +9,48 @@ inherit allarch
 DEPENDS = "aktualizr-native zip-native"
 ALLOW_EMPTY_${PN} = "1"
 
+T = "${WORKDIR}/tmp"
```
lbonn: From the manual (section linked by Patrick), it sounds like this is already defined and you do not need to redefine it yourself. Does it work without?

recipes-sota/aktualizr/aktualizr-shared-prov-creds.bb
```diff
@@ -9,18 +9,48 @@ inherit allarch
...
     if [ -n "${SOTA_PACKED_CREDENTIALS}" ]; then
         install -m 0700 -d ${D}${localstatedir}/sota
-        cp "${SOTA_PACKED_CREDENTIALS}" ${D}${localstatedir}/sota/sota_provisioning_credentials.zip
-        # Device should not be able to push data to treehub
-        zip -d ${D}${localstatedir}/sota/sota_provisioning_credentials.zip treehub.json
-        # Device has no use for the API Gateway. Remove if present. See:
-        # https://github.com/advancedtelematic/ota-plus-server/pull/1913/
-        if unzip -l ${D}${localstatedir}/sota/sota_provisioning_credentials.zip api_gateway.url > /dev/null; then
-            zip -d ${D}${localstatedir}/sota/sota_provisioning_credentials.zip api_gateway.url 
+        # root.json contains the root metadata for bootstrapping the Uptane metadata verification process.
```
lbonn: Since it's not using a loop as some missing elements cause warnings and others not, we might as well group the two different policies together: first the ones with warning, then the others.

patrickvacek: Actually, some of these shouldn't be re-zipped at all. The various targets files and tufrepo.url should be left out here; they aren't used by libaktualizr. Then the three remaining items (root.json, autoprov.url, and the p12) should each have warnings, so maybe it is possible to get fancy with a loop.

Zee314159: > The various targets files and tufrepo.url should be left out here

Do you mean this three files should not be zipped?

> remaining items (root.json, autoprov.url, and the p12) should each have warnings, so maybe it is possible to get fancy with a loop.

With an infinite loop?

recipes-sota/aktualizr/aktualizr-shared-prov-creds.bb
Zee314159: Yes, it works without.

recipes-sota/aktualizr/aktualizr-shared-prov-creds.bb
patrickvacek: > Do you mean this three files should not be zipped?

Yes, libaktualizr has no use for them.

> With an infinite loop?

No, you can make the unzip and rezip logic generic by looping over the list of strings "root.json", "autoprov.url", and "autoprov_credentials.p12". I think that's what @lbonn was referring to when he mentioned a loop; please correct me if I misunderstood!

lbonn: yes, that's what I meant

Zee314159: Oh, I see !

PR #570, opened 2019-07-26
freedom-u540: Remove the fitImage overrides

alistair23:
meta-riscv now generates a fitImage by default for the freedom-u540
machine so remove the unneeded overrides.

Signed-off-by: Alistair Francis <alistair.francis@wdc.com>

### DIFF
--- a/classes/sota_freedom-u540.bbclass
+++ b/classes/sota_freedom-u540.bbclass
@@ -1,15 +1,11 @@
-# Support fitImage by default
-KERNEL_CLASSES_sota = " kernel-fitimage "
-KERNEL_IMAGETYPE_sota = "fitImage"
-KERNEL_IMAGETYPES_remove_sota = "uImage"
 OSTREE_KERNEL = "${KERNEL_IMAGETYPE}-${INITRAMFS_IMAGE}-${MACHINE}-${MACHINE}"
 
 OSTREE_BOOTLOADER ?= "u-boot"
 INITRAMFS_FSTYPES = "cpio.gz"
 PREFERRED_PROVIDER_virtual/bootloader_sota ?= "u-boot"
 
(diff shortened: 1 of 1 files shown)


alistair23:
@ricardosalveti

classes/sota_freedom-u540.bbclass
```diff
@@ -1,15 +1,10 @@
...
-KERNEL_IMAGETYPES_remove_sota = "uImage"
 OSTREE_KERNEL = "${KERNEL_IMAGETYPE}-${INITRAMFS_IMAGE}-${MACHINE}-${MACHINE}"
 
 OSTREE_BOOTLOADER ?= "u-boot"
 INITRAMFS_FSTYPES = "cpio.gz"
 PREFERRED_PROVIDER_virtual/bootloader_sota ?= "u-boot"
 
-IMAGE_BOOT_FILES_sota ?= "fw_payload.bin boot.scr uEnv.txt"
-KERNEL_DEVICETREE_sota ?= "sifive/${RISCV_SBI_FDT}"
+IMAGE_BOOT_FILES += "uEnv.txt"
```
ricardosalveti: Looking at https://github.com/riscv/meta-riscv/blob/cb0857efe5fb3550c036ee05378e5c4098099914/conf/machine/freedom-u540.conf#L56 we would have the kernel installed at the boot partition, which is not needed in this case (and which might be quite big depending on the initramfs used).

Maybe just keep the original IMAGE_BOOT_FILES_sota instead?

alistair23: I would like to try and use the original IMAGE_BOOT_FILES variable to avoid making changes here when the boot files change (as they will again when we add U-Boot SPL support).

How does my new patch look? I am removing the image from the files

patrickvacek:
Haven't tested but looks fine.

PR #574, opened 2019-08-01
Connman fix master

lbonn:
Fix for OTA-3401.

In summary, moving from using connman to systemd-networkd broke dns.

This fix allows:

- move from current version with connman to new versions with systemd-networkd (even without hack). This could maybe be upstreamed in some form
- move from old versions with connman to this one, with a systemd hack at boot (fix the broken symlink)


lbonn:
Bumped aktualizr to try to fix these oe-selftest failures + added a small helpful change in the qemu run script.

PR #576, opened 2019-08-01
Connman fix thud

lbonn:
See #574


lbonn:
Was confirmed by Raigi, also on thud.

PR #578, opened 2019-08-06
qemucommand.py: redirect qemu monitor to null chardev

eu-smirnov:
By default, qemu monitor gets attached to stdio, and that interferes with the test output.
We also don't need vga options in nographic mode.

Signed-off-by: Eugene Smirnov <evgenii.smirnov@here.com>

### DIFF
--- a/scripts/qemucommand.py
+++ b/scripts/qemucommand.py
@@ -102,12 +102,8 @@ class QemuCommand(object):
         cmdline += [
             "-serial", "tcp:127.0.0.1:%d,server,nowait" % self.serial_port,
             "-m", self.mem,
-            "-usb",
             "-object", "rng-random,id=rng0,filename=/dev/urandom",
             "-device", "virtio-rng-pci,rng=rng0",
-            "-device", "usb-tablet",
-            "-show-cursor",
-            "-vga", "std",
             "-net", netuser,
             "-net", "nic,macaddr=%s" % self.mac_address
         ]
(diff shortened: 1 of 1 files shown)


scripts/qemucommand.py
```diff
@@ -102,12 +102,8 @@ def command_line(self):
         cmdline += [
             "-serial", "tcp:127.0.0.1:%d,server,nowait" % self.serial_port,
```
lbonn: While we're at it, maybe we should remove this serial option by default in --no-gui. When not present, the system is directly usable with stdio and I don't remember having to use the serial port to debug something.
It could be made optional just in case though.

eu-smirnov: I only didn't move it because I was afraid that it might break some tests and I also was lazy to check that. I guess I'll move it and see what happens.

eu-smirnov: @lbonn Didn't realize that in this case kernel boot log gets into the test log, which is probably not something that we want.

lbonn: What do you mean by test logs?

Currently, there is no output at all with `--no-gui`, so I thought it would be useful to at least be able to get a prompt there. kernel logs can be annoying sure, but they are also useful when debugging u-boot / boot issues.

It could also be behind a switch though and we'd also better make sure that the message "To connect to the serial console:" only shows when it would actually work.

eu-smirnov: @lbonn sorry, I mean oe-selftest logs

lbonn: ah ok, yes that makes sense. Anyway looks like it requires more care than I thought, so I guess we can already merge this PR and come back on that if needed.

PR #579, opened 2019-08-14
Backport/rocko 19 08 14

lbonn:


lib/oeqa/selftest/cases/updater_qemux86_64.py
```diff
@@ -358,7 +360,7 @@ def configure(self):
             self._test_ctx.append_config('SOTA_CLIENT_PROV = " aktualizr-shared-prov "')
 
         def is_ecu_registered(self, ecu_id):
-            max_number_of_tries = 40
+            max_number_of_tries = 120
```
mike-sul: I think, I can replace the current implementation that iterates over check & wait with `aktualizr-info --wait-until-provisioned`

patrickvacek: Sounds like a good plan, but obviously not a high priority. :)

lbonn:
The OSTREE_BRANCHNAME backport went wrong, will instead take something like that: https://github.com/advancedtelematic/meta-updater/pull/585

lbonn:
Weird, minnowboard keeps failing. Also on sumo and thud.

patrickvacek:
> Weird, minnowboard keeps failing. Also on sumo and thud.

Can you reproduce locally? That test seems surprisingly flaky. I haven't determined why.

lbonn:
Seems like it's actually broken:

```
error: file
`/ostree/poky-ee5dda54cd602959a6b11de59df52e0f7d1de44b70436617c864101e1fdad072/
vmlinuz' not found.
error: you need to load the kernel first.

Press any key to continue...
```

lbonn:
Ok I think the last commit should fix it. Wasn't so easy to track down when everything depends on some variables being globally exported...

lbonn:
But then, I don't know how it works on master... Will investigate.

patrickvacek:
> Will investigate

In `src/boot/grub2/ostree-grub-generator` of ostree, I see the comment "Default to /boot if OSTREE_BOOT_PARTITION is not set and /boot is on the same device than ostree/repo". Perhaps the directory layout is different in master?

In any case, it won't hurt to have the same export in master, right?

And is this ready to go by your estimation? Looks fine to me and ready for QA.

lbonn:
Yes, you're right it also belongs to master probably. Ready to test indeed!

PR #581, opened 2019-08-14
Backport/thud 19 08 14

lbonn:


classes/image_types_ostree.bbclass
```diff
@@ -95,15 +97,40 @@ IMAGE_CMD_ostree () {
...
+        rm -rf usr/local
+    fi
+
+    if [ -n "${SYSTEMD_USED}" ]; then
+        echo "d /var/usrlocal 0755 root root -" >>${tmpfiles_conf}
+    else
+        echo "mkdir -p /var/usrlocal; chown 755 /var/usrlocal" >>${tmpfiles_conf}
+    fi
+
+    dirs="bin etc games include lib man sbin share src"
```
mike-sul: I like `games` directory here :)

patrickvacek: I was also amused by that, but that is based on an actual recommendation from OSTree!

Issue #583, opened 2019-08-15
Simultaneous bitbaking generates a wrong commit for the ref of the basename-image branch


psaavedra:
Related to: 202a8c70ba8cdcfb39cf8e2b19a08b5f8850a0f9 (#489):

```
image_types_ostree: Add a unique ref to fix simultaneous bitbaking.
    
    To enable simultaneous bitbaking of two images with the same branch name, create
    a new ref in the OSTree repo using the basename of the image.
```

**PROBLEM**

The resulting ref for the `${OSTREE_BRANCHNAME}-${IMAGE_BASENAME}` is malformed because refs points to a commit not related to the branch.

**HOW TO REPRODUCE**

Clone the OSTREE repro create by `bitbake`:

```
ostree --repo=${LOCAL_CLONE}/ostree_repo init --mode=archive
ostree --repo=${LOCAL_CLONE}/ostree_repo prod remote add sota  file:///${DEPLOY_DIR_IMAGE}/ostree_repo --no-gpg-verify
```
Check pull with the  `${OSTREE_BRANCHNAME}` ref is `OK`:

```
ostree   pull  sota:${OSTREE_BRANCHNAME}
80 metadata, 910 content objects imported                                                                                                                                                   
 # OK
```
Check pull with the  `${OSTREE_BRANCHNAME}-${IMAGE_BASENAME}` ref is `KO`:

```
ostree   pull  sota:${OSTREE_BRANCHNAME}-${IMAGE_BASENAME}
error: Commit 2f1205f045a0d37906897bf02947b8df42404306d01742da858b251f422c97d0: Commit has no requested ref ?${OSTREE_BRANCHNAME}-${IMAGE_BASENAME}? in ref binding metadata (?${OSTREE_BRANCHNAME}?)
```

patrickvacek:
Fixed by #614.

PR #584, opened 2019-08-15
Fix wrong commit in simultaneous bitbaking (fixes: #583)

psaavedra:
Using commit instead of an artificial created reference in the ref of the basename-image branch for the simultaneous bitbaking of two images

In other case, the commit referenciated for the basename-image ref has not the requested ref for the branch and  `ostree pull` throws an error.

Related to: 
* Commit: 202a8c70ba8cdcfb39cf8e2b19a08b5f8850a0f9 (#489)
* Issue: #583 

Signed-off-by: Pablo Saavedra <psaavedra@igalia.com>

Target: master, warrior (backport)

### DIFF
--- a/classes/image_types_ostree.bbclass
+++ b/classes/image_types_ostree.bbclass
@@ -145,16 +145,21 @@ IMAGE_CMD_ostreecommit () {
            --subject="${OSTREE_COMMIT_SUBJECT}" \
            --body="${OSTREE_COMMIT_BODY}"
 
+    # Commit the result in the image basename branch:
+    # To enable simultaneous bitbaking of two images with the same branch name,
+    # create a new branc in the repo using the basename of the image. (This creates an
+    # additional commit). Fixes OTA-2211.
+    ostree_target_hash=$(cat ${OSTREE_REPO}/refs/heads/${OSTREE_BRANCHNAME})
+    ostree --repo=${OSTREE_REPO} commit \
+           --tree=ref=${ostree_target_hash} \
+           --skip-if-unchanged \
+           --branch=${OSTREE_BRANCHNAME}-${IMAGE_BASENAME} \
+           --subject="${OSTREE_COMMIT_SUBJECT}" \
+           --body="${OSTREE_COMMIT_BODY}"
+
     if [ "${OSTREE_UPDATE_SUMMARY}" = "1" ]; then
         ostree --repo=${OSTREE_REPO} summary -u
     fi
-
-    # To enable simultaneous bitbaking of two images with the same branch name,
-    # create a new ref in the repo using the basename of the image. (This first
-    # requires deleting it if it already exists.) Fixes OTA-2211.
-    ostree --repo=${OSTREE_REPO} refs --delete ${OSTREE_BRANCHNAME}-${IMAGE_BASENAME}
-    ostree_target_hash=$(cat ${OSTREE_REPO}/refs/heads/${OSTREE_BRANCHNAME})
-    ostree --repo=${OSTREE_REPO} refs --create=${OSTREE_BRANCHNAME}-${IMAGE_BASENAME} ${ostree_target_hash}
 }
 
 IMAGE_TYPEDEP_ostreepush = "ostreecommit"


psaavedra:
> Does this work successfully for you? Unfortunately, it does not for me. I get an error like this:
> 
> ```
> ERROR: core-image-minimal-1.0-r0 do_image_garagecheck: Execution of '/home/patrick/Code/updater_repo_master/build/tmp/work/qemux86_64-poky-linux/core-image-minimal/1.0-r0/temp/run.do_image_garagecheck.14662' failed with exit code 1:
> Using oauth2 authentication token
> OSTree commit 3f4f07b4464310e10cc7a41c3e92df431c8d5b23d0c321077e3039f3bc094f6d is missing in treehub
> Check if the ref is present on the server or in targets.json failed
> ```
> 
> The problem here is that when you add another commit, then the subsequent `garagecheck` task that checks for the original commit on the server fails. The hashes won't match as the two commits have unique hashes. Using a ref might be a hack, but that was the way I could still reference the same hash without relying on the branchname.
> 

Sorry, I'm not using garage but a simple _porvisioner_ done for my own: https://gitlab.com/saavedra.pablo/ostree-provisioner. 

This one executes a  `ostree pull $remote $ref` for systemd updates:

When the ref is a hash or the distro references works well. However, when I try to use the image based reference ostree throws the error described in #583 .

> I could reproduce the problem you described in #583, but I'm not really convinced it _is_ a problem. What actually goes wrong for you with the current implementation? You mention the ref is malformed, but what are the implications of that?


Apparently the way of how is created this reference for the branch is not OK for ostree ( see: `_ostree_repo_verify_bindings()` in https://github.com/ostreedev/ostree/blob/master/src/libostree/ostree-repo-pull.c#L1590). I guess because, the commit is not associated to the branch.



... 

if garage needs a unique hash for all the references, I wonder if could be a way to make a single commit/hash referenciated to much more than one branches at the same time.

patrickvacek:
> However, when I try to use the image based reference ostree throws the error described in #583 .

I'm not sure what you mean by "image based reference" here. (Nor "distro reference".) Every ref is just a hash. What I really wanted here was something like a git tag, but to my knowledge ostree doesn't support tags, and refs were as close as I could get.

> I guess because, the commit is not associated to the branch.

I also don't understand that. The ref is made on the same branch as the commit. But I realize that's an ostree thing there must be some implicit meaning to a ref that doesn't map directly to the concept of a tag.

> if garage needs a unique hash for all the references, I wonder if could be a way to make a single commit/hash referenciated to much more than one branches at the same time.

I don't think that's quite the problem. The problem is that when bitbaking simultaneously, you can have multiple commits on one or more branches, and each task in `image_types_ostree` and `image_types_ota` needs a way to make sure each bitbake instance can refer to the appropriate commit. If the branch is the same in multiple bitbake instances, we need some way to distinguish the two commits. If you have other ideas, I'm open to it.

All that said, I'm not sure the bug this was intended to fix is still relevant. It was caused by simultaneously bitbaking the primary-image and secondary-image in our secondary tests. Since then, we changed the default `SOTA_HARDWARE_ID` (and thus the `OSTREE_BRANCHNAME`) [for the secondary image](https://github.com/advancedtelematic/meta-updater/commit/d0b6547fde5038cb7ab9837681ea370af24de855#diff-29c0907360d6172143c95ab1720edeacR5). Therefore, we probably don't even need the ref logic anyway. Which then makes me wonder: do you need it? What's stopping you from using the existing branch name and commit?

psaavedra:
> > However, when I try to use the image based reference ostree throws the error described in #583 .
> 
> I'm not sure what you mean by "image based reference" here. (Nor "distro reference".) Every ref is 

* distro ref: https://github.com/advancedtelematic/meta-updater/pull/584/files#diff-e2abc18e16bca9de6396cc6ab17ac27fL155
* image based ref: https://github.com/advancedtelematic/meta-updater/pull/584/files#diff-e2abc18e16bca9de6396cc6ab17ac27fL157

> just a hash. What I really wanted here was something like a git tag, but to my knowledge ostree 
> doesn't support tags, and refs were as close as I could get.
> 
> > I guess because, the commit is not associated to the branch.
> 
> I also don't understand that. The ref is made on the same branch as the commit. But I realize that's an ostree thing there must be some implicit meaning to a ref that doesn't map directly to the concept of a tag.
> 
> > if garage needs a unique hash for all the references, I wonder if could be a way to make a single commit/hash referenciated to much more than one branches at the same time.
> 
> I don't think that's quite the problem. The problem is that when bitbaking simultaneously, you can have multiple commits on one or more branches, and each task in `image_types_ostree` and `image_types_ota` needs a way to make sure each bitbake instance can refer to the appropriate commit. If the branch is the same in multiple bitbake instances, we need some way to distinguish the two commits. If you have other ideas, I'm open to it.
> 
> All that said, I'm not sure the bug this was intended to fix is still relevant. It was caused by simultaneously bitbaking the primary-image and secondary-image in our secondary tests. Since then, we changed the default `SOTA_HARDWARE_ID` (and thus the `OSTREE_BRANCHNAME`) [for the secondary image](https://github.com/advancedtelematic/meta-updater/commit/d0b6547fde5038cb7ab9837681ea370af24de855#diff-29c0907360d6172143c95ab1720edeacR5). Therefore, we probably don't even need the ref logic anyway. Which then makes me wonder: do you need it? What's stopping you from using the existing branch name and commit?

It is not necessary a blocker but I was looking for a simple way to have a `HEAD` reference for all the images (each one in its own  different ostree branch) created in a single ostree repository.

If this is not possible probably I will use a different repository (directory) for each built image.

Anyway, the `image based ref` reference as is defined in https://github.com/advancedtelematic/meta-updater/pull/584/files#diff-e2abc18e16bca9de6396cc6ab17ac27fL15 is still not OK for allowing `pulling`.

patrickvacek:
> It is not necessary a blocker but I was looking for a simple way to have a `HEAD` reference for all the images (each one in its own different ostree branch) created in a single ostree repository.

This is already possible if each image has a different `OSTREE_BRANCHNAME` value. You can safely ignore the `${OSTREE_BRANCHNAME}-${IMAGE_BASENAME}` hack; that was only done to fix an issue with two different image types bitbaking with the same branch.

> Anyway, the `image based ref` reference as is defined in https://github.com/advancedtelematic/meta-updater/pull/584/files#diff-e2abc18e16bca9de6396cc6ab17ac27fL15 is still not OK for allowing `pulling`.

That is true and unfortunate. However, we don't ever need to pull that specific ref, and it doesn't sound like you need to either. As I said, I think that logic is no longer necessary at all, so we might be able to just get rid of it entirely.

> distro ref: https://github.com/advancedtelematic/meta-updater/pull/584/files#diff-e2abc18e16bca9de6396cc6ab17ac27fL155

That line deletes a ref in case it already exists so it can be recreated two lines below:
```
ostree --repo=${OSTREE_REPO} refs --delete ${OSTREE_BRANCHNAME}-${IMAGE_BASENAME}
```

Is that the line you meant to link to? Anyway, I'm not sure it matters if that logic is unnecessary anymore.

psaavedra:
(Taking up this topic)

> > It is not necessary a blocker but I was looking for a simple way to have a `HEAD` reference for all the images (each one in its own different ostree branch) created in a single ostree repository.
> 
> This is already possible if each image has a different `OSTREE_BRANCHNAME` value. You can safely ignore the `${OSTREE_BRANCHNAME}-${IMAGE_BASENAME}` hack; that was only done to fix an issue with two different image types bitbaking with the same branch.

I think I spoke poorly. I still think to have a unique branch even for different machines it is still  cool feature. But maybe it is not good idea mix different images in the same branch.  

Having said that If you think not worth to work in a solution for that then I agree in close this PR.

patrickvacek:
> I still think to have a unique branch even for different machines it is still cool feature.

By default, `OSTREE_BRANCHNAME ?= "${SOTA_HARDWARE_ID}"` and `SOTA_HARDWARE_ID ??= "${MACHINE}"`. Unless you've changed those variables, you will get a unique branch per `MACHINE`. Or am I still misunderstanding what you need?

patrickvacek:
@psaavedra Is this still something you're interested in? From the sound of it, what you want to do should already work, and the code that you were concerned about appears to be unnecessary and can probably be removed.

patrickvacek:
Thanks again for your work here, but https://github.com/advancedtelematic/meta-updater/pull/614 has done the job quite simply and it appears to fix the problem.

psaavedra:
> Thanks again for your work here, but #614 has done the job quite simply and it appears to fix the problem.

Yes. Sorry for the delay replying. The #614 fix the issue. I was looking for a moment to test it properly. Thanks to keep this in mind and to fix it at last.

PR #585, opened 2019-08-15
Fix/ostree branchname 2

lbonn:
Another fix for that: do not export variables, it's not needed and completely wipes out the cache.


lbonn:
Yes, I've changed OSTREE_BRANCHNAME and it was only a couple of tasks. I would like to run CI anyway on that though if we can fix it...

PR #587, opened 2019-08-19
Expose OSTREE_BOOT_PARTITION in do_image_ota

lbonn:
Used by ostree-grub-generator


lbonn:
This is what's in other branches, before the line:

```
	if ${@bb.utils.contains('IMAGE_FSTYPES', 'otaimg', 'false', 'true', d)}; then
		return
	fi
	if [ -z "${OSTREE_REPO}" ]; then
		bbfatal "OSTREE_REPO should be set in your local.conf"
	fi
	if [ -z "${OSTREE_OSNAME}" ]; then
		bbfatal "OSTREE_OSNAME should be set in your local.conf"
	fi
	if [ -z "${OSTREE_BRANCHNAME}" ]; then
		bbfatal "OSTREE_BRANCHNAME should be set in your local.conf"
	fi
```

Just sanity checks.

PR #588, opened 2019-08-19
Add Zabbix configuration recipe

kbushgit:
Zabbix configuration recipe allows to configure zabbix_agernt in order to connect to zabbix_server app:

Options:
SOTA_COMM_CONF_ZABBIX_SERVER // set the zabbix "Server" address
SOTA_COMM_CONF_ZABBIX_SERVERACTIVE // zabbix "ServerActive" address
SOTA_COMM_CONF_ZABBIX_HOSTNAME // zabbix "Hostname"

### DIFF
--- a/recipes-connectivity/zabbix/zabbix_%.bbappend
+++ b/recipes-connectivity/zabbix/zabbix_%.bbappend
@@ -1,0 +1,20 @@
+
+do_install_append() {
+
+    # Set the zabbix Server 
+    if [ ! -z ${SOTA_COMM_CONF_ZABBIX_SERVER} ]; then
+        sed -i "s/Server=\([0-9]\{1,3\}\.\)\{3\}[0-9]\{1,3\}/Server=${SOTA_COMM_CONF_ZABBIX_SERVER}/g" ${D}${sysconfdir}/zabbix_agentd.conf
+        if ! grep -Fxq "Server=${SOTA_COMM_CONF_ZABBIX_SERVER}" ${D}${sysconfdir}/zabbix_agentd.conf; then
+            echo -e '\nServer='${SOTA_COMM_CONF_ZABBIX_SERVER} >> ${D}${sysconfdir}/zabbix_agentd.conf
+        fi
+    fi
+
+    # Set ServerActive
+    if [ ! -z ${SOTA_COMM_CONF_ZABBIX_SERVERACTIVE} ]; then
+        sed -i "s/ServerActive=\([0-9]\{1,3\}\.\)\{3\}[0-9]\{1,3\}/ServerActive=${SOTA_COMM_CONF_ZABBIX_SERVERACTIVE}/g" ${D}${sysconfdir}/zabbix_agentd.conf
+        if ! grep -Fxq "ServerActive=${SOTA_COMM_CONF_ZABBIX_SERVERACTIVE}" ${D}${sysconfdir}/zabbix_agentd.conf; then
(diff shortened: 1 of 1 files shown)


kbushgit:
> Can you document somewhere the steps you took to get the server running and connected to a client running zabbix?

Yes, I've added comments in OTA-2150, please check an let me know if something is not clear.
In general the server configuration is default and you can use official documentation or links I had provided in the comment.

recipes-connectivity/zabbix/zabbix_%.bbappend
```diff
@@ -0,0 +1,30 @@
+FILESEXTRAPATHS_prepend := "${THISDIR}/${PN}:"
```
patrickvacek: What is this line for?

kbushgit: probably is unnecessary(: thanks!

kbushgit:
I think if we remove the "Hostname" config, this recipe could have some practical sense, even for the production case. 

The Server and ServerActive are ok and they have a practical sense and this option can be applied to the group of the devices, so the only "Hostname"  must have a unique name for each device, and normally this must be set manually by Admin.

So my proposal is to remove "Hostname" option, in this case, it will have a "system hostname" which is ok for the testing purpose.

patrickvacek:
Is the hostname here specific to the device or specific to the server where the information is reported? I had thought it was the server, which would be fine, but if it is per-device, then I agree, we probably shouldn't put it in the image. Is there some way to read it from the actual hostname of the device, or the device ID from aktualizr?

kbushgit:
> Is the hostname here specific to the device or specific to the server where the information is reported? I had thought it was the server, which would be fine, but if it is per-device, then I agree, we probably shouldn't put it in the image. Is there some way to read it from the actual hostname of the device, or the device ID from aktualizr?

Actually the actual hostname will be applied by default, and yes we can obtain actualizr device ID and apply it.

patrickvacek:
> Actually the actual hostname will be applied by default, and yes we can obtain actualizr device ID and apply it.

Cool, that sounds good. IIRC the default hostname on qemu is the hardware ID ("qemux86-64"), which might not be helpful, but if you can change it to the device ID, that'd be great. How does one do that?

kbushgit:
> Cool, that sounds good. IIRC the default hostname on qemu is the hardware ID ("qemux86-64"), which might not be helpful, but if you can change it to the device ID, that'd be great. How does one do that?

For the testing purpose, qemux86-64 is fine, but if someone decides to use Zabbix agent for production, it requires some additional work in run time, I'm not sure it should be a part of aktualizr.
It requires to patch config file and restart zabbix service. 
So we can add the script, as a part of Zabbix recipe, which will be able to run aktualizr-info grep an ID and patch Zabbix config file and then restart zabbix service, this script can be run as a system-d unit with the type OneShot.

patrickvacek:
> For the testing purpose, qemux86-64 is fine, but if someone decides to use Zabbix agent for production, it requires some additional work in run time, I'm not sure it should be a part of aktualizr.

Agreed. No need to worry about it for now, then.

> It requires to patch config file and restart zabbix service.
> So we can add the script, as a part of Zabbix recipe, which will be able to run aktualizr-info grep an ID and patch Zabbix config file and then restart zabbix service, this script can be run as a system-d unit with the type OneShot.

Thanks for the details! This can probably also go in the document that describes how to set all of this up.

PR #589, opened 2019-08-19
aktualizr: aktualizr-repo was renamed to uptane-generator.

patrickvacek:
Why were `aktualizr-repo` and `aktualizr-cert-provider` listed in `RDEPENDS_${PN}-host-tools`? I see them added in 7ec1ffc4 but I'm wondering if that was really desirable or necessary. After renaming `aktualizr-repo`, I got a complaint for `uptane-generator`, so I removed it to solve the problem, as it wasn't doing anything. Not sure why the reference to `aktualizr-cert-provider` is necessary either, but I left it untouched. @liuming50 do you have any insight here?


liuming50:
Hi, Patrick:

My original intention was to introduce a package to rdepend on all the
binary packages (already split to per-binary packages), as a convenience in
case a end user wants to install all the tools binaries.

the best,
thank you


Why were aktualizr-repo and aktualizr-cert-provider listed in

patrickvacek:
> My original intention was to introduce a package to rdepend on all the binary packages (already split to per-binary packages), as a convenience in case a end user wants to install all the tools binaries.

Okay, that makes sense. Thanks for the quick clarification! So it won't hurt to remove it since it wasn't being used.

patrickvacek:
FYI I bumped the aktualizr version to the latest. Once https://github.com/advancedtelematic/aktualizr/pull/1342 is merged, I'll make a new tag/release there and then change the aktualizr version here to match that.

patrickvacek:
This is ready for review to match the [2019.7 release](https://github.com/advancedtelematic/aktualizr/releases/tag/2019.7).

PR #590, opened 2019-08-20
aktualizr: latest garage-sign with expiration support.

patrickvacek:
Use GARAGE_TARGET_EXPIRES or GARAGE_TARGET_EXPIRE_AFTER (but not both!) to set the garage-sign parameters.

Tested manually with all combinations.


classes/sota.bbclass
```diff
@@ -51,6 +51,8 @@ GARAGE_SIGN_KEYNAME ?= "garage-key"
 GARAGE_TARGET_NAME ?= "${OSTREE_BRANCHNAME}"
 GARAGE_TARGET_VERSION ?= ""
 GARAGE_TARGET_URL ?= ""
+GARAGE_TARGET_EXPIRES ?= ""
+GARAGE_TARGET_EXPIRE_AFTER ?= ""
```
simao: What if none of these are set?

patrickvacek: Then nothing is specified and the default from `garage-sign` is applied.

simao: but `garage-sign` doesn't have a default anymore. By default it will use what is in the unsigned role. If the user just uses bitbake it will expire.

patrickvacek: Default still appears to be one month.

jochenschneider: The default of `targets init` is 31 days, yes. This doesn't get increased when signing anymore, though.

patrickvacek: Ah, sorry, missed a comment. So what's the right thing to do? Should we set a default of our own (31 days, presumably) if both variables are unset?

jochenschneider: Depends on what behavior you want. If you want the default behavior to mimic the previous behavior, you could use `GARAGE_TARGET_EXPIRE_AFTER ?= "1m"` if I understand that syntax correctly, `1m` standing for "1 month". (Actually before it was more like `GARAGE_TARGET_EXPIRE_AFTER ?= "31d"`, but `1m` probably makes more sense in general.)

patrickvacek: Thanks, done!

lbonn:
Do you think it's important enough that it deserves an entry in conf/local.conf.sample.append.

I'm also not sure that the documentation is self-contained here, users have to look at garage sign docs to understand what this is for.

patrickvacek:
> Do you think it's important enough that it deserves an entry in conf/local.conf.sample.append.

Probably not, since there is a reasonably default value now.

> I'm also not sure that the documentation is self-contained here, users have to look at garage sign docs to understand what this is for.

I agree with that. I explained the two variables in the README, but perhaps we need a new section? Or is this something that should be explained in the official docs? What do you think? (@tkfu or @merltron, any opinions from your side?)

merltron:
> What do you think? (@tkfu or @merltron, any opinions from your side?)

Yes, it should be in the offical docs. Add this to the README for now, and I'll create a case to have it documented poperly. It should be in a procedure to supplement this topic: https://docs.ota.here.com/ota-client/dev/metadata-expiry.html

NOTE: I plan to merge all of this merge all of this content into the docs portal source anyway. (like we did for aktualizr). So most doc in Github would by definition be "the offical docs"....there should be very little content that counts as "unoffical docs".

tkfu:
> I agree with that. I explained the two variables in the README, but perhaps we need a new section? Or is this something that should be explained in the official docs? What do you think? (@tkfu or @merltron, any opinions from your side?)

We should probably add it to the docs portal, yeah--I can do it today. We still refresh targets metadata on the server side for users who haven't rotated their keys offline, right?

patrickvacek:
Added a new section to the README to briefly explain the metadata expiry options. Is that sufficient?

merltron:
> Added a new section to the README to briefly explain the metadata expiry options. Is that sufficient?

I guess so,  for now, we need to add an equivalent section to the "docs" folder in the aktualizr repo (which is currently how it gets into the portal).

But I would rewrite it to fit the other content we have about metadata expiry. I'd prefer to use a copy of the same content here, but at the same time, I don't want to delay this PR as I see there are config changes needed.

patrickvacek:
> I guess so, for now, we need to add an equivalent section to the "docs" folder in the aktualizr repo (which is currently how it gets into the portal).
>
> But I would rewrite it to fit the other content we have about metadata expiry. I'd prefer to use a copy of the same content here, but at the same time, I don't want to delay this PR as I see there are config changes needed.

Agreed, except that I think it's best to have one definitive source of truth and have everything else just point at that. I already added a link to what seemed like where the information would go in the portal, so hopefully I've got a head start on that. My explanation here was admittedly brief, but that's because I knew it would be an interim solution and I didn't want to bikeshed it too much.

But this reminds me of a related problem... once it's in the doc portal, then a casual observer will think that anything there applies to any version/branch of Yocto/meta-updater, which will not immediately be the case. This is only merging to master, and it could be a few weeks before we backport it.

merltron:
> But this reminds me of a related problem... once it's in the doc portal, then a casual observer will think that anything there applies to any version/branch of Yocto/meta-updater, which will not immediately be the case. This is only merging to master, and it could be a few weeks before we backport it.

Yes, I have this ticket to handle versioning: https://saeljira.it.here.com/browse/OTA-3618

However, this would necessitate a separate guide for this content. My working title is "Using the OTA Connect build tools "

PR #591, opened 2019-08-26
Use --wait-until-provisioned option for IP Secondary Test

mike-sul:
Use --wait-until-provisioned option while getting aktualizr status in the ip secondary tests

### DIFF
--- a/lib/oeqa/selftest/cases/updater_qemux86_64.py
+++ b/lib/oeqa/selftest/cases/updater_qemux86_64.py
@@ -360,17 +360,7 @@ class IpSecondaryTests(OESelftestTestCase):
             self._test_ctx.append_config('SOTA_CLIENT_PROV = " aktualizr-shared-prov "')
 
         def is_ecu_registered(self, ecu_id):
-            max_number_of_tries = 120
-            try_counter = 0
-
-            # aktualizr-info is not always able to load ECU serials from DB
-            # so, let's run it a few times until it actually succeeds
-            while try_counter < max_number_of_tries:
-                device_status = self.get_info()
-                try_counter += 1
-                if device_status.find("load ECU serials") == -1:
-                    break
-                sleep(1)
+            device_status = self.get_info()
 
             if not ((device_status.find(ecu_id[0]) != -1) and (device_status.find(ecu_id[1]) != -1)):
                 return False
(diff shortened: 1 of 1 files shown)


mike-sul:
> Is the while loop in "is_ecu_registered(self, ecu_id)" OK?
Well, I recall, running into the case of `aktualizr-info` returning no any info about installed ECUs even after a device registration/provisioning. IIRC, it had to do with an access to the DB, if its locked (on write) aktualizr-info  fails. I'll double check if it still true in case of `--wait-until-provisioned`.

mike-sul:
`oe-selftest -r updater_native updater_qemux86_64 updater_minnowboard updater_raspberrypi` has passed
`2019-08-26 16:49:03,909 - oe-selftest - INFO - oe-selftest - OK - All required tests passed (successes=19, skipped=0, failures=0, errors=0`

mike-sul:
It seems like the `while` loop can be removed from `is_ecu_registered` function.

lib/oeqa/selftest/cases/updater_qemux86_64.py
```diff
@@ -379,7 +369,7 @@ def is_ecu_registered(self, ecu_id):
             return not_reg_start == -1 or (device_status.find(ecu_id[1], not_reg_start) == -1)
 
         def get_info(self):
-            stdout, stderr, retcode = self.send_command('aktualizr-info')
+            stdout, stderr, retcode = self.send_command('aktualizr-info --wait-until-provisioned', timeout=620)
```
patrickvacek: `620` seems like an arbitrary number, and perhaps a bit high. Any particular reason for that choice? Do you think it needs to be that long?

mike-sul: I think, this number has to be higher than this one https://github.com/advancedtelematic/meta-updater/blob/9b397b1e047c23921e07258b8a1664d55d850488/recipes-test/demo-config/shared-conf.inc#L5 which is equal to 240 currently.

patrickvacek: Yes, agreed, but almost three times as long is probably more than necessary. But it doesn't really matter, and let's just go with what you have until we have a good reason to change it. 10 minutes isn't too bad, anyway.

PR #592, opened 2019-08-28
OTA-3594: adding restructured Antora-friendly content

merltron:
This is the "meta-updater" equivalent of the merge that we did for aktualizr docs. 

-Adding Antora dir structure
- Break up content into individual files for each section.
- Setting up files for including in other repos
- Making options easier to read and more consistent with other
reference doc.
- Updated README to link to individual topic files rather than
all on one page.
- Added Mathias's requested HW and bootloader info
- Added xaviers requested NRE clause.

Doc Infrastructure Note:
The meta-updater docs as been packaged as an Antora "component" (distinct guide) but is hidden from the portal nav. Rather, I would prefer to include the main pieces of content in the developer guide (with include directives). But I can only use include directives if this doc is configured as a Antoria "component" (ID: "ota-build") - thats why I've configured it to be a component but hidden it from the nav.

Signed-off-by: Merlin Carter <merlin.carter@here.com>


README.adoc
```diff
@@ -1,309 +1,49 @@
 = meta-updater
 :toc: macro
 :toc-title:
+:meta-updater-github-url: https://github.com/advancedtelematic/meta-updater/blob/master/
```
patrickvacek: I wish there was a way to get this to point at whatever the current branch is instead of always assuming master. @tkfu is that asking for too much?

docs/modules/ROOT/pages/build.adoc
```diff
@@ -0,0 +1,62 @@
+= Build
+:meta-updater-github-url: https://github.com/advancedtelematic/meta-updater/tree/master
+
+== Quickstart
+
+If you don't already have a Yocto project that you want to add OTA to, you can use the xref:dev@getstarted::raspberry-pi.adoc[HERE OTA Connect Quickstart] project to rapidly get up and running on a Raspberry Pi. It takes a standard https://www.yoctoproject.org/tools-resources/projects/poky[poky] distribution, and adds OTA and OSTree capabilities.
+
+== Dependencies
+
+//MC: TOMERGE: These "dependencies" mostly just duplicates the prerequisite sections: https://main.gitlab.in.here.com/olp/edge/ota/documentation/ota-connect-docs/blob/master/docs/getstarted/modules/ROOT/pages/raspberry-pi.adoc
```
patrickvacek: Yes, but the target audience is slightly different. This assumes that the user already has a Yocto environment, so this specifically calls out only the extra things that we need in addition to the default Yocto dependencies.

That said, this may or may not be useful, so I'm not saying it has to stay the way it is.

docs/modules/ROOT/pages/meta-updater-testing.adoc
```diff
@@ -0,0 +1,44 @@
+= Testing
+
+//MC: No overlap with any content currently in the developer guide, but probably useful content to clean up and include.
```
patrickvacek: BWT this is mostly only interesting for contributors or people integrating meta-updater into another Yocto project.

docs/modules/ROOT/pages/meta-updater-usage.adoc
```diff
@@ -0,0 +1,87 @@
+= Usage
+:meta-updater-github-url: https://github.com/advancedtelematic/meta-updater/tree/master
+:metadata-expiry-article: xref:dev@ota-client::metadata-expiry.adoc[OTA Connect documentation]
+ifdef::env-github[:metadata-expiry-article: link:https://docs.ota.here.com/ota-client/dev/metadata-expiry.html[OTA Connect documentation]]
+
+//MC: No overlap with any content currently in the developer guide, but probably useful content to clean up and include. eg: use cases.
+
+== OSTree
```
patrickvacek: FYI I'm not sure anyone has used or reviewed this OSTree section in years. It may or may not still be relevant and useful.

docs/modules/ROOT/pages/sota-variables.adoc
```diff
@@ -0,0 +1,50 @@
...
+| `OSTREE_UPDATE_SUMMARY`|Set this to '1' to update summary of OSTree repository on each commit. '0' by default.
+| `OSTREE_DEPLOY_DEVICETREE`|Set this to '1' to include devicetree(s) to boot
+| `GARAGE_SIGN_AUTOVERSION`|Set this to '1' to automatically fetch the last version of the garage tools installed by the aktualizr-native. Otherwise use the fixed version specified in the recipe.
+| `GARAGE_TARGET_URL` | Sets the `--url` parameter of `garage-sign targets add`, which sets a custom URL for the Image repository targets.
+ifndef::env-github[]
+[NOTE]
+====
+Currently, this only works when using the master branch of meta-updater.
+====
+endif::[]
```
patrickvacek: Is this note for `GARAGE_TARGET_URL`? That exists in all supported branches.

README.adoc
```diff
@@ -1,309 +1,49 @@
...
++
+Learn how to use the `oe-selftest` framework for quality assurance.
++
+* xref:{meta-updater-github-url}docs/modules/ROOT/pages/meta-updater-testing.adoc#_aktualizr_test_suite_with_ptest[Aktualizr test suite with ptest]
++
+Learn how to enable Yocto's package test functionality and run parts of the aktualizr test suite.
++
+* xref:{meta-updater-github-url}docs/modules/ROOT/pages/provisioning-methods.adoc[Provisoning methods]
++
+Learn how to enable different methods for provisioning devices.
```
patrickvacek: Does it make sense to explain more explicitly that most docs can be found in the aktualizr repo and/or docs portal?

README.adoc
tkfu: That's possible, yes, though it's mildly ugly. You'd need these lines in place of the line above in each file that uses it:

```
ifdef::env-github[:meta-updater-github-url: ../../../../]
ifndef::env-github[:meta-updater-github-url: https://github.com/advancedtelematic/meta-updater/blob/master/]
```

tkfu: @merltron This also gets at some important IA questions surrounding versioning. Antora's conceptual model of "components" is designed to make it easy to have docs affecting a specific component get versioned. However, this depends on having the docs structured in a way that the docs for each component are more or less independent. Right now, that's not generally the case; the developer guide has content that could potentially change based on the meta-updater version and/or the aktualizr version, including many files that talk about the different ways to set options in the sota.toml file vs. yocto's local.conf.

I think that overall that's probably the right decision, but it does mean that our versioning strategy, once we sort it out better, will mean that there will have to be a version option in the docs for each possible tuple of aktualizr_version and meta_updater_version.

@patrickvacek Let's have a meeting about versioning/tagging plans going forward? I think in my ideal world, every time we made a release of aktualizr the embedded team would decide which branches they wanted to backport it to, and then tag a commit on each branch with something like `rocko-2019.5`. (Also in this ideal world, no branch but master would ever be building an aktualizr that isn't a point release.)

patrickvacek: > That's possible, yes, though it's mildly ugly.

Do you have a strong preference here? I'll generally default to you and Merlin, but I'm still gonna express my opinions. :)

> Let's have a meeting about versioning/tagging plans going forward?

Sure, sounds like a good idea.

> I think in my ideal world, every time we made a release of aktualizr the embedded team would decide which branches they wanted to backport it to, and then tag a commit on each branch with something like `rocko-2019.5`.

That sounds like a reasonable strategy. However, I would posit that every release should be backported to every Yocto release we claim to actively support. To make that easier, we should try to limit how many releases we support, and that why we should drop rocko/sumo ASAP.

We might also want to try to automate the release process to simplify this somewhat. There's a ticket for that work, but it hasn't been a high priority since it looks like a fair bit of work for not much gain. But if we want to release more often, it might be worth it.

> Also in this ideal world, no branch but master would ever be building an aktualizr that isn't a point release.

That also sounds reasonable.

merltron: > I think that overall that's probably the right decision, but it does mean that our versioning strategy, once we sort it out better.

@tkfu : Yes, but when Patrick raised the topic of versioning, I did have a consider having a separate guide about the "OTA Connect Yocto Layer" or more generally the "OTA Connect Build Tools". It's just a challenge to give people some of the more general conceptual information. Once solution is to treat some "components" more as "Chapters", so that you're using the top nav to cycle through the different chapters.  (ie: most conceptual content doesn't need versioning) Its just a challenge as to how to slice it up. But a meeting is a good idea.

merltron: > That's possible, yes, though it's mildly ugly. You'd need these lines in place of the line above in each file that uses it:
> 
> ```
> ifdef::env-github[:meta-updater-github-url: ../../../../]
> ifndef::env-github[:meta-updater-github-url: https://github.com/advancedtelematic/meta-updater/blob/master/]
> ```
@tkfu 
Why is this going 4 levels up if the "env" is Github? - the README is already in the root of the repo. I originally used just relative path, but Github seems to want `blob/<branch>/` to be in the path.

docs/modules/ROOT/pages/sota-variables.adoc
merltron: oops - No, it was intended for the expiry parameters - will fix

README.adoc
merltron: Yep, good idea, will add

README.adoc
tkfu: Ah, you're right. I didn't notice this was the top-level README--I thought it was part of the content catalog.

@patrickvacek Unfortunately, for READMEs, there's no good way to do relative links on github. The most likely case is that the reader is viewing it from https://github.com/advancedtelematic/meta-updater, but it's also possible that they're viewing it from https://github.com/advancedtelematic/meta-updater/blob/master/README.adoc. We can't even distinguish between those two cases. Add in the possibility of https://github.com/advancedtelematic/meta-updater/tree/{branchname} and https://github.com/advancedtelematic/meta-updater/blob/{branchname}/README.adoc and it's even more hairy.

Even if we don't consider the possibility of someone reading via the `/blob/` urls, there's no relative link that works in the https://github.com/advancedtelematic/meta-updater/tree/{branchname} case, because we need to replace 'tree' with 'blob' in the link, and there's no URL rewriting logic available to us in the github renderer.

merltron: Updated to use "tree" instead of "blob" which seems to work better.

docs/modules/ROOT/pages/sota-variables.adoc
merltron: fixed

README.adoc
merltron: done

docs/modules/ROOT/pages/meta-updater-usage.adoc
merltron: Yes, I'll do a quality review and rewrite of this content in a second PR. This PR just to get it into the format/structure that Antora expects.

README.adoc
tkfu: I think they're basically equivalent nowadays...github will redirect tree to blob when it's a file, and redirect blob to tree when it's a directory (both of those cases used to be 404s). It's just that we still don't have a solution for the most common case, where the source page is the main page of the project and there is no branch specified in the URL.

PR #597, opened 2019-09-05
raspberrypi4 support

lbonn:


lbonn:
rpi4 32 bits uccessfully booted in combination with https://github.com/advancedtelematic/meta-updater-raspberrypi/pull/60.

raspberrypi4-64 still needs a little bit of work...

lbonn:
I'll just quickly boot a raspberrypi3 and see if this did not break it.

classes/sota_raspberrypi.bbclass
```diff
@@ -21,20 +21,50 @@ UBOOT_ENTRYPOINT_sota ?= "0x00008000"
...
+        if dtb.endswith('dtb'):
+            return base
+        elif dtb.endswith('dtbo'):
+            return '{};{}'.format(base, dtb)
+
+    return ' '.join([transform(dtb) for dtb in alldtbs.split(' ') if dtb])
+
+IMAGE_BOOT_FILES_sota = "bcm2835-bootfiles/* \
+                         u-boot.bin;${SDIMG_KERNELIMAGE} \
+                         "
```
patrickvacek: Does this also need to be rpi4-specific as well?

lbonn: No it's needed and was there before. It just has been split in several lines in my version.

lbonn:
The last commit makes it work on 64 bit! I'll test older platforms though.

lbonn:
Ok rpi3, rpi3-64, rpi4 still boot with last changes. Couldn't find a rpi2 lying around.

rpi (first generation) was already broken, it looks like...

classes/sota_raspberrypi.bbclass
```diff
@@ -16,25 +16,55 @@ DEV_MATCH_DIRECTIVE_pn-networkd-dhcp-conf = "Driver=smsc95xx lan78xx"
 IMAGE_INSTALL_append_sota = " virtual/network-configuration "
 
 PREFERRED_PROVIDER_virtual/bootloader_sota ?= "u-boot"
-UBOOT_ENTRYPOINT_sota ?= "0x00008000"
+UBOOT_ENTRYPOINT_sota ?= "0x00080000"
```
xcheng-here: Just a little bit curious about this part. I think this address should be the entry point of kernel, so it should be same as the TEXT_OFFSET in the kernel. I found the TEXT_OFFSET for rpi3-32 is 0x8000 while it's 0x80000 for rpi3-64. So, maybe use 0x8000 for 32 bit build and use 0x80000 for 64 bit build would make it even better. But the most interesting thing I think is that why the rpi3-32 still work well after this change ? Since rpi3-32 bit is already verified, maybe I made something wrong, I hope I can do some test about it later. And if you already found what's wrong in my understanding, it will be very appreciated.

lbonn: Thanks for looking into that!

Here is the source for this change: https://gitlab.denx.de/u-boot/u-boot/blob/HEAD/include/configs/rpi.h#L112. From the info I got, it actually doesn't need to match with TEXT_OFFSET exactly as we use zImage which takes care for itself of the relocation to the correct location: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/Documentation/arm/booting.rst#n161

patrickvacek:
> Couldn't find a rpi2 lying around.

I also couldn't find one last time I looked around. I wouldn't worry about it too much.

> rpi (first generation) was already broken, it looks like...

Yeah, we do actually have one but we've never supported them.

classes/sota_raspberrypi.bbclass
xcheng-here: Thank you very much !

PR #600, opened 2019-09-09
Don't try to put the repo manifest in the initramfs image.

patrickvacek:
It isn't very useful there and just takes up space.


eu-smirnov:
Good point! But maybe we even can add it to certain image types only instead of inheriting `image_repo_manifest` in `sota.bbclass`?

patrickvacek:
> But maybe we even can add it to certain image types only instead of inheriting `image_repo_manifest` in `sota.bbclass`?

I wondered about that. I think this way is probably slightly preferable, though, because we want to encourage this as the default behavior. For example, if someone were to build their own custom image, we'd want them to have this enabled by default so that the repo manifest is available without any extra work.

Issue #603, opened 2019-09-12
sstate signature issues


shr-project:
There are quite a few issues reported by oe-core/scripts/sstate-diff-machines.sh script in meta-updater recipes, mostly allarch recipes depending on TUNE_PKGARCH without setting SIGGEN_EXCLUDERECIPES_ABISAFE/SIGGEN_EXCLUDE_SAFE_RECIPE_DEPS

ERROR: 32 issues were found in these recipes: aktualizr-device-prov aktualizr-device-prov-creds aktualizr-device-prov-hsm aktualizr-uboot-env-rollback networkd-dhcp-conf primary-network-config secondary-network-config

Can you please run the script and fix the issues it reports?

patrickvacek:
Hello @shr-project, thanks for reporting this issue! I'm attempting to look into it, but unfortunately haven't been able to wrap my head around the tooling yet. I ran `../poky/scripts/sstate-diff-machines.sh --tmpdir=tmp --targets=core-image-minimal` (with and without `--analyze`), and then I diffed the list.M files as described [here](https://lists.yoctoproject.org/pipermail/poky/2014-August/009840.html). However, I can't find the sigdata files referenced in those. Thus I can't get `bitbake-diffsigs` to provide meaningful output. What do I do next?

Is there documentation for how to use these tools?

shr-project:
Hi @patrickvacek,
there is short documentation in the script itself.

You need to test it on something which includes all the recipes from meta-updater (e.g. world, core-image-minimal won't include them).

--analyze can be used only when you provide meaningful --machines parameter e.g. qemux86 qemux86copy qemuarm

Here is the output from:
oe-core/scripts/sstate-diff-machines.sh --tmpdir=tmp-glibc --machines="qemux86 qemux86copy qemuarm" --targets=world --analyze
with meta-updater layer included.
For qemux86copy you'll need meta-selftest layer (./meta-selftest/conf/machine/qemux86copy.conf) or you can cherry-pick this commit to oe-core:
http://git.openembedded.org/openembedded-core-contrib/commit/?h=jansa/master&id=dd27e838ad87dcc82e79c1284c76304d4128a230

http://dpaste.com/22CF7JD

Simple exemple would be recipes-sota/aktualizr/aktualizr-device-prov.bb which inherits allarch, but has
`
DEPENDS = "aktualizr aktualizr-native openssl-native"
[...]

PR #612, opened 2019-10-11
Fix/603/arch deps

patrickvacek:
First stab at resolving issues reported in https://github.com/advancedtelematic/meta-updater/issues/603.


recipes-sota/aktualizr/aktualizr-device-prov-creds.bb
```diff
@@ -6,13 +6,20 @@ LIC_FILES_CHKSUM = "file://${COMMON_LICENSE_DIR}/MPL-2.0;md5=815ca599c9df247a0c7
...
 # is to create certificate request directly on the device (either with HSM/TPM
 # or with software) and then sign it with a CA stored on a disconnected machine.
 
-DEPENDS = "aktualizr aktualizr-native"
+# We need to get the config files from the aktualizr-host-tools package built by
+# the aktualizr (target) recipe.
+DEPENDS = "aktualizr aktualizr-native openssl-native"
 ALLOW_EMPTY_${PN} = "1"
 
+# If the config file from aktualizr used here is changed, you will need to bump
```
lbonn: That sounds dangerous. Would we lose a lot if we remove allarch from these recipes?

patrickvacek: That is the other option. But then we have to rebuild the recipes every time aktualizr changes, regardless of whether it actually affects these recipes. I opted for the case that is a bit painful in rare situations over the annoyance of slightly longer builds in the general case.

lbonn: Ok I see. I'm not too worried about the "painful" part but that it's hard to notice that the files have changed on bitbake's side and that it would cause wrong builds.

patrickvacek: > it's hard to notice that the files have changed on bitbake's side and that it would cause wrong builds.

Yes, agreed. I hoped the comment would be a hint, but it is a concern. Originally I was thinking about just removing `allarch` from those recipes, but it seemed like an unnecessary performance hit, so I went this way instead. Do you think the risk of unexpected build problems is too great?

lbonn:
I've pushed another commit that should fix the problem with aktualizr's dependency on MACHINE (hopefully)

lbonn:
> I cannot figure out what in the `do_install` step of `aktualizr-device-prov-creds` depends on `MACHINE`.

There is a reference to `${DEPLOY_DIR_IMAGE}` in the install step which is defined as `${DEPLOY_DIR}/images/${MACHINE}`.

recipes-sota/aktualizr/aktualizr-device-prov-creds.bb
```diff
@@ -32,9 +32,10 @@ export SOTA_CAKEY_PATH
 do_install() {
     if [ -n "${SOTA_PACKED_CREDENTIALS}" ]; then
         if [ -z ${SOTA_CACERT_PATH} ]; then
-            SOTA_CACERT_PATH=${DEPLOY_DIR_IMAGE}/CA/cacert.pem
-            SOTA_CAKEY_PATH=${DEPLOY_DIR_IMAGE}/CA/ca.private.pem
-            mkdir -p ${DEPLOY_DIR_IMAGE}/CA
+            SOTA_CA_DIR=${DEPLOY_DIR}/sota-ca
```
lbonn: It's quite unusual to store things in `${DEPLOY_DIR}` directly.
Making this recipe MACHINE_ARCH sounds better to me.

patrickvacek: > It's quite unusual to store things in `${DEPLOY_DIR}` directly.
> Making this recipe MACHINE_ARCH sounds better to me.

But the CA files we generate are certainly not machine-specific. They aren't specific to anything, really. It's an unnecessary dependency that we can easily avoid. Is there a better place to put them?

lbonn: Well I think the wrong idea we had is that the MACHINE_ARCH packages means that the files should be generated differently for each machine. It seems like it's more about that the resulting files are not shared between machines.

And it's different from the aktualizr case where the recipe is actually expensive to build, this one is very cheap. I'm not convinced it's worth the optimization

But this code path also generates a lot of warnings, it's a bit of a pathological case.

patrickvacek: > this one is very cheap. I'm not convinced it's worth the optimization

That is a fair point. Also, this recipe is fairly unrealistic in that if you actually use the full extent of it, it puts device-specific credentials in an image that can be shared among many devices, so maybe we should change what we're doing here...

> But this code path also generates a lot of warnings, it's a bit of a pathological case.

What other warnings are you referring to?

lbonn: I was talking about all the bbwarn we added.

patrickvacek: I see. Should we just remove this whole recipe? It's doing the "wrong" thing as mentioned above, it's not covered by oe-selftest, and AFAIK it isn't used by anyone. If you want device credential provisioning, you can't really use bitbake to generate device-specific credentials.

lbonn: Yes maybe it doesn't make much sense. A middle ground could be to just remove the logic in `if [ -z ${SOTA_CACERT_PATH} ];` and just bbfail there. It would solve the machine arch dependency issue that we're trying to solve here and we can still remove the recipe later.

patrickvacek: Decent suggestion, but I think it's time to just remove it and be done with it, so that's what I did.

shr-project: > It's quite unusual to store things in `${DEPLOY_DIR}` directly.
> Making this recipe MACHINE_ARCH sounds better to me.

It's wrong to install anything to any DEPLOY* variable directly.

do_install should install just to D and do_deploy installs files from D to DEPLOYDIR and then creates deploy sstate which installs it from DEPLOYDIR to DEPLOY_DIR_IMAGE.

Writting to DEPLOY_DIR directly from do_install (or any other global directory from any other task not covered by sstate) only means that the files won't be there once the component is reused from sstate.

E.g. if you build aktualizr-device-prov-creds, then the files will be in deploy, but if you delete tmp-glibc and build it again, then the files will be missing (because do_install wasn't executed at all when there is valid sstate-cache for aktualizr-device-prov-creds).

So good that this is removed, but keep this in mind for other recipes as well.

patrickvacek: @shr-project Thanks for the detailed description here! I was starting to put these pieces together and realized that this recipe was just not doing the right thing. It was meant to just be for demonstration anyway, but it was demonstrating it wrongly enough that it was better to just remove it and be done with it.

PR #613, opened 2019-10-14
README: update and restructure the links to the dev portal.

patrickvacek:
In the wake of https://github.com/advancedtelematic/aktualizr/pull/1410. That PR must get merged before this can be.

Also: should the links point to the docs portal instead of aktualizr's github pages?

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -2,6 +2,7 @@
 :toc: macro
 :toc-title:
 :aktualizr-docsroot: https://github.com/advancedtelematic/aktualizr/tree/master/docs/ota-client-guide/modules/ROOT/pages/
+:getstarted-docsroot: https://docs.ota.here.com/getstarted/dev/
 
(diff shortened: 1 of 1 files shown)


tkfu:
I'd say yes to linking straight into the docs portal; it's very clearly a better experience than reading github's renderings of the documents.

There's something that's a bit of a weird code smell here, though: the `xref:dev@getstarted::raspberry-pi.adoc` that you removed. That would only work if this file were part of an Antora catalog, and it could only be part of an Antora catalog if it were hardlinked in somewhere. I can't imagine a reason to do that, so hopefully it's just an oversight or a copy-paste error.

patrickvacek:
> I'd say yes to linking straight into the docs portal; it's very clearly a better experience than reading github's renderings of the documents.

I'll fix that now.

> There's something that's a bit of a weird code smell here, though: the `xref:dev@getstarted::raspberry-pi.adoc` that you removed. That would only work if this file were part of an Antora catalog, and it could only be part of an Antora catalog if it were hardlinked in somewhere. I can't imagine a reason to do that, so hopefully it's just an oversight or a copy-paste error.

Agreed. I suspect that was from the brief period where we originally had the meta-updater docs as a separate collection in Antora, before we migrated all of that into the aktualizr repo. It didn't show up as a link in the final rendering, which is why it probably wasn't caught until now. I assume it was just an oversight when we did that last transition.

patrickvacek:
> > I'd say yes to linking straight into the docs portal; it's very clearly a better experience than reading github's renderings of the documents.
> 
> I'll fix that now.

Done. Took a few tries. :)

PR #616, opened 2019-10-21
aktualizr: Bump to 2019.9 release and latest garage-sign.

patrickvacek:


patrickvacek:
> I don't have a way to verify the serial numbers...

Do you mean the hashes of the garage-sign? Generally that should be caught by CI if there was a problem, so that's not something you have to worry about. :) Or did you mean something else? There aren't any big surprises here, I think.

Zee314159:
> I don't have a way to verify the serial numbers...
> 
> Do you mean the hashes of the garage-sign? Generally that should be caught by CI if there was a problem, so that's not something you have to worry about. :) Or did you mean something else? There aren't any big surprises here, I think.

That‘s what I meant. So there's nothing to be worried about.

PR #619, opened 2019-10-22
Fix/rm systemd

patrickvacek:
Currently includes https://github.com/advancedtelematic/meta-updater/pull/616.

Note that this does not change the fact that aktualizr is built in meta-updater as a systemd service.


patrickvacek:
Oops, this was misguided. Please ignore, we do still use systemd for some secondary-related things. However, we probably don't need to build that by default.

PR #621, opened 2019-10-25
Drop ostree_git in favor of the ostree recipe from meta-oe.

ricardosalveti:
A newer ostree recipe is now provided via the meta-oe layer, so drop
local ostree_git recipe in favor of the latest one from meta-oe.

Local bbappend created to enable curl, libarchive and
ostree-grub-generator, which were all enabled by default in ostree_git.


ricardosalveti:
This facilitates maintenance on master but will complicate backporting to older branches, so mostly proposing here for discussion. We could update the recipe locally to sync with latest from meta-oe, but since this layer depends on meta-oe already, it is probably good to let the recipe be maintained by meta-oe instead.

ricardosalveti:
I'm unable to read the gitlab/ostree CI build job log, so unsure why it failed.

lbonn:
The only failure is the ptest stage, that will be fixed here: #623.

It probably means it's going to be fine for this PR but we might as well wait for the merge of the other and relaunch a pipeline.

patrickvacek:
> I'm unable to read the gitlab/ostree CI build job log, so unsure why it failed.

Hey @ricardosalveti, thanks for submitting! So far, this looks good. The ptest thing is not a blocker, and that's the only thing that broke. We don't normally even require that for this type of PR, I was just being extra careful. I'm also going to run some manual tests and make sure it all works as expected.

ricardosalveti:
> Hey @ricardosalveti, thanks for submitting! So far, this looks good. The ptest thing is not a blocker, and that's the only thing that broke. We don't normally even require that for this type of PR, I was just being extra careful. I'm also going to run some manual tests and make sure it all works as expected.

Great, thanks!

patrickvacek:
Tested a full update cycle with qemu and it all went as planned. For some reason I'm having trouble getting my RPi to boot, but that might just be something wrong on my side.

recipes-extended/ostree/ostree_%.bbappend
```diff
@@ -0,0 +1,10 @@
+PACKAGECONFIG_append = " curl libarchive"
+PACKAGECONFIG_class-native_append = " curl"
```
lbonn: Should we remove `soup` from PACKAGECONFIG (on by default), as we use curl instead?

ricardosalveti: I had that initially but removed as curl takes priority and doesn't really affect the build, besides having another dependency. I can remove to make it the same as before, will give it another shot locally.

ricardosalveti:
PR updated to also remove soup.

ricardosalveti:
Added static as we don't have systemd running as part of the initramfs by default (change done in 2019.2).

lbonn:
Do you have any background on this? I found this issue https://github.com/ostreedev/ostree/issues/1845 but this doesn't look like a problem we've had before.

Also, any idea why the `static` PACKAGECONFIG would not be on by default then?

ricardosalveti:
> Do you have any background on this? I found this issue [ostreedev/ostree#1845](https://github.com/ostreedev/ostree/issues/1845) but this doesn't look like a problem we've had before.

This started with 2019.2 when they first added support to log the output to systemd, which is enabled by default, which is why we didn't have the issue with the previous recipe.

> Also, any idea why the `static` PACKAGECONFIG would not be on by default then?

I would say because it is a distro decision to have or not to have systemd as part of the initramfs. This only affects users of switch-root.

patrickvacek:
My ability to test this with RPi has been blocked by a bug in booting introduced with https://github.com/agherzan/meta-raspberrypi/commit/baba59dc284acb4a44e4f6722f78f331f2d94e4e. I haven't forgotten this, just haven't had time to sort it out.

ricardosalveti:
> My ability to test this with RPi has been blocked by an unrelated bug in booting introduced with [agherzan/meta-raspberrypi@baba59d](https://github.com/agherzan/meta-raspberrypi/commit/baba59dc284acb4a44e4f6722f78f331f2d94e4e). I haven't forgotten this, just haven't had time to sort it out.

Cool, thanks for the heads up :-)

I found one issue yesterday that affects the grub targets (intel), and send a patch upstream https://github.com/ostreedev/ostree/pull/1969, so we should also wait for the final fix and backport the patch as well.

ricardosalveti:
Pull request updated to include the additional ostree fix for grub targets (backport from upstream).

PR #624, opened 2019-11-01
OTA-3988: Non-systemd/poky-sota local.conf

mike-sul:
Usage: `source meta-updater/scripts/envsetup.sh qemux86-64 build poky-sota` to build non-systemd image

### DIFF
--- a/conf/distro/poky-sota-systemd.conf
+++ b/conf/distro/poky-sota-systemd.conf
@@ -2,7 +2,7 @@ require conf/distro/poky.conf
 
 require conf/distro/sota.conf.inc
 
-DISTRO = "poky-sota"
+DISTRO = "poky-sota-systemd"
 DISTRO_NAME = "OTA-enabled Linux"
 DISTRO_VERSION = "1.0"
 DISTRO_CODENAME = "sota"
--- a/conf/distro/poky-sota.conf
+++ b/conf/distro/poky-sota.conf
@@ -5,3 +5,5 @@ DISTRO = "poky-sota"
 DISTRO_NAME = "OTA-enabled Linux"
 DISTRO_VERSION = "1.0"
 DISTRO_CODENAME = "sota"
+
+IMAGE_INSTALL_append_sota = " ostree-booted"
--- a/conf/local.nonsystemd.conf.sample.append
+++ b/conf/local.nonsystemd.conf.sample.append
@@ -1,0 +1,19 @@
+
+#
+# meta-updater configuration, see README.adoc and aktualizr's
+# documentation for more options and detailed documentation
+#
+
+MACHINE = "##MACHINE##"
+DISTRO = "poky-sota"
+
+# General SOTA setup
+#SOTA_CLIENT_PROV = "aktualizr-shared-prov"
+#SOTA_PACKED_CREDENTIALS = "/path/to/credentials.zip"
+
+# Uncomment this line to start an ssh server at boot automatically
+#IMAGE_FEATURES += "ssh-server-dropbear"
+
+# Uncomment this line to set the log level of aktualizr to 'debug' (from 'info'
+# by default)
+#IMAGE_INSTALL_append += " aktualizr-log-debug"
--- a/recipes-connectivity/networkd-dhcp-conf/networkd-dhcp-conf.bb
+++ b/recipes-connectivity/networkd-dhcp-conf/networkd-dhcp-conf.bb
@@ -15,7 +15,7 @@ SRC_URI = " \
   "
 PR = "r1"
 
-RDEPENDS_${PN} = "systemd"
+REQUIRED_DISTRO_FEATURES_${PN} = "systemd"
 RCONFLICTS_${PN} = "connman"
 
 S = "${WORKDIR}"
--- a/scripts/envsetup.sh
+++ b/scripts/envsetup.sh
@@ -4,10 +4,11 @@ SCRIPT="envsetup.sh"
 MACHINE="$1"
 BUILDDIR="build"
 DISTRO="poky-sota-systemd"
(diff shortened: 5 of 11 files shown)


conf/distro/poky-sota-systemd.conf
```diff
@@ -9,4 +9,4 @@ DISTRO_CODENAME = "sota"
 
 DISTRO_FEATURES_append = " systemd"
 VIRTUAL-RUNTIME_init_manager = "systemd"
-PREFERRED_RPROVIDER_virtual/network-configuration ??= "networkd-dhcp-conf"
+#PREFERRED_RPROVIDER_virtual/network-configuration ??= "networkd-dhcp-conf"
```
eu-smirnov: why?

recipes-connectivity/networkd-dhcp-conf/networkd-dhcp-conf.bb
```diff
@@ -15,8 +15,8 @@ SRC_URI = " \
   "
 PR = "r1"
 
-RDEPENDS_${PN} = "systemd"
-RCONFLICTS_${PN} = "connman"
+REQUIRED_DISTRO_FEATURES_${PN} = "systemd"
+CONFLICT_DISTRO_FEATURES_${PN} = "connman"
```
eu-smirnov: But connman is not a distro feature, is it?

conf/distro/poky-sota-systemd.conf
mike-sul: leftover from my experiments, corrected it.

recipes-connectivity/networkd-dhcp-conf/networkd-dhcp-conf.bb
mike-sul: I am not sure if I actually understand  a notion of distro feature at all :), but based of what I read here https://www.yoctoproject.org/docs/2.5/ref-manual/ref-manual.html#var-DISTRO_FEATURES and here https://www.yoctoproject.org/docs/2.5/ref-manual/ref-manual.html#ref-features-distro any package or set of packages can be defined and exposed as a distro feature. 
Anyway, I am rolling that line back to the previous version.

eu-smirnov: Yeah, I'm also not sure what is the exact difference, but I think connman just never gets into the DISTRO_FEATURES variable.

eu-smirnov:
Looks good! But if I understand correctly, ostree updates won't work in the non-systemd image, because of the missing https://github.com/advancedtelematic/meta-updater/commit/be2128c39322ee885b967fd05d6c7b9c7b3f3bd7

mike-sul:
> Looks fine to me so far, although I haven't tested.
> 
> Does `conf/distro/poky-sota-systemd.conf` need to get updated? It still has `DISTRO = "poky-sota"` which doesn't seem right to me.

It gets rewritten by the DISTRO value in conf/local.conf.sample.append. That's why it worked before and works now. Having said that, I think, it makes sense to update it to `poky-sota-systemd` since poky-sota-systemd.conf is systemd specific/dependent anyway.

conf/distro/poky-sota.conf
```diff
@@ -5,3 +5,5 @@ DISTRO = "poky-sota"
 DISTRO_NAME = "OTA-enabled Linux"
 DISTRO_VERSION = "1.0"
 DISTRO_CODENAME = "sota"
+
+IMAGE_INSTALL_append_sota = " ostree-booted"
```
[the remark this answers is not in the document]
mike-sul: No, we don't need it for the systemd case.

conf/local.nonsystemd.conf.sample.append
```diff
@@ -0,0 +1,19 @@
...
+# General SOTA setup
+#SOTA_CLIENT_PROV = "aktualizr-shared-prov"
+#SOTA_PACKED_CREDENTIALS = "/path/to/credentials.zip"
+
+# Uncomment this line to start an ssh server at boot automatically
+#IMAGE_FEATURES += "ssh-server-dropbear"
+
+# Uncomment this line to set the log level of aktualizr to 'debug' (from 'info'
+# by default)
+#IMAGE_INSTALL_append += " aktualizr-log-debug"
```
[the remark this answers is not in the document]
mike-sul: I can do it, not sure if it worth doing since most of the common part are just comments.

mike-sul:
@eu-smirnov @patrickvacek Is this good enough ? Do you think of something else I can add and/or update here ?

scripts/envsetup.sh
```diff
@@ -3,9 +3,13 @@
 SCRIPT="envsetup.sh"
 MACHINE="$1"
 BUILDDIR="build"
+DISTRO="poky-sota-systemd"
+BASE_CONF="local.conf.base.append"
+declare -A supported_distros=( ["poky-sota-systemd"]="local.conf.systemd.append" ["poky-sota"]="local.conf.base.append" )
 
-[[ "$#" -lt 1 ]] && { echo "Usage: ${SCRIPT} <machine> [builddir]"; return 1; }
+[[ "$#" -lt 1 ]] && { echo "Usage: ${SCRIPT} <machine> [builddir] [distro=< poky-sota-systemd | poky-sota >]"; return 1; }
 [[ "$#" -eq 2 ]] && { BUILDDIR="$2"; }
```
patrickvacek: This `-eq` should actually be `-ge`. Currently if three args are provided, the build dir arg is ignored.

mike-sul: corrected it.

lbonn:
Also looks good for me.

recipes-connectivity/networkd-dhcp-conf/networkd-dhcp-conf.bb
shr-project: This was still wrong, see https://github.com/advancedtelematic/meta-updater/pull/771

PR #626, opened 2019-11-06
meta: change OTA_IMAGE_ROOTFS to TAR_IMAGE_ROOTFS

liuming50:
No functional changes, rename OTA_IMAGE_ROOTFS to TAR_IMAGE_ROOTFS
since the later is a more common name.


liuming50:
Hi, Patrick:

I just thought it's not actually a "OTA" related variable since it would be used in both do_image_ostree and do_image_ota, so better to drop the name prefix "OTA", and we want to override it in one of our meta layers which are not SOTA specific.

PR #627, opened 2019-11-08
Send a copy of repo manifest through garage-push

lbonn:
Depends on ~~https://github.com/advancedtelematic/aktualizr/pull/1438~~
https://github.com/advancedtelematic/aktualizr/pull/1440

### DIFF
--- a/classes/image_types_ostree.bbclass
+++ b/classes/image_types_ostree.bbclass
@@ -188,7 +188,8 @@ do_image_ostreepush[depends] += "aktualizr-native:do_populate_sysroot ca-certifi
 IMAGE_CMD_ostreepush () {
     # send a copy of the repo manifest to backend if available
     local SEND_MANIFEST=""
-    if [ -f ${IMAGE_ROOTFS}${sysconfdir}/manifest.xml ]; then
+    # check if garage-push supports the --repo-manifest option before trying
+    if $(garage-push --help | grep -q '^\s*--repo-manifest') && [ -f ${IMAGE_ROOTFS}${sysconfdir}/manifest.xml ]; then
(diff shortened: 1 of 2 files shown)


lbonn:
One concern I have right now is that this PR is coupled with the aktualizr change: as we're using a recently added cli flag, the garage-push command will fail if used on an older aktualizr version.

This is not a major issue but can be inconvenient nonetheless, I'll try to see if there is a clean workaround for that.

patrickvacek:
> One concern I have right now is that this PR is coupled with the aktualizr change: as we're using a recently added cli flag, the garage-push command will fail if used on an older aktualizr version.

What's the problem as long as we bump the aktualizr version as part of this PR? I think it's fine.

lbonn:
> What's the problem as long as we bump the aktualizr version as part of this PR? I think it's fine.

It's a bit more coupling than necessary and could be annoying if we're trying to bisect some issue on aktualizr for example. But yes, not major.

Still, I just pushed a small change to first check if the option shows up in `--help`. If you think it makes things worse, we can drop it.

classes/image_types_ostree.bbclass
```diff
@@ -188,7 +188,8 @@ do_image_ostreepush[depends] += "aktualizr-native:do_populate_sysroot ca-certifi
 IMAGE_CMD_ostreepush () {
     # send a copy of the repo manifest to backend if available
     local SEND_MANIFEST=""
-    if [ -f ${IMAGE_ROOTFS}${sysconfdir}/manifest.xml ]; then
+    # check if garage-push supports the --repo-manifest option before trying
+    if $(garage-push --help | grep -q '^\s*--repo-manifest') && [ -f ${IMAGE_ROOTFS}${sysconfdir}/manifest.xml ]; then
```
patrickvacek: Ha, nice!

PR #629, opened 2019-11-12
Fix uart output on recent raspberrypi3

lbonn:
The broadcom firmware now needs device trees to be present in/boot to
set it up.


patrickvacek:
It booted, which is definitely progress, but aktualizr is failing with `could not create directory:`. Still looking into it.

lbonn:
The fix for this is here: https://github.com/advancedtelematic/aktualizr/pull/1444.

PR #631, opened 2019-11-18
Add sanity checks for user-defined SOTA variables

Zee314159:
Draft code.

### DIFF
--- a/classes/sota_sanity.bbclass
+++ b/classes/sota_sanity.bbclass
@@ -27,12 +27,16 @@ def sota_check_variables_validity(status, d):
         prov = d.getVar("SOTA_CLIENT_PROV").strip()
         if prov not in ("aktualizr-shared-prov", "aktualizr-device-prov", "aktualizr-device-prov-hsm", ""):
             status.addresult("Valid options for SOTA_CLIENT_PROV are aktualizr-shared-prov, aktualizr-device-prov and aktualizr-device-prov-hsm.\n")
+            if prov == "aktualizr-auto-prov":
+                bb.warn('aktualizr-auto-prov is deprecated. Please use aktualizr-shared-prov instead.')
+            elif prov == "aktualizr-ca-implicit-prov":
+                bb.warn('aktualizr-ca-implicit-prov is deprecated. Please use aktualizr-device-prov instead.')
+            elif prov == "aktualizr-hsm-prov":
+                bb.warn('aktualizr-hsm-prov is deprecated. Please use aktualizr-device-prov-hsm instead.')
     if d.getVar("GARAGE_TARGET_URL") and re.match("^(https?|ftp|file)://.+$", d.getVar("GARAGE_TARGET_URL")) is None:
         status.addresult("GARAGE_TARGET_URL is set to a bad url.\n")
-    if d.getVar("SOTA_POLLING_SEC") and re.match("^(0|\+?[1-9][0-9]*)$", d.getVar("SOTA_POLLING_SEC")) is None:
+    if d.getVar("SOTA_POLLING_SEC") and re.match("^[1-9]\d*|0$", d.getVar("SOTA_POLLING_SEC")) is None:
         status.addresult("SOTA_POLLING_SEC should be an integer.\n")
-    if d.getVar("OSTREE_REPO") and re.match("^\/([a-zA-Z0-9_-]+\/?)+$", d.getVar("OSTREE_REPO")) is None:
-        status.addresult("OSTREE_REPO is not set correctly. Path to your OSTree repository is invalid.\n")
     config = d.getVar("SOTA_SECONDARY_CONFIG")
     if config is not None and config != "":
         path = os.path.abspath(config)
--- a/classes/sota_sanity.bbclass
+++ b/classes/sota_sanity.bbclass
@@ -10,6 +10,32 @@ def sota_check_required_variables(status, d):
         if not d.getVar(var):
             status.addresult("%s should be set in your local.conf.\n" % var)
 
+def sota_check_variables_validity(status, d):
+    var = d.getVar("OSTREE_BRANCHNAME")
+    if var != "":
+        for ch in var:
+            if not (ch >= 'a' and ch <= 'z' or ch >= 'A' and ch <= 'Z' or ch >= '0' and ch <= '9' or ch = '_' or ch == '-'):
+                status.addresult("OSTREE_BRANCHNAME Should only contain characters from the character set [a-zA-Z0-9_-].\n")
+                break
+    var = d.getVar("{SOTA_HARDWARE_ID")
+    if var != "":
+        for ch in var:
+            if not (ch >= 'a' and ch <= 'z' or ch >= 'A' and ch <= 'Z' or ch >= '0' and ch <= '9' or ch = '_' or ch == '-'):
+                status.addresult("SOTA_HARDWARE_ID Should only contain characters from the character set [a-zA-Z0-9_-].\n")
+                break
+    var = d.getVar("SOTA_CLIENT_FEATURES")
+    if var != "hsm" and var != "secondary-network" and var != "":
+        status.addresult("SOTA_CLIENT_FEATURES should be set to hsm or secondary-network.\n")
+    var = d.getVar("OSTREE_UPDATE_SUMMARY")
+    if var != "0" and var != "1" and var != "":
+        status.addresult("OSTREE_UPDATE_SUMMARY should be set to 0 or 1.\n")
+    var = d.getVar("OSTREE_DEPLOY_DEVICETREE")
+    if var != "0" and var != "1" and var != "":
+        status.addresult("OSTREE_DEPLOY_DEVICETREE should be set to 0 or 1.\n")
+    var = GARAGE_SIGN_AUTOVERSION
+    if var != "0" and var != "1" and var != "":
+        status.addresult("GARAGE_SIGN_AUTOVERSION should be set to 0 or 1.\n")
+
 def sota_raise_sanity_error(msg, d):
     if d.getVar("SANITY_USE_EVENTS") == "1":
         bb.event.fire(bb.event.SanityCheckFailed(msg), d)
@@ -52,3 +78,19 @@ python sota_check_sanity_eventhandler() {
 
     return
 }
+
+# Translate old provisioning recipe names into the new versions.
+python () {
+    prov = d.getVar("SOTA_CLIENT_PROV")
+    if prov == "aktualizr-auto-prov":
+        bb.warn('aktualizr-auto-prov is deprecated. Please use aktualizr-shared-prov instead.')
+        d.setVar("SOTA_CLIENT_PROV", "aktualizr-shared-prov")
+    elif prov == "aktualizr-ca-implicit-prov":
+        bb.warn('aktualizr-ca-implicit-prov is deprecated. Please use aktualizr-device-prov instead.')
+        d.setVar("SOTA_CLIENT_PROV", "aktualizr-device-prov")
+    elif prov == "aktualizr-hsm-prov":
+        bb.warn('aktualizr-hsm-prov is deprecated. Please use aktualizr-device-prov-hsm instead.')
+        d.setVar("SOTA_CLIENT_PROV", "aktualizr-device-prov-hsm")
+}
+
+
--- a/classes/sota_sanity.bbclass
+++ b/classes/sota_sanity.bbclass
@@ -11,31 +11,47 @@ def sota_check_required_variables(status, d):
             status.addresult("%s should be set in your local.conf.\n" % var)
 
 def sota_check_variables_validity(status, d):
-    var = d.getVar("OSTREE_BRANCHNAME")
-    if var != "":
-        for ch in var:
-            if not (ch >= 'a' and ch <= 'z' or ch >= 'A' and ch <= 'Z' or ch >= '0' and ch <= '9' or ch = '_' or ch == '-'):
-                status.addresult("OSTREE_BRANCHNAME Should only contain characters from the character set [a-zA-Z0-9_-].\n")
-                break
-    var = d.getVar("{SOTA_HARDWARE_ID")
-    if var != "":
-        for ch in var:
-            if not (ch >= 'a' and ch <= 'z' or ch >= 'A' and ch <= 'Z' or ch >= '0' and ch <= '9' or ch = '_' or ch == '-'):
-                status.addresult("SOTA_HARDWARE_ID Should only contain characters from the character set [a-zA-Z0-9_-].\n")
-                break
-    var = d.getVar("SOTA_CLIENT_FEATURES")
-    if var != "hsm" and var != "secondary-network" and var != "":
-        status.addresult("SOTA_CLIENT_FEATURES should be set to hsm or secondary-network.\n")
-    var = d.getVar("OSTREE_UPDATE_SUMMARY")
-    if var != "0" and var != "1" and var != "":
-        status.addresult("OSTREE_UPDATE_SUMMARY should be set to 0 or 1.\n")
-    var = d.getVar("OSTREE_DEPLOY_DEVICETREE")
-    if var != "0" and var != "1" and var != "":
-        status.addresult("OSTREE_DEPLOY_DEVICETREE should be set to 0 or 1.\n")
-    var = GARAGE_SIGN_AUTOVERSION
-    if var != "0" and var != "1" and var != "":
-        status.addresult("GARAGE_SIGN_AUTOVERSION should be set to 0 or 1.\n")
+    import re
+    import os.path
 
+    if d.getVar("OSTREE_BRANCHNAME") and re.match("^[a-zA-Z0-9_-]*$", d.getVar("OSTREE_BRANCHNAME")) is None:
+        status.addresult("OSTREE_BRANCHNAME Should only contain characters from the character set [a-zA-Z0-9_-].\n")
+    if d.getVar("SOTA_HARDWARE_ID") and re.match("^[a-zA-Z0-9_-]*$", d.getVar("SOTA_HARDWARE_ID")) is None:
+        status.addresult("SOTA_HARDWARE_ID Should only contain characters from the character set [a-zA-Z0-9_-].\n")
+    if d.getVar("SOTA_CLIENT_FEATURES") is not None:
+        for feat in d.getVar("SOTA_CLIENT_FEATURES").split(' '):
+            if feat not in ("hsm", "serialcan", "ubootenv", ""):
+                status.addresult("SOTA_CLIENT_FEATURES should only include hsm, serialcan and bootenv.\n")
+                break
+    if d.getVar("SOTA_CLIENT_PROV") is not None:
+        prov = d.getVar("SOTA_CLIENT_PROV").strip()
+        if prov not in ("aktualizr-shared-prov", "aktualizr-device-prov", "aktualizr-device-prov-hsm", ""):
+            status.addresult("Valid options for SOTA_CLIENT_PROV are aktualizr-shared-prov, aktualizr-device-prov and aktualizr-device-prov-hsm.\n")
+    if d.getVar("GARAGE_TARGET_URL") and re.match("^(https?|ftp|file)://.+$", d.getVar("GARAGE_TARGET_URL")) is None:
+        status.addresult("GARAGE_TARGET_URL is set to a bad url.\n")
+    if d.getVar("SOTA_POLLING_SEC") and re.match("^(0|\+?[1-9][0-9]*)$", d.getVar("SOTA_POLLING_SEC")) is None:
+        status.addresult("SOTA_POLLING_SEC should be an integer.\n")
+    if d.getVar("OSTREE_REPO") and re.match("^\/([a-zA-Z0-9_-]+\/?)+$", d.getVar("OSTREE_REPO")) is None:
+        status.addresult("OSTREE_REPO is not set correctly. Path to your OSTree repository is invalid.\n")
+    config = d.getVar("SOTA_SECONDARY_CONFIG")
+    if config is not None and config != "":
+        path = os.path.abspath(config)
+        if not os.path.exists(path):
+            status.addresult("SOTA_SECONDARY_CONFIG is not set correctly. The file containing JSON configuration for secondaries does not exist.\n")
+    credentials = d.getVar("SOTA_PACKED_CREDENTIALS")
+    if credentials is not None and credentials != "":
+        path = os.path.abspath(credentials)
+        if not os.path.exists(path):
+            status.addresult("SOTA_PACKED_CREDENTIALS is not set correctly. The zipped credentials file does not exist.\n")
+    if d.getVar("OSTREE_UPDATE_SUMMARY") and d.getVar("OSTREE_UPDATE_SUMMARY") not in ("0", "1", ""):
+        status.addresult("OSTREE_UPDATE_SUMMARY should be set to 0 or 1.\n")
+    if d.getVar("OSTREE_DEPLOY_DEVICETREE") and d.getVar("OSTREE_DEPLOY_DEVICETREE") not in ("0", "1", ""):
+        status.addresult("OSTREE_DEPLOY_DEVICETREE should be set to 0 or 1.\n")
+    if d.getVar("GARAGE_SIGN_AUTOVERSION") and d.getVar("GARAGE_SIGN_AUTOVERSION") not in ("0", "1", ""):
+        status.addresult("GARAGE_SIGN_AUTOVERSION should be set to 0 or 1.\n")
+    if d.getVar("SOTA_DEPLOY_CREDENTIALS") and d.getVar("SOTA_DEPLOY_CREDENTIALS") not in ("0", "1", ""):
+        status.addresult("SOTA_DEPLOY_CREDENTIALS should be set to 0 or 1.\n")
+    
 def sota_raise_sanity_error(msg, d):
     if d.getVar("SANITY_USE_EVENTS") == "1":
         bb.event.fire(bb.event.SanityCheckFailed(msg), d)
@@ -60,6 +76,7 @@ def sota_check_sanity(sanity_data):
 
     sota_check_overrides(status, sanity_data)
     sota_check_required_variables(status, sanity_data)
+    sota_check_variables_validity(status, sanity_data)
 
     if status.messages != "":
         sota_raise_sanity_error(sanity_data.expand(status.messages), sanity_data)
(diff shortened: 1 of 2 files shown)


classes/sota_sanity.bbclass
```diff
@@ -10,6 +10,32 @@ def sota_check_required_variables(status, d):
         if not d.getVar(var):
             status.addresult("%s should be set in your local.conf.\n" % var)
 
+def sota_check_variables_validity(status, d):
+    var = d.getVar("OSTREE_BRANCHNAME")
+    if var != "":
```
lbonn: The non empty check is redundant.

More concise with regex: `if re.match("^[a-zA-Z0-9_-]*$", d.getVar("OSTREE_BRANCHNAME")) is not None: ...`

classes/sota_sanity.bbclass
```diff
@@ -10,6 +10,32 @@ def sota_check_required_variables(status, d):
...
+                status.addresult("OSTREE_BRANCHNAME Should only contain characters from the character set [a-zA-Z0-9_-].\n")
+                break
+    var = d.getVar("{SOTA_HARDWARE_ID")
+    if var != "":
+        for ch in var:
+            if not (ch >= 'a' and ch <= 'z' or ch >= 'A' and ch <= 'Z' or ch >= '0' and ch <= '9' or ch = '_' or ch == '-'):
+                status.addresult("SOTA_HARDWARE_ID Should only contain characters from the character set [a-zA-Z0-9_-].\n")
+                break
+    var = d.getVar("SOTA_CLIENT_FEATURES")
+    if var != "hsm" and var != "secondary-network" and var != "":
```
lbonn: more consise: `if d.getVar("SOTA_CLIENT_FEATURES") not in ("hsm", "secondary-network", ""):`

classes/sota_sanity.bbclass
Zee314159: thanks @lbonn , looks way better!

Zee314159:
`SOTA_MAIN_DTB`  `SOTA_EXTRA_CONF_FRAGS` `INITRAMFS_IMAGE` `OSTREE_OSNAME`
`OSTREE_COMMIT_BODY` `OSTREE_COMMIT_SUBJECT`  `SOTA_DT_OVERLAYS` have not been checked yet.
And I'm not sure how to handle case like `RESOURCE_xxx_pn-aktualizr` , when it's  not a definite  variable name.

classes/sota_sanity.bbclass
```diff
@@ -52,3 +86,19 @@ python sota_check_sanity_eventhandler() {
...
+    if prov == "aktualizr-auto-prov":
+        bb.warn('aktualizr-auto-prov is deprecated. Please use aktualizr-shared-prov instead.')
+        d.setVar("SOTA_CLIENT_PROV", "aktualizr-shared-prov")
+    elif prov == "aktualizr-ca-implicit-prov":
+        bb.warn('aktualizr-ca-implicit-prov is deprecated. Please use aktualizr-device-prov instead.')
+        d.setVar("SOTA_CLIENT_PROV", "aktualizr-device-prov")
+    elif prov == "aktualizr-hsm-prov":
+        bb.warn('aktualizr-hsm-prov is deprecated. Please use aktualizr-device-prov-hsm instead.')
+        d.setVar("SOTA_CLIENT_PROV", "aktualizr-device-prov-hsm")
+}
```
patrickvacek: Ah, thanks, this is a better place for this!

classes/sota_sanity.bbclass
```diff
@@ -10,6 +10,40 @@ def sota_check_required_variables(status, d):
...
 
+def sota_check_variables_validity(status, d):
+    if re.match("^[a-zA-Z0-9_-]*$", d.getVar("OSTREE_BRANCHNAME")) is None:
+        status.addresult("OSTREE_BRANCHNAME Should only contain characters from the character set [a-zA-Z0-9_-].\n")
+    if re.match("^[a-zA-Z0-9_-]*$", d.getVar("SOTA_HARDWARE_ID")) is None:
+        status.addresult("SOTA_HARDWARE_ID Should only contain characters from the character set [a-zA-Z0-9_-].\n")
+    if d.getVar("SOTA_CLIENT_FEATURES") not in ("hsm", "secondary-network", ""):
+        status.addresult("SOTA_CLIENT_FEATURES should be set to hsm or secondary-network.\n")
+    if d.getVar("SOTA_CLIENT_PROV") not in ("aktualizr-shared-prov", "aktualizr-device-prov", "aktualizr-device-prov-hsm", ""):
+        status.addresult("Valid options for SOTA_CLIENT_PROV are aktualizr-shared-prov, aktualizr-device-prov, and aktualizr-device-prov-hsm.\n")
```
patrickvacek: Which gets called first, this variable check function or the prov recipe translation python bit below? If this is called first, then this requirement for the provisioning names will prevent the translator from having an effect, as the old names will be rejected.

classes/sota_sanity.bbclass
```diff
@@ -10,6 +10,40 @@ def sota_check_required_variables(status, d):
...
+    if re.match("^(https?|ftp|file)://.+$", d.getVar("GARAGE_TARGET_URL")) is None:
+        status.addresult("GARAGE_TARGET_URL is set to a bad url.\n")
+    if re.match("^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}Z$", d.getVar("GARAGE_TARGET_EXPIRES")) is None:
+        status.addresult("GARAGE_TARGET_EXPIRES expects a valid UTC instant.\n")
+    if re.match("^\dY\dM\dD$", d.getVar("GARAGE_TARGET_EXPIRE_AFTER")) is None:
```
patrickvacek: Will this allow an arbitrary number of years/months/days, or only a single digit for each?

classes/sota_sanity.bbclass
```diff
@@ -10,6 +10,40 @@ def sota_check_required_variables(status, d):
...
+        status.addresult("Specify the number of years, months and days like \"1Y3M5D\" for GARAGE_TARGET_EXPIRE_AFTER.\n")
+    if re.match("^[1-9][0-9]*$", d.getVar("SOTA_POLLING_SEC")) is None:
+        status.addresult("SOTA_POLLING_SEC should be an integer.\n")
+    if re.match("^[1-9][0-9]*$", d.getVar("RESOURCE_xxx_pn-aktualizr")) is None:
```
patrickvacek: I think we can't really check this since `xxx` isn't a valid name. :)

classes/sota_sanity.bbclass
```diff
@@ -10,6 +10,40 @@ def sota_check_required_variables(status, d):
...
+        status.addresult("RESOURCE_xxx_pn-aktualizr should be an integer.\n")
+    if re.match(r"^[a-zA-Z0-9_-\:\/\\]*$", d.getVar("OSTREE_REPO")) is None:
+        status.addresult("OSTREE_REPO path to your OSTree repository is invalid.\n")
+    if re.match(r"^[a-zA-Z0-9_-\:\/\\]*$", d.getVar("SOTA_SECONDARY_CONFIG")) is None:
+        status.addresult("SOTA_SECONDARY_CONFIG path to your OSTree repository is invalid.\n")
+    if re.match(r"^[a-zA-Z0-9_-\:\/\\\.]*$", d.getVar("SOTA_PACKED_CREDENTIALS")) is None:
```
patrickvacek: `SOTA_SECONDARY_CONFIG` and `SOTA_PACKED_CREDENTIALS` can simply be checked if they are non-empty and exist. Seems easier than regexing for proper file name syntax. `OSTREE_REPO` doesn't have to exist before we use it, though, so for that, I suppose the regex is the best we can do.

Zee314159:
> And one other question: does sota_sanity get executed on every recipe or once per image/bitbake?
I don't know actually. I'll look into it.

patrickvacek:
@eu-smirnov You originally made this ticket; was there anything else you think should be covered here?

classes/sota_sanity.bbclass
Zee314159: Yes, so I think I should just simply remove this one?

patrickvacek: Yeah, I would.

classes/sota_sanity.bbclass
```diff
@@ -10,6 +24,40 @@ def sota_check_required_variables(status, d):
...
+    if re.match("^[a-zA-Z0-9_-]*$", d.getVar("SOTA_HARDWARE_ID")) is None:
+        status.addresult("SOTA_HARDWARE_ID Should only contain characters from the character set [a-zA-Z0-9_-].\n")
+    if d.getVar("SOTA_CLIENT_FEATURES") not in ("hsm", "secondary-network", ""):
+        status.addresult("SOTA_CLIENT_FEATURES should be set to hsm or secondary-network.\n")
+    if d.getVar("SOTA_CLIENT_PROV") not in ("aktualizr-shared-prov", "aktualizr-device-prov", "aktualizr-device-prov-hsm", ""):
+        status.addresult("Valid options for SOTA_CLIENT_PROV are aktualizr-shared-prov, aktualizr-device-prov, and aktualizr-device-prov-hsm.\n")
+    if re.match("^(https?|ftp|file)://.+$", d.getVar("GARAGE_TARGET_URL")) is None:
+        status.addresult("GARAGE_TARGET_URL is set to a bad url.\n")
+    if re.match("^^((([0-9]{3}[1-9]|[0-9]{2}[1-9][0-9]{1}|[0-9]{1}[1-9][0-9]{2}|[1-9][0-9]{3})-(((0[13578]|1[02])-(0[1-9]|[12][0-9]|3[01]))|((0[469]|11)-(0[1-9]|[12][0-9]|30))|(02-(0[1-9]|[1][0-9]|2[0-8]))))|((([0-9]{2})(0[48]|[2468][048]|[13579][26])|((0[48]|[2468][048]|[3579][26])00))-02-29))T([0-1]?[0-9]|2[0-3]):([0-5][0-9]):([0-5][0-9])Z$", d.getVar("GARAGE_TARGET_EXPIRES")) is None:
```
Zee314159: This bit was stole from internet, but it's ugly.

patrickvacek: While the intention is good, I think this might be going too far. `garage-sign` should already be verifying if this value is correctly formatted and complaining appropriately. The goal of having these checks up front is to prevent problems that are otherwise hard to detect or not obvious. The perfect example is spaces or "unusual" characters in `OSTREE_BRANCHNAME`, which I think has burned several of us at different times. But these inputs to garage-sign have a much more limited use and AFAIK are already verified appropriately.

classes/sota_sanity.bbclass
```diff
@@ -10,6 +24,40 @@ def sota_check_required_variables(status, d):
...
+    if re.match("^[1-9][0-9]*$", d.getVar("SOTA_POLLING_SEC")) is None:
+        status.addresult("SOTA_POLLING_SEC should be an integer.\n")
+    if re.match(r"^[a-zA-Z0-9_-\:\/\\]*$", d.getVar("OSTREE_REPO")) is None:
+        status.addresult("OSTREE_REPO is not set correctly. Path to your OSTree repository is invalid.\n")
+    path = d.getVar("SOTA_SECONDARY_CONFIG")
+    if not os.path.exists(path):
+        status.addresult("SOTA_SECONDARY_CONFIG is not set correctly.The file containing JSON configuration for secondaries does not exist.\n")
+    path = d.getVar("SOTA_PACKED_CREDENTIALS")
+    if not os.path.exists(path):
+        status.addresult("SOTA_PACKED_CREDENTIALS is not set correctly.The zipped credentials file does not exist.\n")
```
patrickvacek: For both `SOTA_SECONDARY_CONFIG` and `SOTA_PACKED_CREDENTIALS`, leaving them unset is also a valid option. And less critical, but there should be spaces after the periods between the two sentences in the error message.

classes/sota_sanity.bbclass
Zee314159: Ok, I removed those two garage-sign checks.

classes/sota_sanity.bbclass
Zee314159: > leaving them unset
Is setting `SOTA_PACKED_CREDENTIALS` to  "" a valid option?

Zee314159:
I tested all the variables, all looks good except `OSTREE_REPO`. And my code also looks not concise.

classes/sota_sanity.bbclass
patrickvacek: That's worth double-checking. I know that garage-push and garage-sign won't work without credentials, but I think you should be able to bitbake an image without credentials. If you're using device credential provisioning and you don't want to push the image to our backend for some reason, you shouldn't need credentials. So I think empty is valid, even if not recommended.

classes/sota_sanity.bbclass
```diff
@@ -10,6 +10,55 @@ def sota_check_required_variables(status, d):
         if not d.getVar(var):
             status.addresult("%s should be set in your local.conf.\n" % var)
 
+def sota_check_variables_validity(status, d):
+    import re
+    import os.path
+
+# Translate old provisioning recipe names into the new versions.
```
Zee314159: If I put the `Translate old provisioning recipe names` bits here, and test with an old name, the build will fail. The warning message showed the translation should have been executed, once. So, is it that the translation should be executed more than once? Here is the bitbake log:
```
WARNING: aktualizr-auto-prov is deprecated. Please use aktualizr-shared-prov instead.
Time: 0:00:03
Sstate summary: Wanted 2 Found 0 Missed 2 Current 1246 (0% match, 99% complete)
NOTE: Executing Tasks
NOTE: Setscene tasks completed
ERROR: core-image-minimal-1.0-r0 do_rootfs: Could not invoke dnf. Command '/home/liping/meta-updater-repo/master/build/tmp/work/qemux86_64-poky-linux/core-image-minimal/1.0-r0/recipe-sysroot-native/usr/bin/dnf -v --rpmverbosity=info -y -c /home/liping/meta-updater-repo/master/build/tmp/work/qemux86_64-poky-linux/core-image-minimal/1.0-r0/rootfs/etc/dnf/dnf.conf --setopt=reposdir=/home/liping/meta-updater-repo/master/build/tmp/work/qemux86_64-poky-linux/core-image-minimal/1.0-r0/rootfs/etc/yum.repos.d --installroot=/home/liping/meta-updater-repo/master/build/tmp/work/qemux86_64-poky-linux/core-image-minimal/1.0-r0/rootfs --setopt=logdir=/home/liping/meta-updater-repo/master/build/tmp/work/qemux86_64-poky-linux/core-image-minimal/1.0-r0/temp --repofrompath=oe-repo,/home/liping/meta-updater-repo/master/build/tmp/work/qemux86_64-poky-linux/core-image-minimal/1.0-r0/oe-rootfs-repo --nogpgcheck install aktualizr aktualizr-auto-prov aktualizr-resource-control os-release ostree packagegroup-core-boot run-postinsts systemd-journald-persistent virtual/network-configuration' returned 1:
DNF version: 4.2.2
cachedir: /home/liping/meta-updater-repo/master/build/tmp/work/qemux86_64-poky-linux/core-image-minimal/1.0-r0/rootfs/var/cache/dnf
Added oe-repo repo from /home/liping/meta-updater-repo/master/build/tmp/work/qemux86_64-poky-linux/core-image-minimal/1.0-r0/oe-rootfs-repo
repo: using cache for: oe-repo
not found other for: 
not found modules for: 
not found deltainfo for: 
not found updateinfo for: 
oe-repo: using metadata from Wed 20 Nov 2019 11:06:39 AM UTC.
Last metadata expiration check: 0:00:01 ago on Wed 20 Nov 2019 11:06:39 AM UTC.
No module defaults found
No match for argument: aktualizr-auto-prov
Error: Unable to find a match

ERROR: Logfile of failure stored in: /home/liping/meta-updater-repo/master/build/tmp/work/qemux86_64-poky-linux/core-image-minimal/1.0-r0/temp/log.do_rootfs.24967
ERROR: Task (/home/liping/meta-updater-repo/master/poky/meta/recipes-core/images/core-image-minimal.bb:do_rootfs) failed with exit code '1'


```

patrickvacek: I think I see the problem. Unfortunately, the provisioning translation logic might have to be in `sota.bbclass` to apply everywhere that is used.

However, it's been about six months since we introduced the new names, and I wonder if it's time to deprecate the old names entirely. So instead of translating them, we should print a warning and abort. Then we can get rid of that python script, not have to check and translate recipe names everywhere, and only do the check for the old names once here in a sanity function.

Zee314159: I prefer this solution.

eu-smirnov:
@patrickvacek nope, LGTM.

classes/sota_sanity.bbclass
```diff
@@ -10,6 +10,55 @@ def sota_check_required_variables(status, d):
...
+#        d.setVar("SOTA_CLIENT_PROV", "aktualizr-device-prov")
+#    elif prov == "aktualizr-hsm-prov":
+#        bb.warn('aktualizr-hsm-prov is deprecated. Please use aktualizr-device-prov-hsm instead.')
+#        d.setVar("SOTA_CLIENT_PROV", "aktualizr-device-prov-hsm")
+
+    if d.getVar("OSTREE_BRANCHNAME") and re.match("^[a-zA-Z0-9_-]*$", d.getVar("OSTREE_BRANCHNAME")) is None:
+        status.addresult("OSTREE_BRANCHNAME Should only contain characters from the character set [a-zA-Z0-9_-].\n")
+    if d.getVar("SOTA_HARDWARE_ID") and re.match("^[a-zA-Z0-9_-]*$", d.getVar("SOTA_HARDWARE_ID")) is None:
+        status.addresult("SOTA_HARDWARE_ID Should only contain characters from the character set [a-zA-Z0-9_-].\n")
+    if d.getVar("SOTA_CLIENT_FEATURES") and d.getVar("SOTA_CLIENT_FEATURES") not in ("hsm", "secondary-network", ""):
```
patrickvacek: This is cool, but I just realized that it might not work if `SOTA_CLIENT_FEATURES` includes both possible options. We might need to split it into tokens and verify that each token is in the expected list of features.

classes/sota_sanity.bbclass
patrickvacek: Just to be clear, by "this solution" do you mean my proposal to stop translating and just abort instead?

Zee314159: Yes, that’s what I meant.

classes/sota_sanity.bbclass
Zee314159: Not sure if I’ getting this right, you mean SOTA_CLIENT_FEATURES = “ham secondary-network” might not work?

patrickvacek: Yep!

Zee314159: But isn't "hsm secondary-network" or "secondary-network hsm" not in ("hsm", "secondary-network", "")?

Zee314159: I searched `SOTA_CLIENT_FEATURES` on GitHub found `serialcan` `ubootenv` might also be possible options.

patrickvacek: Good catch. I just did some searching as well and realized that `secondary-network` is no longer used. Would you mind making a PR to update the docs? They live in the aktualizr repo now (`docs/ota-client-guide/modules/ROOT/pages/build-configuration.adoc`).

Zee314159: Of course, but how should I describe `serialcan` and `ubootenv`?

classes/sota_sanity.bbclass
Zee314159: Check `SOTA_CLIENT_PROV` in `sota_sanity.bbclass` would cause os-selftest failure, even if I put the translating bits back to `sota.bbclass`. But I can't find where the deprecated name is used?

classes/sota_sanity.bbclass
patrickvacek: Good point. Let me look into it and I'll open a PR.

classes/sota_sanity.bbclass
Zee314159: Looks `SOTA_CLIENT_PROV` may also include more than one option.
Setting `SOTA_CLIENT_PROV`  to an old name in `local.conf` would cause bitbake failure if the translation is in` sota_sanity.bbclass`. It's maybe a lot easier just stop translating and abort. Is there any other concerns?

patrickvacek: > Looks `SOTA_CLIENT_PROV` may also include more than one option.

That isn't valid; it wouldn't make sense.

> It's maybe a lot easier just stop translating and abort.

Yes, I agree, we shouldn't translate anymore. If someone tries to use the old names, we can suggest the new name, but I think that's enough at this point.

Zee314159: `self.append_config('SOTA_CLIENT_PROV = " aktualizr-shared-prov "') `
（From `lib/oeqa/selftest/cases/updater_qemux86_64.py` `lib/oeqa/selftest/cases/updater_minnowboard.py` `lib/oeqa/selftest/cases/updater_raspberrypi.py` ) 
The blank space behind `aktualizr-shared-prov` seems to be the reason to cause oe-selftest failure.

patrickvacek: Wow, good catch. So I guess then you either have to strip the whitespace out or just search if the expected values are in the value (instead of equivalent).

classes/sota_sanity.bbclass
```diff
@@ -10,6 +10,48 @@ def sota_check_required_variables(status, d):
...
+    if d.getVar("SOTA_CLIENT_PROV") is not None:
+        prov = d.getVar("SOTA_CLIENT_PROV").strip()
+        if prov not in ("aktualizr-shared-prov", "aktualizr-device-prov", "aktualizr-device-prov-hsm", ""):
+            status.addresult("Valid options for SOTA_CLIENT_PROV are aktualizr-shared-prov, aktualizr-device-prov and aktualizr-device-prov-hsm.\n")
+    if d.getVar("GARAGE_TARGET_URL") and re.match("^(https?|ftp|file)://.+$", d.getVar("GARAGE_TARGET_URL")) is None:
+        status.addresult("GARAGE_TARGET_URL is set to a bad url.\n")
+    if d.getVar("SOTA_POLLING_SEC") and re.match("^(0|\+?[1-9][0-9]*)$", d.getVar("SOTA_POLLING_SEC")) is None:
+        status.addresult("SOTA_POLLING_SEC should be an integer.\n")
+    if d.getVar("OSTREE_REPO") and re.match("^\/([a-zA-Z0-9_-]+\/?)+$", d.getVar("OSTREE_REPO")) is None:
+        status.addresult("OSTREE_REPO is not set correctly. Path to your OSTree repository is invalid.\n")
```
patrickvacek: This is tricky. It could happen that someone could have a space in a path, for example, although bitbake seems to barf on that. Is there a pythonic way to verify if a path name is POSIX-conformant? Or does Yocto specify anywhere what characters are valid in their variables names and/or paths?

classes/sota_sanity.bbclass
```diff
@@ -10,6 +10,48 @@ def sota_check_required_variables(status, d):
...
+            if feat not in ("hsm", "serialcan", "ubootenv", ""):
+                status.addresult("SOTA_CLIENT_FEATURES should only include hsm, serialcan and bootenv.\n")
+                break
+    if d.getVar("SOTA_CLIENT_PROV") is not None:
+        prov = d.getVar("SOTA_CLIENT_PROV").strip()
+        if prov not in ("aktualizr-shared-prov", "aktualizr-device-prov", "aktualizr-device-prov-hsm", ""):
+            status.addresult("Valid options for SOTA_CLIENT_PROV are aktualizr-shared-prov, aktualizr-device-prov and aktualizr-device-prov-hsm.\n")
+    if d.getVar("GARAGE_TARGET_URL") and re.match("^(https?|ftp|file)://.+$", d.getVar("GARAGE_TARGET_URL")) is None:
+        status.addresult("GARAGE_TARGET_URL is set to a bad url.\n")
+    if d.getVar("SOTA_POLLING_SEC") and re.match("^(0|\+?[1-9][0-9]*)$", d.getVar("SOTA_POLLING_SEC")) is None:
```
patrickvacek: Was this trying to prevent "0"? "0" works fine for me, and I think it's actually valid, albeit questionably desirable. Can we just verify that the value is an integer and nonnegative?

classes/sota_sanity.bbclass
```diff
@@ -10,6 +10,48 @@ def sota_check_required_variables(status, d):
...
+    if d.getVar("SOTA_HARDWARE_ID") and re.match("^[a-zA-Z0-9_-]*$", d.getVar("SOTA_HARDWARE_ID")) is None:
+        status.addresult("SOTA_HARDWARE_ID Should only contain characters from the character set [a-zA-Z0-9_-].\n")
+    if d.getVar("SOTA_CLIENT_FEATURES") is not None:
+        for feat in d.getVar("SOTA_CLIENT_FEATURES").split(' '):
+            if feat not in ("hsm", "serialcan", "ubootenv", ""):
+                status.addresult("SOTA_CLIENT_FEATURES should only include hsm, serialcan and bootenv.\n")
+                break
+    if d.getVar("SOTA_CLIENT_PROV") is not None:
+        prov = d.getVar("SOTA_CLIENT_PROV").strip()
+        if prov not in ("aktualizr-shared-prov", "aktualizr-device-prov", "aktualizr-device-prov-hsm", ""):
```
patrickvacek: I think getting rid of the translation is good, but can we still suggest the new versions if a user tries the old versions? It does mean checking for each of the old ones individually, but I think that's the polite thing to do.

classes/sota_sanity.bbclass
Zee314159: Seems not like easy :(
https://stackoverflow.com/questions/9532499/check-whether-a-path-is-valid-in-python-without-creating-a-file-at-the-paths-ta/9532586#9532586

patrickvacek: I see. Thanks for looking around for a solution, but too bad it isn't easy. I don't think the `os.path.isdir` solution is going to work, though. I think we might be overthinking it, so perhaps for now we should just leave `OSTREE_REPO` alone and not worry about it. It's normally set automatically, anyway, so the risk here is fairly low.

classes/sota_sanity.bbclass
```diff
@@ -10,6 +10,52 @@ def sota_check_required_variables(status, d):
...
+            status.addresult("Valid options for SOTA_CLIENT_PROV are aktualizr-shared-prov, aktualizr-device-prov and aktualizr-device-prov-hsm.\n")
+            if prov == "aktualizr-auto-prov":
+                bb.warn('aktualizr-auto-prov is deprecated. Please use aktualizr-shared-prov instead.')
+            elif prov == "aktualizr-ca-implicit-prov":
+                bb.warn('aktualizr-ca-implicit-prov is deprecated. Please use aktualizr-device-prov instead.')
+            elif prov == "aktualizr-hsm-prov":
+                bb.warn('aktualizr-hsm-prov is deprecated. Please use aktualizr-device-prov-hsm instead.')
+    if d.getVar("GARAGE_TARGET_URL") and re.match("^(https?|ftp|file)://.+$", d.getVar("GARAGE_TARGET_URL")) is None:
+        status.addresult("GARAGE_TARGET_URL is set to a bad url.\n")
+    if d.getVar("SOTA_POLLING_SEC") and re.match("^[1-9]\d*|0$", d.getVar("SOTA_POLLING_SEC")) is None:
```
patrickvacek: Any reason not to use just `\d*` instead of the bits with `[1-9]` and `|0`?

Zee314159: To avoid numbers like `0123`

patrickvacek: Fair enough. I think this is good, then. Thanks again!

PR #635, opened 2019-11-21
Zeus gitlab pipeline definitions

lbonn:


patrickvacek:
Qemu and minnowboard as well?

lbonn:
I noticed that we actually don't have pipelines for thud and warrior either on the qemu and minnowboard repos. I could go ahead and open the 6 PRs though

patrickvacek:
> I noticed that we actually don't have pipelines for thud and warrior either on the qemu and minnowboard repos. I could go ahead and open the 6 PRs though

Do you mean nightly runs in CI, or just to run on new PRs? Mike set up the latter just recently and I backported that to thud and warrior. If the former, shouldn't the existing nightly runs of meta-updater take care of that?

lbonn:
> Do you mean nightly runs in CI, or just to run on new PRs? Mike set up the latter just recently and I backported that to thud and warrior. If the former, shouldn't the existing nightly runs of meta-updater take care of that?

I meant for PRs but nevermind, I was looking at the wrong place... Sorry

PR #637, opened 2019-11-26
build and run image without ostree support for qemux86-64

kbushgit:
This is draft PR for non OSTree image build, currently only for qemyx86-64.
1. Setup environment:
    $ source envsetup.sh qemux86-64 "path-to-yocto-build-folder" poky
2. Build:
    $bitbake core-image-minimal
3. Run QEMU:
    $ ./run-qemu-ota --uboot-enable=no

### DIFF
--- a/conf/local.conf.nonostree.append
+++ b/conf/local.conf.nonostree.append
@@ -1,0 +1,11 @@
+
+DISTRO_FEATURES_append = " systemd"
+VIRTUAL-RUNTIME_init_manager = "systemd"
+
+PREFERRED_RPROVIDER_virtual/network-configuration ??= "networkd-dhcp-conf"
+
+SOTA_DEPLOY_CREDENTIALS ?= "1"
+PACKAGECONFIG_pn-aktualizr = ""
+
+IMAGE_INSTALL_append += "aktualizr"
+IMAGE_INSTALL_append += "aktualizr-shared-prov"
--- a/scripts/envsetup.sh
+++ b/scripts/envsetup.sh
@@ -5,17 +5,24 @@ MACHINE="$1"
 BUILDDIR="build"
 DISTRO="poky-sota-systemd"
 BASE_CONF="local.conf.base.append"
-declare -A supported_distros=( ["poky-sota-systemd"]="local.conf.systemd.append" ["poky-sota"]="local.conf.base.append" )
 
-[[ "$#" -lt 1 ]] && { echo "Usage: ${SCRIPT} <machine> [builddir] [distro=< poky-sota-systemd | poky-sota >]"; return 1; }
+# A definition of a dictionary with a list of configuration files that must be appended
+# to resulting conf/local.conf file for each particular distribution.
+declare -A supported_distros=(
+    ["poky-sota-systemd"]="local.conf.systemd.append"
+    ["poky-sota"]="local.conf.base.append"
+    ["poky"]="local.conf.systemd.append local.conf.nonostree.append"
+)
+
+[[ "$#" -lt 1 ]] && { echo "Usage: ${SCRIPT} <machine> [builddir] [distro=< poky-sota-systemd | poky-sota | poky >]"; return 1; }
 [[ "$#" -ge 2 ]] && { BUILDDIR="$2"; }
 [[ "$#" -eq 3 ]] && { DISTRO="$3"; }
 
 # detect if this script is sourced: see http://stackoverflow.com/a/38128348/6255594
 SOURCED=0
-if [ -n "$ZSH_EVAL_CONTEXT" ]; then
+if [[ -n "$ZSH_EVAL_CONTEXT" ]]; then
   [[ "$ZSH_EVAL_CONTEXT" =~ :file$ ]] && { SOURCED=1; SOURCEDIR=$(cd "$(dirname -- "$0")" && pwd -P); }
-elif [ -n "$BASH_VERSION" ]; then
+elif [[ -n "$BASH_VERSION" ]]; then
   [[ "$0" != "${BASH_SOURCE[0]}" ]] && { SOURCED=1; SOURCEDIR=$(cd "$(dirname -- "${BASH_SOURCE[0]}")" && pwd -P); }
 fi
 
@@ -26,23 +33,25 @@ if [[ $SOURCED -ne 1 ]]; then
 fi
 
 METADIR=${METADIR:-${SOURCEDIR}/../..}
-DISTRO_CONF=${supported_distros[$DISTRO]}
-[[ -n $DISTRO_CONF ]] && { echo "Using $DISTRO_CONF for the specified distro $DISTRO"; } || { echo "The specified distro $DISTRO is not supported"; return 1; }
 
 if [[ ! -f "${BUILDDIR}/conf/local.conf" ]]; then
+  declare -a DISTRO_CONFIGS=${supported_distros[$DISTRO]}
+  [[ -n ${DISTRO_CONFIGS[@]} ]] && { echo "Using (${DISTRO_CONFIGS[*]}) for the specified distro '$DISTRO'"; } || { echo "The specified distro $DISTRO is not supported"; return 1; }
+
   source "$METADIR/poky/oe-init-build-env" "$BUILDDIR"
 
   echo "METADIR  := \"\${@os.path.abspath('${METADIR}')}\"" >> conf/bblayers.conf
   cat "${METADIR}/meta-updater/conf/include/bblayers/sota.inc" >> conf/bblayers.conf
   cat "${METADIR}/meta-updater/conf/include/bblayers/sota_${MACHINE}.inc" >> conf/bblayers.conf
-
   sed -e "s/##MACHINE##/$MACHINE/g" \
       -e "s/##DISTRO##/$DISTRO/g" \
-	 "${METADIR}/meta-updater/conf/$BASE_CONF" >> conf/local.conf
+      "${METADIR}/meta-updater/conf/$BASE_CONF" >> conf/local.conf
 
-  if [ "$BASE_CONF" != "$DISTRO_CONF" ]; then
-    cat "${METADIR}/meta-updater/conf/$DISTRO_CONF" >> conf/local.conf
-  fi
+  for config in ${DISTRO_CONFIGS[@]}; do
+    if [[ "$BASE_CONF" != "$config" ]]; then
+      cat "${METADIR}/meta-updater/conf/$config" >> conf/local.conf
+    fi
+  done
 else
   source "$METADIR/poky/oe-init-build-env" "$BUILDDIR"
 fi
--- a/scripts/envsetup.sh
+++ b/scripts/envsetup.sh
@@ -45,7 +45,7 @@ if [[ ! -f "${BUILDDIR}/conf/local.conf" ]]; then
   cat "${METADIR}/meta-updater/conf/include/bblayers/sota_${MACHINE}.inc" >> conf/bblayers.conf
   sed -e "s/##MACHINE##/$MACHINE/g" \
       -e "s/##DISTRO##/$DISTRO/g" \
-      "${METADIR}/meta-updater/conf/$BASE_CONF" >> conf/local.conf
+         "${METADIR}/meta-updater/conf/$BASE_CONF" >> conf/local.conf
 
   for config in ${DISTRO_CONFIGS[@]}; do
     if [[ "$BASE_CONF" != "$config" ]]; then
--- a/scripts/qemucommand.py
+++ b/scripts/qemucommand.py
@@ -40,9 +40,25 @@ def random_mac():
 
 class QemuCommand(object):
     def __init__(self, args):
+        print(args)
+        self.enable_u_boot = None
         self.dry_run = args.dry_run
         self.overlay = args.overlay
         self.host_fwd = None
+        self.kernel = None
+        self.drive_interface = "ide"
+
+        if hasattr(args, 'uboot_enable'):
+            self.enable_u_boot = args.uboot_enable.lower() in ("yes", "true", "1")
+
+        # Rise an exception if U-Boot is disabled and overlay option is used
+        if not self.enable_u_boot and self.overlay:
+            raise EnvironmentError("An overlay option is currently supported only with U-Boot loader!")
+
+        # If booting with u-boot is disabled we use "ext4" root fs instead of custom one "ota-ext4"
+        if not self.enable_u_boot:
+            self.drive_interface = "virtio"
+            EXTENSIONS['qemux86-64'] = 'ext4'
 
         if args.machine:
             self.machine = args.machine
@@ -59,9 +75,7 @@ class QemuCommand(object):
         # overlay so that we can keep it around just in case.
         if args.efi:
             self.bios = 'OVMF.fd'
-        else:
-            if args.bootloader:
-                uboot_path = args.bootloader
+        elif self.enable_u_boot:
             uboot_path = abspath(join(args.dir, self.machine, 'u-boot-qemux86-64.rom'))
             if self.overlay:
                 new_uboot_path = self.overlay + '.u-boot.rom'
@@ -77,6 +91,8 @@ class QemuCommand(object):
             if not exists(uboot_path) and not (self.dry_run and not exists(self.overlay)):
                 raise ValueError("U-Boot image %s does not exist" % uboot_path)
             self.bios = uboot_path
+        else:
+            self.kernel = abspath(join(args.dir, self.machine, 'bzImage-qemux86-64.bin'))
 
         # If using an overlay, we need to keep the "backing" image around, as
         # bitbake will often clean it up, and the overlay silently depends on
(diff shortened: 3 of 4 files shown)


mike-sul:
>     1. Setup environment:
>        $ source envsetup.sh qemux86-64  poky-sota-nonostree
envsetup.sh treats the second param as a build directory name so a correct call should be
`source envsetup.sh qemux86-64 build poky-sota-nonostree`

conf/distro/poky-sota-nonostree.conf
```diff
@@ -0,0 +1,19 @@
+require conf/distro/poky.conf
+
```
mike-sul: It looks like almost the same as conf/distro/poky-sota-systemd.conf  but I see that
`require conf/distro/sota.conf.inc` is not present here, was it removed for purpose ?

conf/distro/poky-sota-nonostree.conf
```diff
@@ -0,0 +1,19 @@
...
+DISTRO = "poky-sota-nonostree"
+DISTRO_NAME = "OTA-enabled Linux"
+DISTRO_VERSION = "1.0"
+DISTRO_CODENAME = "sota-non-ostree"
+
+DISTRO_FEATURES_append = " systemd"
+VIRTUAL-RUNTIME_init_manager = "systemd"
+PREFERRED_RPROVIDER_virtual/network-configuration ??= "networkd-dhcp-conf"
+
+SOTA_HARDWARE_ID ??= "${MACHINE}-non-ostree"
```
mike-sul: I am not sure that the non-ostree configuration requires another distro configuration as it actually does not use/add any 'distribution' as Yocto defines (it https://www.yoctoproject.org/docs/current/mega-manual/mega-manual.html#ref-features-distro).
Therefore, perhaps, it makes sense just simply move the nonostree specific stuff to the conf/local.conf.nonostree.append.

conf/distro/poky-sota-nonostree.conf
kbushgit: I'm just not sure that it has to inherit from "sota", but I will try! The main idea is that we want to skip to use any ostree related things

conf/distro/poky-sota-nonostree.conf
kbushgit: The reason for such a decision is that the required configurations are placed in the distro config, and keep the local.conf clean, as much as possible.

scripts/envsetup.sh
```diff
@@ -5,9 +5,9 @@ MACHINE="$1"
 BUILDDIR="build"
 DISTRO="poky-sota-systemd"
 BASE_CONF="local.conf.base.append"
-declare -A supported_distros=( ["poky-sota-systemd"]="local.conf.systemd.append" ["poky-sota"]="local.conf.base.append" )
+declare -A supported_distros=( ["poky-sota-systemd"]="local.conf.systemd.append" ["poky-sota"]="local.conf.base.append" ["poky-sota-nonostree"]="local.conf.nonostree.append")
```
lbonn: From the list poky-sota-systemd, poky-sota and poky-sota-nonostree, one would expect that poky-sota-nonostree does not use systemd.

In the current state, I'm not sure why we actually need a distro and not just add the systemd specific features to local.conf.

If we do want to have such distro, I agree with @mike-sul: maybe we should also use some of the options from sota.conf.inc, especially for reproducible builds. It's always a nice property and it should help sharing cached work with ostree builds.

conf/distro/poky-sota-nonostree.conf
```diff
@@ -0,0 +1,19 @@
...
+
+PACKAGECONFIG_pn-aktualizr = ""
+
+IMAGE_INSTALL_append += "aktualizr"
+IMAGE_INSTALL_append += "aktualizr-shared-prov-creds"
+IMAGE_INSTALL_append += "aktualizr-shared-prov"
+IMAGE_INSTALL_append += "aktualizr-hwid"
```
patrickvacek: `aktualizr-hwid` is included automatically and the provisioning recipes need to be configurable (and they include the `-creds` recipes automatically if appropriate). Hence you could make these four lines a one-liner: `IMAGE_INSTALL_append += " aktualizr ${SOTA_CLIENT_PROV} "

conf/local.conf.nonostree.append
```diff
@@ -0,0 +1,19 @@
...
+#
+# Uncomment these lines to change the default parameters.
+#
+#RESOURCE_CPU_WEIGHT_pn-aktualizr = "100"
+#RESOURCE_MEMORY_HIGH_pn-aktualizr = "100M"
+#RESOURCE_MEMORY_MAX_pn-aktualizr = "80%"
+IMAGE_INSTALL_append += " aktualizr-resource-control"
+
+# The default sota hardware id. 
+#SOTA_HARDWARE_ID ?= "${MACHINE}-non-ostree"
```
patrickvacek: Is there a need to suggest changing the `SOTA_HARDWARE_ID` here? The only config appends don't do it and I don't see why we need it here. In fact, I see no problem with using the base and systemd appends as they are. Or am I overlooking something?

scripts/run-qemu-ota
```diff
@@ -13,6 +13,9 @@ def main():
     parser = ArgumentParser(description='Run meta-updater image in qemu')
     parser.add_argument('imagename', default='core-image-minimal', nargs='?',
                         help="Either the name of the bitbake image target, or a path to the image to run")
+    parser.add_argument('--uboot', default='no',
```
patrickvacek: The default here should be "yes" so as to not break backwards compatibility. Also, it appears to me that this impacts much more than just u-boot, so maybe the flag should actually reference OSTree. Maybe better to name it `--ostree` and again default it to "yes". Or just use `--nonostree` as a flag.

conf/distro/poky-sota-nonostree.conf
patrickvacek: I think I'm still leaning towards just dumping this in local.conf. I don't think this belongs as an independent distro.

scripts/qemucommand.py
```diff
@@ -129,10 +144,15 @@ def command_line(self):
             netuser += ',hostfwd=tcp:0.0.0.0:2159-:2159'
         cmdline = [
             "qemu-system-x86_64",
-            "-bios", self.bios
         ]
+        if self.enable_u_boot:
+            cmdline += ["-bios", self.bios]
+        else:
+            cmdline += ["-device", "virtio-net-pci"]
+            cmdline += ["-kernel", self.kernel]
```
patrickvacek: Interesting that we have to supply the kernel. Why do we need to here if we don't have to in the OSTree/u-boot case?

scripts/qemucommand.py
```diff
@@ -166,6 +186,11 @@ def command_line(self):
             cmdline += ['-cpu', 'Haswell']
         if self.overlay:
             cmdline.append(self.overlay)
+
+        # If booting with u-boot is disabled, add kernel command line arguments through qemu -append option
+        if not self.enable_u_boot:
+            cmdline += ["-append", "root=/dev/vda rw highres=off mem=256M console=ttyS0 "
+                                   "ip=one oprofile.timer=1 uvesafb.task_timeout=-1"]
```
patrickvacek: Why hardcore the `mem` here? I think we can safely rely on the existing mechanism for that. And what is the rest doing? All of `ip=one oprofile.timer=1 uvesafb.task_timeout=-1` looks mysterious to me.

KostiantynBushko: **ip=one** is a  mistake, must be **ip=dhcp**! Without this option, the network interface is not started.
This is a reference to kernel command-line arguments 
https://www.kernel.org/doc/html/v4.14/admin-guide/kernel-parameters.html
which for this option redirect to this doc 
https://www.kernel.org/doc/Documentation/filesystems/nfs/nfsroot.txt

**"oprofile.timer"** and **"uvesafb.task_timeout=-1"** are introduced here
https://git.yoctoproject.org/cgit.cgi/poky/tree/meta/conf/machine/include/qemuboot-x86.inc
and actually both these arguments can be removed

**oprofile.timer=1** is related to this tool https://linux.die.net/man/1/oprofile
the man page says:
`OProfile makes use of the hardware performance counters provided on Intel, AMD, and other 
processors, and uses a timer-interrupt based mechanism on CPUs without counters.`
**It can be removed!**

**uvesafb** - is related to the frame buffer driver for the console which is not necessary for our case an also can be removed

**mem** argument, you're right, this value is set by default.

scripts/qemucommand.py
KostiantynBushko: When we use u-boot, we specify it using -bios option in QEMU, so (u-boot.rom) is interpreted by QEMU as a kind of BIOS, and is responsible to load the kernel. In the case of QEMU, this is just a path to the kernel binary file i.e. 
./tmp/deploy/image/qemux86-64/bzImage
A relative path to the kernel image is specified in the patch to u-boot
https://github.com/advancedtelematic/meta-updater-qemux86-64/blob/b1d134cbb5a47e5008db4a65510b4c102f636b4d/recipes-bsp/u-boot/files/0001-Set-up-environment-for-OSTree-integration.patch#L41
To check it out we can simple remove\rename this file
./tmp/deploy/images/qemux86-64/bzImage
which is just a link to the actual kernel binary file, and the run-qemu-ota won't work because u-boot won't be able to find kernel image.

scripts/run-qemu-ota
patrickvacek: Also worth pointing out: I've been intending to add a parameter to this script to allow supplying a custom u-boot rom, and the nicest thing to do would be to called it `--uboot`. I might try to make a PR today. It would be another argument in favor of renaming this option.

patrickvacek: See https://github.com/advancedtelematic/meta-updater/pull/638 for my draft.

scripts/envsetup.sh
kbushgit: Another reason to use distro is to keep a uniform approach to generate local.conf in the `envsetup.sh`,  introduced by @mike-sul 

For me, it also looks better to add some exceptional case in` envsetup.sh` and remove unnecessary distro(poky-sota-nonostree).

Just for reminder, I still want to use "poky" distro without inheriting `poky-sota`

conf/distro/poky-sota-nonostree.conf
kbushgit: This variable is not set since I do not use `sota` distro and sota.bbclass is not applied in this case

scripts/run-qemu-ota
kbushgit: Yes, I see, I think this is two separate things, one option allows as to specify u-boot.rom file and second allow as to disable booting using u-boot at all. 
So my proposal is to have two options,  and both of them with the default values:
1. `--uboot=None` (allows to specify custom  ROM file)
2. `--uboot-enable=yes` (disable\enable,  ignoring any other option related to u-boot)

`--ostree` or `--nonostree` is ok but those names just do not represent what exactly we do, so my vote for (`--uboot, --uboot-enable`)

conf/distro/poky-sota-nonostree.conf
patrickvacek: I see your point. There might be a smart way to do this, but I suppose for now, specifying the prov recipe is okay. But `-hwid` and `-creds` are unnecessary; they will be included automatically by other recipes.

scripts/qemucommand.py
```diff
@@ -40,8 +40,22 @@ def random_mac():
...
+        print(args)
+        self.enable_u_boot = args.uboot_enable.lower() in ("yes", "true", "1")
         self.dry_run = args.dry_run
         self.overlay = args.overlay
+        self.kernel = None
+        self.drive_interface = "ide"
+
+        # Rise an exception if U-Boot is disabled and overlay option is used
+        if not self.enable_u_boot and self.overlay:
+            raise EnvironmentError("An overlay option is currently supported only with U-Boot loader!")
```
patrickvacek: Why? I don't think that is true.

conf/local.conf.nonostree.append
```diff
@@ -0,0 +1,28 @@
...
+#RESOURCE_MEMORY_HIGH_pn-aktualizr = "100M"
+#RESOURCE_MEMORY_MAX_pn-aktualizr = "80%"
+
+IMAGE_INSTALL_append += " aktualizr-resource-control"
+
+PACKAGECONFIG_pn-aktualizr = ""
+
+IMAGE_INSTALL_append += "aktualizr"
+IMAGE_INSTALL_append += "aktualizr-shared-prov-creds"
+IMAGE_INSTALL_append += "aktualizr-shared-prov"
```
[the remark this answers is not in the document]
kbushgit: Yes, You are right this recipe is added in aktualizr-shared-prov,  now I see my mistake, we have to set  SOTA_DEPLOY_CREDENTIALS, this var is specified in sota.class which is not used in this case. Thanks!

conf/local.conf.nonostree.append
```diff
@@ -0,0 +1,28 @@
...
+
+# Set resource limits for aktualizr service
+#
+# Uncomment these lines to change the default parameters.
+#
+#RESOURCE_CPU_WEIGHT_pn-aktualizr = "100"
+#RESOURCE_MEMORY_HIGH_pn-aktualizr = "100M"
+#RESOURCE_MEMORY_MAX_pn-aktualizr = "80%"
+
+IMAGE_INSTALL_append += " aktualizr-resource-control"
```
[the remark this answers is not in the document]
kbushgit: yes, of course, we can reuse local.conf.systemd.append but I think that it can add more confusion.
The  local.conf.systemd.append is intended to be used in the context of sota.bbclass, and actually, "conf.append" does not configure system-d.
I think it's not a good idea just to reuse "conf.append", because the different configurations files may be intended to different bb classes, does it make sense?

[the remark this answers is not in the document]
patrickvacek: But the point here is that these `.append` files should be transparent to the user. The `envsetup` script does the work of automatically appending them to `local.conf`, so I'm not sure who would get confused by the end result.

> The local.conf.systemd.append is intended to be used in the context of sota.bbclass

I disagree; it's just some logging and resource limiting that isn't specific to `sota.bbclass`.

> actually, "conf.append" does not configure system-d

True, but it's configuration that should only be used with systemd, which we are assuming applies in this case as well. If someone doesn't want it, they can remove it from their `local.conf` along with everything else systemd-specific.

scripts/envsetup.sh
```diff
@@ -26,23 +33,28 @@ if [[ $SOURCED -ne 1 ]]; then
...
+[[ -n ${DISTRO_CONFIGS[@]} ]] && { echo "Using (${DISTRO_CONFIGS[*]}) for the specified distro '$DISTRO'"; } || { echo "The specified distro $DISTRO is not supported"; return 1; }
 
 if [[ ! -f "${BUILDDIR}/conf/local.conf" ]]; then
   source "$METADIR/poky/oe-init-build-env" "$BUILDDIR"
 
   echo "METADIR  := \"\${@os.path.abspath('${METADIR}')}\"" >> conf/bblayers.conf
   cat "${METADIR}/meta-updater/conf/include/bblayers/sota.inc" >> conf/bblayers.conf
   cat "${METADIR}/meta-updater/conf/include/bblayers/sota_${MACHINE}.inc" >> conf/bblayers.conf
-
+
```
patrickvacek: Nit: unnecessary extra space.

scripts/envsetup.sh
```diff
@@ -26,23 +33,28 @@ if [[ $SOURCED -ne 1 ]]; then
...
 
   echo "METADIR  := \"\${@os.path.abspath('${METADIR}')}\"" >> conf/bblayers.conf
   cat "${METADIR}/meta-updater/conf/include/bblayers/sota.inc" >> conf/bblayers.conf
   cat "${METADIR}/meta-updater/conf/include/bblayers/sota_${MACHINE}.inc" >> conf/bblayers.conf
-
+ 
   sed -e "s/##MACHINE##/$MACHINE/g" \
       -e "s/##DISTRO##/$DISTRO/g" \
-	 "${METADIR}/meta-updater/conf/$BASE_CONF" >> conf/local.conf
+         "${METADIR}/meta-updater/conf/$BASE_CONF" >> conf/local.conf
```
patrickvacek: Nit: inconsistent indenting.

patrickvacek:
>  $ source envsetup.sh qemux86-64 poky

BTW, this is missing the build directory parameter. It should be something like `source envsetup.sh qemux86-64 build poky`

patrickvacek:
I restarted the CI because it was failing on something unrelated. Now, however, it appears to be failing on something relevant. Can you take a look into that?

Otherwise, I've tested this and it looks pretty good to me. Thanks! And now I see what you mean about the disk space error causing a loop. I'll look into that and/or will make a ticket. Have you managed to update a device successfully using these non-ostree images?

patrickvacek:
> Have you managed to update a device successfully using these non-ostree images?

I just did it. Cool! (And yes, I know the "installation" is basically fake here. Still!)

patrickvacek:
CI is still failing. Also, another thought: every time I re-run `envsetup.sh`, I now see something like `Using (local.conf.systemd.append local.conf.nonostree.append) for the specified distro 'poky'`, despite that after the config has been initialized once, those files are not re-appended. Can you silence that message after the first time?

scripts/run-qemu-ota
```diff
@@ -88,4 +91,4 @@ def main():
 
 
 if __name__ == '__main__':
-    main()
+    main()
```
eu-smirnov: No newline

patrickvacek:
I merged the reversion of the CA cert issue. If you use the latest aktualizr on this branch, does it work?

PR #638, opened 2019-11-29
Add --uboot flag to specify path to a custom u-boot rom.

patrickvacek:
This is intended to help with keeping around older versions of images with a similarly old bootloader.

@Raigi I'm hoping this is what you *actually* need for keeping around old images. Using this should give you the same features as https://github.com/advancedtelematic/meta-updater/pull/599 but without having to mess with clearing credentials. To use this, you'll need to copy the ota-ext4 image and the u-boot rom in your deploy directory and then launch qemu with something like `../meta-updater/scripts/run-qemu-ota --no-gui --machine qemux86-64 --uboot ./u-boot-qemux86-64.rom ./core-image-minimal-qemux86-64.ota-ext4`. What do you think?

### DIFF
--- a/lib/oeqa/selftest/cases/testutils.py
+++ b/lib/oeqa/selftest/cases/testutils.py
@@ -33,6 +33,7 @@ def qemu_boot_image(imagename, **kwargs):
     # subdirectory.
     args.dir = 'tmp/deploy/images'
     args.efi = kwargs.get('efi', False)
+    args.bootloader = kwargs.get('bootloader', None)
     args.machine = kwargs.get('machine', None)
     args.mem = kwargs.get('mem', '128M')
     qemu_use_kvm = get_bb_var("QEMU_USE_KVM")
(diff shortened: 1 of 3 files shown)


kbushgit:
Theoretically, we can have more than one command-line options related to u-boot, I think we have to define more specific name for this variable, like:
`--u-boot-file` or` --uboot-rom`

patrickvacek:
> Theoretically, we can have more than one command-line options related to u-boot, I think we have to define more specific name for this variable, like:
> `--u-boot-file` or` --uboot-rom`

Fair point. Since this actually isn't even specific to u-boot, maybe it should be just `--bootloader` or `--boot-rom`?

kbushgit:
> > Theoretically, we can have more than one command-line options related to u-boot, I think we have to define more specific name for this variable, like:
> > `--u-boot-file` or` --uboot-rom`
> 
> Fair point. Since this actually isn't even specific to u-boot, maybe it should be just `--bootloader` or `--boot-rom`?

--bootloader looks better for me, we just have to provide a good description(help) for this option

patrickvacek:
> --bootloader looks better for me, we just have to provide a good description(help) for this option

Done. Look better now?

PR #640, opened 2019-12-03
Trigger otf bitbake tests on PR

Raigi:
Not completely sure, how you create PRs here.. Let me know if this is correct, then I will create also for other branches and projects

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -2,6 +2,7 @@ stages:
   - docker
   - checkout
   - test
+  - trigger
 
 variables:
   BITBAKE_IMAGE: ${CI_REGISTRY_IMAGE}:ci-master-bitbake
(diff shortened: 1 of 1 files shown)


patrickvacek:
Normally we add new work to master and then backport, but this is probably fine for now. Seems okay to me and appears to be working. @lbonn do you have any further thoughts here?

Raigi:
> Normally we add new work to master and then backport, but this is probably fine for now. Seems okay to me and appears to be working. @lbonn do you have any further thoughts here?

Not sure if we want to run otf on master PRs?

patrickvacek:
> Not sure if we want to run otf on master PRs?

Up to you. I can understand why you'd prefer to only run it on release branch (backport) PRs. Maybe we only do that for now and reevaluate later if we want more.

Raigi:
> > Not sure if we want to run otf on master PRs?
> 
> Up to you. I can understand why you'd prefer to only run it on release branch (backport) PRs. Maybe we only do that for now and reevaluate later if we want more.

Sounds good to me.

patrickvacek:
Uh-oh, the trigger failed.

Raigi:
> Uh-oh, the trigger failed.

yes, it is trying to do smth with docker, which otf build machine doesn't need. @lbonn any ideas?

Raigi:
The pipeline got triggered now :)

PR #641, opened 2019-12-04
Force a simple reproducible build mechanism

lbonn:
In 25a17f0346ab05b3212bb280e4a8b622bcf99bce, poky sets
reproducible_build by defaults. However, it is quite complex as it tries
to compute a plausible timestamp from source files, in a reproducible
ways.

Setting the timestamps to 0, just like ostree, is probably safer in our
case, so let's switch to our old behavior.

Tagging @agners who originally proposed this enhancement.

### DIFF
--- a/conf/distro/sota.conf.inc
+++ b/conf/distro/sota.conf.inc
@@ -11,9 +11,10 @@ INHERIT += " sota"
 USER_CLASSES_remove = "image-prelink"
 
 # Enable reproducible builds. Use 0 as mtime, the same as OSTree is using.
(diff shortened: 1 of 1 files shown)


patrickvacek:
Took a really long time to run CI for this, which I think is expected. What's a good way to test this? Run a second pipeline on the branch and check if the caching works?

lbonn:
@patrickvacek yes, I've just restarted the qemu build and oe-selftest.

lbonn:
Ok, it seems that everything works fine.

PR #642, opened 2019-12-04
Set PKCS11_ENGINE_PATH when using aktualizr with HSM

lbonn:
Depends on https://github.com/advancedtelematic/aktualizr/pull/1471.

### DIFF
--- a/recipes-sota/aktualizr/aktualizr-device-prov-hsm.bb
+++ b/recipes-sota/aktualizr/aktualizr-device-prov-hsm.bb
@@ -14,7 +14,7 @@ DEPENDS = "aktualizr"
 # If the config file from aktualizr used here is changed, you will need to bump
 # the version here because of SIGGEN_EXCLUDE_SAFE_RECIPE_DEPS!
 PV = "1.0"
-PR = "6"
+PR = "7"
 
 SRC_URI = ""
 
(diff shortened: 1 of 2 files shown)


patrickvacek:
This looks fine, but is there a way to make it configurable if a user wants to use a different HSM or pkcs11 provider?

lbonn:
@patrickvacek I think it's more about system configuration than user choice as it should be fixed when using libp11 (which aktualizr assumes).

That said, it would probably need to be changed when using older versions of openssl so I'll add it as a variable in the recipe so it can be easily overriden.

recipes-sota/aktualizr/aktualizr-device-prov-hsm.bb
```diff
@@ -14,7 +14,7 @@ DEPENDS = "aktualizr"
 # If the config file from aktualizr used here is changed, you will need to bump
 # the version here because of SIGGEN_EXCLUDE_SAFE_RECIPE_DEPS!
 PV = "1.0"
-PR = "6"
```
lbonn: Got bitten by this...

patrickvacek: Argh, sorry. That is annoying.

PR #647, opened 2019-12-12
Thud backport for 2019.11

lbonn:


Raigi:
OTF tests passed.
Creating a raspberrypi3 device and updating it passed.

PR #651, opened 2019-12-13
Don't trigger otf on nightly meta-updater schedules

Raigi:
Nightly schedules are interfering with oft pipeline nightly schedules, so I would like to disable meta-updater pipelines from triggering otf on schedules.


Raigi:
@lbonn @patrickvacek Can you please look at this one as well?

PR #655, opened 2019-12-31
meta: add python to HOSTTOOLS_NONFATAL

liuming50:
repo is requiring python to be able to work correctly, add python to
HOSTTOOLS_NONFATAL as well, meanwhile, combine HOSTTOOLS_NONFATAL
definitions in sota.bbclass.

image_repo_manifest is actually a image bbclass, move it to
IMAGE_CLASSES instead of inheriting it in sota.bbclass.

### DIFF
--- a/classes/sota.bbclass
+++ b/classes/sota.bbclass
@@ -8,7 +8,7 @@ SOTA_DEPLOY_CREDENTIALS ?= "1"
 SOTA_HARDWARE_ID ??= "${MACHINE}"
 
 IMAGE_INSTALL_append_sota = " ostree os-release ${SOTA_CLIENT} ${SOTA_CLIENT_PROV}"
-IMAGE_CLASSES += " image_types_ostree image_types_ota"
+IMAGE_CLASSES += " image_types_ostree image_types_ota image_repo_manifest"
 
 IMAGE_FSTYPES += "${@bb.utils.contains('DISTRO_FEATURES', 'sota', 'ostreepush garagesign garagecheck ota-ext4 wic', ' ', d)}"
 IMAGE_FSTYPES += "${@bb.utils.contains('BUILD_OSTREE_TARBALL', '1', 'ostree.tar.bz2', ' ', d)}"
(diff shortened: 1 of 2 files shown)


classes/sota.bbclass
```diff
@@ -1,14 +1,14 @@
...
+HOSTTOOLS_NONFATAL += "java repo python"
 
 SOTA_CLIENT ??= "aktualizr"
 SOTA_CLIENT_PROV ??= "aktualizr-shared-prov"
 SOTA_DEPLOY_CREDENTIALS ?= "1"
 SOTA_HARDWARE_ID ??= "${MACHINE}"
 
 IMAGE_INSTALL_append_sota = " ostree os-release ${SOTA_CLIENT} ${SOTA_CLIENT_PROV}"
-IMAGE_CLASSES += " image_types_ostree image_types_ota"
+IMAGE_CLASSES += " image_types_ostree image_types_ota image_repo_manifest"
```
patrickvacek: Thanks for the submission! What is the benefit of adding to `IMAGE_CLASSES` instead of inheriting? I couldn't find guidance in the docs.

And why is it better to add to `HOSTTOOLS_NONFATAL` here instead of in `image_repo_manifest.bbclass`?

In general, looks like it can't hurt, but if we do go through with this, we need to fix the comment in `image_repo_manifest.bbclass` that mentions inheriting.

liuming50: Hi, Patrick:

Thanks for reviewing my patch.

Adding it to IMAGE_CLASSES would only impact image recipes, but inheriting it in sota.bbclass would impact all recipes, like it will end up with adding a IMAGE_PREPROCESS_COMMAND command for aktualizr, ostree..., which makes no sense.

Putting HOSTTOOLS_NONFATAL into image_repo_manifest.bbclass was not a problem, but usually HOSTTOOLS_NONFATAL is a global variable (set in bitbake.conf, base.bbclass in OE), and we already have a HOSTTOOLS_NONFATAL setting in sota.bbclass, so my intention was to combine them to avoiding being redundant.

patrickvacek: > Adding it to IMAGE_CLASSES would only impact image recipes, but inheriting it in sota.bbclass would impact all recipes, like it will end up with adding a IMAGE_PREPROCESS_COMMAND command for aktualizr, ostree..., which makes no sense

Okay, that makes sense to me now. Thanks for clearing that up!

> Putting HOSTTOOLS_NONFATAL into image_repo_manifest.bbclass was not a problem, but usually HOSTTOOLS_NONFATAL is a global variable (set in bitbake.conf, base.bbclass in OE), and we already have a HOSTTOOLS_NONFATAL setting in sota.bbclass, so my intention was to combine them to avoiding being redundant.

This part I'm less sure of. If anyone were to reuse `image_repo_manifest.bbclass`, it would then be less obvious that `repo` and `python` are (optional) dependencies.

liuming50:
@patrickvacek Thanks for the review, have pushed a V2 as you suggested.

PR #656, opened 2020-01-03
OTA-4174: Make an ostree update a default on IP Secondary

mike-sul:

### DIFF
--- a/recipes-test/demo-config/secondary-config.bb
+++ b/recipes-test/demo-config/secondary-config.bb
@@ -12,14 +12,14 @@ SOTA_HARDWARE_ID ?= "${MACHINE}-sndry"
 SECONDARY_HARDWARE_ID ?= "${SOTA_HARDWARE_ID}"
 
 SRC_URI = "\
-    file://30-fake-pacman.toml \
+    file://30-ostree-pacman.toml \
     file://35-network-config.toml \
     file://45-id-config.toml \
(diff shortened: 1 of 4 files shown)


recipes-test/demo-config/secondary-config.bb
```diff
@@ -12,14 +12,14 @@ SOTA_HARDWARE_ID ?= "${MACHINE}-sndry"
 SECONDARY_HARDWARE_ID ?= "${SOTA_HARDWARE_ID}"
 
 SRC_URI = "\
-    file://30-fake-pacman.toml \
+    file://30-ostree-pacman.toml \
```
patrickvacek: What will happen to `30-fake-pacman.toml`? Will it be unused?

mike-sul: I don't know, it depends on our decision on the file/binary update configuration.

patrickvacek: We're only testing ostree in CI, and if there is no demo or test for the fake package manager on the secondary, I would prefer to remove the unused file.

mike-sul: ok

patrickvacek:
Looks fine now, but of course depends on https://github.com/advancedtelematic/aktualizr/pull/1500.

PR #657, opened 2020-01-08
Allow dots in ostree branch names and hw ids

lbonn:


lbonn:
Oops, the regexp was invalid because `[_-.]` is interpreted at a range, I've moved the `.` back and it should be good...

lbonn:
The CI errors are related, one is transient and the other a new change on poky master we need to accommodate in meta-updater-raspberrypi.

PR #659, opened 2020-01-08
Adapt to shared libraries changes in aktualizr

lbonn:
Depends on https://github.com/advancedtelematic/aktualizr/pull/1512

PR #660, opened 2020-01-09
Adapt to shared libraries changes in aktualizr (bis)

lbonn:
Respin of #659, still depends on https://github.com/advancedtelematic/aktualizr/pull/1512


eu-smirnov:
LGTM

PR #661, opened 2020-01-13
QEMU run script option: allowing to pass qemu <hostfwd> as an optiona…

kbushgit:
…l argument.

Added a new QEMU command line argument <--host-forward> which allows pass arguments for QEMU option <hostfwd> an an optional arguments. 
This option is currently necessary only for running the Zabbix agent on QEMU. 


patrickvacek:
I'm also seeing the ` AttributeError: '' object has no attribute 'host_forward'` error here now, so either this is the same bug or something changed upstream.

PR #663, opened 2020-01-20
Feat/zeus/2020.1

patrickvacek:


.gitlab-ci.yml
```diff
@@ -43,10 +43,6 @@ Build core-image-minimal:
   variables:
     TEST_BUILD_DIR: 'build-core-image-minimal'
     BITBAKE_TARGETS: 'core-image-minimal'
-  artifacts:
```
lbonn: This could have been the other way around: adding these artifacts to other branches. But I don't think anybody is really using that right now (and it's a little bit tricky as the image needs to be provisioned externally), so it should be fine to remove it for the sake of uniformity.

patrickvacek: Yeah, I don't even know how to get/use those myself. Could be interesting, though, but if it becomes necessary later we can always re-add it later.

patrickvacek:
>     1. This change is breaking otf test which is installing a binary update on IP secondary
>        How should I change my test? Do I have to manually change Pacman on the device or can you introduce another configuration variable for the secondary bitbake process?

Sorry about that. I think the right thing to do here is to have two tests now: one for binary updates on IP Secondaries and one for OSTree updates. We changed the default to OSTree, as you've noticed, so we probably need to make a convenient way to switch to binary. @mike-sul does that make sense to you? I think you basically proposed as much and unfortunately I didn't prioritize that.

In the meantime, you can indeed manually change the pacman configuration on the device to get your test working.

>     2\. I was not able to connect Raspberrypi secondary to the network

That is annoying. Have you had any luck with the switch?

mike-sul:
> We changed the default to OSTree, as you've noticed, so we probably need to make a convenient way to switch to binary. @mike-sul does that make sense to you? I think you basically proposed as much and unfortunately I didn't prioritize that.

Yes, it does, should I create a ticket for it ?

mike-sul:
> > ```
> > 2\. I was not able to connect Raspberrypi secondary to the network
> > ```
> 
> 
> That is annoying. Have you had any luck with the switch?

I actually think that it might be possible to improve the secondary networking configuration recipe to cover this case. Effectively, the current recipe does its work based on an assumption/requirement that Secondary just need a connection to Primary and does not care much about having connection to Internet/Treehub itself as it, to some extend, contradicts to the Uptane spec.

patrickvacek:
> > we probably need to make a convenient way to switch to binary.
> 
> Yes, it does, should I create a ticket for it ?

Yes please. Thanks!

> I actually think that it might be possible to improve the secondary networking configuration recipe to cover this case. Effectively, the current recipe does its work based on an assumption/requirement that Secondary just need a connection to Primary and does not care much about having connection to Internet/Treehub itself as it, to some extend, contradicts to the Uptane spec.

I'm not following. I mean, ideally, the Secondary really only should be talking to the Primary. It's only in the OSTree case that we cheat where there is an exception. Is that what you mean by contradicting the Uptane spec?

mike-sul:
> > > we probably need to make a convenient way to switch to binary.
> > 
> > 
> > Yes, it does, should I create a ticket for it ?
> 
> Yes please. Thanks!

https://saeljira.it.here.com/browse/OTA-4388



> > I actually think that it might be possible to improve the secondary networking configuration recipe to cover this case. Effectively, the current recipe does its work based on an assumption/requirement that Secondary just need a connection to Primary and does not care much about having connection to Internet/Treehub itself as it, to some extend, contradicts to the Uptane spec.
> 
> I'm not following. I mean, ideally, the Secondary really only should be talking to the Primary. It's only in the OSTree case that we cheat where there is an exception. Is that what you mean by contradicting the Uptane spec?

Yes, and our Secondary networking configuration recipe is designed just for one use-case/requirement - to be able to communicate with Primary. So, if we would like to support the exception, i.e. the ostree use-case, then we might wanna to revisit the recipe to make sure that Secondary can communicate to both Primary and Treehub (which is effectively means connection to Internet)

Raigi:
> That is annoying. Have you had any luck with the switch?

Tried with the switch and also trying to connect the raspberries into our lab network, no luck.

patrickvacek:
@Raigi I think Mike has a fix for you in https://github.com/advancedtelematic/meta-updater/pull/668. At this point, we already know we need to do another release and bump aktualizr versions again (as is done in his PR), so I'm not sure either of the issues you're having here should block this anymore. Can we perhaps merge this round of backports and then start working on getting the next one into shape?

Raigi:
> @Raigi I think Mike has a fix for you in #668. At this point, we already know we need to do another release and bump aktualizr versions again (as is done in his PR), so I'm not sure either of the issues you're having here should block this anymore. Can we perhaps merge this round of backports and then start working on getting the next one into shape?

ok, let's do so. I'm gonna disable secondary binary updating for the moment, so they won't be constantly failing after merge.

PR #669, opened 2020-01-26
Fix some problems in image recipes

liuming50:


lbonn:
The changes look good thanks! Should be mergeable when the tests will pass.

conf/distro/sota.conf.inc
```diff
@@ -17,4 +17,5 @@ INHERIT += "reproducible_build_simple"
 export SOURCE_DATE_EPOCH = "0"
 REPRODUCIBLE_TIMESTAMP_ROOTFS = "0"
 
-HOSTTOOLS_append = " sync sha256sum"
+HOSTTOOLS += "git sync repo python sha256sum"
```
lbonn: `repo` should be in `HOSTTOOLS_NONFATAL`. About `python`, I don't know why we still have it here, we can probably remove it.

liuming50: @lbonn thanks for the reviewing, yes, you are right about repo should be in HOSTTOOLS_NONFATAL, I have pushed a V2 to fix that.

As for python, it also should be in HOSTTOOLS_NONFATAL per my test, for instance, in my repo binary, it has "#!/usr/bin/env python" at the first line, which will try to locate python in PATH, but in a bitbake environment, only the binaries in HOSTTOOLS_DIR could be found.

patrickvacek:
CI looks good. Thanks for the submission!

PR #671, opened 2020-01-29
Feat/thud/2020.2

mike-sul:


mike-sul:
Ideally it should point to the aktualizr version that includes commits of the following PRs

https://github.com/advancedtelematic/aktualizr/pull/1542  (update changlelog and version in the docs
https://github.com/advancedtelematic/aktualizr/pull/1545 (update docs)

But it's not critical from functional standpoint of view

mike-sul:
@Raigi Would it be possible for you to verify this PR in general an specifically if an ostree update works on Secondary?

Raigi:
> @Raigi Would it be possible for you to verify this PR in general an specifically if an ostree update works on Secondary?

yes, is it fine to check ostree update on secondary with qemu?

mike-sul:
> > @Raigi Would it be possible for you to verify this PR in general an specifically if an ostree update works on Secondary?
> 
> yes, is it fine to check ostree update on secondary with qemu?

I think, it would be safer to verify it on RPi as it is a bit "closer" to the customer environment.

Raigi:
> > > @Raigi Would it be possible for you to verify this PR in general an specifically if an ostree update works on Secondary?
> > 
> > 
> > yes, is it fine to check ostree update on secondary with qemu?
> 
> I think, it would be safer to verify it on RPi as it is a bit "closer" to the customer environment.

Is the dynamic IP fix included for RPi, since I was not able to connect secondary to network with this internet we have here. I will try to create the devices, but I think I will have the same connection problem on secondary.

mike-sul:
> > > > @Raigi Would it be possible for you to verify this PR in general an specifically if an ostree update works on Secondary?
> > > 
> > > 
> > > yes, is it fine to check ostree update on secondary with qemu?
> > 
> > 
> > I think, it would be safer to verify it on RPi as it is a bit "closer" to the customer environment.
> 
> Is the dynamic IP fix included for RPi, since I was not able to connect secondary to network with this internet we have here. I will try to create the devices, but I think I will have the same connection problem on secondary.

unfortunately, no, let's just check on qemu then

PR #675, opened 2020-01-31
Feat/ota 4310/non ostree/oe tests

kbushgit:
Porting oe-selftest cases for non-ostree image

### DIFF
--- a/lib/oeqa/selftest/cases/updater_qemux86_64.py
+++ b/lib/oeqa/selftest/cases/updater_qemux86_64.py
@@ -112,6 +112,32 @@ class SharedCredProvTests(OESelftestTestCase):
         verifyProvisioned(self, machine, hwid)
 
 
+class SharedCredProvTestsNonOSTree(SharedCredProvTests):
+
+    def setUpLocal(self):
+        layer = "meta-updater-qemux86-64"
+        result = runCmd('bitbake-layers show-layers')
+        if re.search(layer, result.output) is None:
+            self.meta_qemu = metadir() + layer
+            runCmd('bitbake-layers add-layer "%s"' % self.meta_qemu)
+        else:
+            self.meta_qemu = None
+        self.append_config('MACHINE = "qemux86-64"')
+        self.append_config('SOTA_CLIENT_PROV = ""')
+        self.append_config('IMAGE_FSTYPES_remove = "ostreepush garagesign garagecheck"')
+        self.append_config('SOTA_HARDWARE_ID = "plain_reibekuchen_314"')
+
+        self.append_config('DISTRO = "poky"')
+        self.append_config('DISTRO_FEATURES_append = " systemd"')
+        self.append_config('VIRTUAL-RUNTIME_init_manager = "systemd"')
+        self.append_config('PREFERRED_RPROVIDER_virtual/network-configuration ??= "networkd-dhcp-conf"')
+        self.append_config('PACKAGECONFIG_pn-aktualizr = ""')
+        self.append_config('SOTA_DEPLOY_CREDENTIALS = "1"')
+        self.append_config('IMAGE_INSTALL_append += "aktualizr"')
+        self.append_config('IMAGE_INSTALL_append += " aktualizr-shared-prov"')
+        self.qemu, self.s = qemu_launch(machine='qemux86-64', uboot_enable='no')
+
+
 class ManualControlTests(OESelftestTestCase):
 
     def setUpLocal(self):
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 __pycache__
 *.pyc
+.idea/
--- a/lib/oeqa/selftest/cases/testutils.py
+++ b/lib/oeqa/selftest/cases/testutils.py
@@ -49,6 +49,7 @@ def qemu_boot_image(imagename, **kwargs):
     args.overlay = kwargs.get('overlay', None)
     args.dry_run = kwargs.get('dry_run', False)
     args.secondary_network = kwargs.get('secondary_network', False)
+    args.uboot_enable = kwargs.get('uboot_enable', 'yes')
 
     qemu = QemuCommand(args)
     cmdline = qemu.command_line()


lib/oeqa/selftest/cases/updater_qemux86_64.py
```diff
@@ -86,7 +89,7 @@ def setUpLocal(self):
         self.append_config('SOTA_CLIENT_PROV = " aktualizr-shared-prov "')
         self.append_config('IMAGE_FSTYPES_remove = "ostreepush garagesign garagecheck"')
         self.append_config('SOTA_HARDWARE_ID = "plain_reibekuchen_314"')
-        self.qemu, self.s = qemu_launch(machine='qemux86-64')
+        self.qemu, self.s = qemu_launch(machine='qemux86-64', uboot_enable=UBOOT_ENABLE)
```
mike-sul: Maybe it can be a default param, so no need to specify it in each call.

kbushgit: Actually the default parameter is ```yes``` which is ok for ostree cases, for the non-ostree, in order to make those tests reusable we have to explicitly set it to ```no```

lib/oeqa/selftest/cases/updater_native_non_ostree.py
```diff
@@ -0,0 +1,13 @@
...
+class SotaToolsTestsNonOSTree(SotaToolsTests):
+    pass
+
+
+class GeneralTestsNonOSTree(GeneralTests):
+    def test_feature_sota(self):
+        pass
+
+    def test_feature_usrmerge(self):
+        pass
```
patrickvacek: Are you intending to fill these in? I'm not quite sure what would be meaningful here.

lib/oeqa/selftest/cases/updater_qemux86_64.py
patrickvacek: Do we need to specify these at all if the default is `yes`?

lib/oeqa/selftest/cases/updater_qemux86_64_non_ostree.py
```diff
@@ -0,0 +1,47 @@
...
+        self.append_config('IMAGE_INSTALL_append += " aktualizr-device-prov-hsm"')
+        super().setUpLocal()
+
+
+class IpSecondaryTestsNonOSTree(IpSecondaryTests):
+    pass
+
+
+class ResourceControlTestsNonOSTree(ResourceControlTests):
+    pass
```
patrickvacek: I'm pretty confused by the goal here. I don't think you need to make a copy of every oe-selftest for the non-ostree case. One test case that sets up the environment, builds an image, and boots it should be enough.

lib/oeqa/selftest/cases/updater_native_non_ostree.py
kbushgit: No, These are tests that we want to avoid in the non-ostere, since we inherit from GeneralTests class these two cases are not relevant, so we override them with an empty body. This looks ugly but it's how we can reuse existing test cases for the non-ostree.

patrickvacek: > it's how we can reuse existing test cases for the non-ostree.

As we discussed yesterday, I don't think reusing test cases makes sense. The tests should be independent so that we can run them regardless of the current environment setup.

kbushgit: That's actually what we do,  with the current approach we can run the tests for both cases independently, we just reuse some common code and test cases from existing classes just to avoid duplication code (test cases) which is general for both cases `ostrere` and `non-ostree`.
So we can now separately run, for example, `updater_qemux86_64` and `updater_qemux86_64_non_ostree`, but not together for the same setup.
I'm just not sure that I understand you correctly, what do you mean when you say, - The tests should be independent..., they are independent!

lib/oeqa/selftest/cases/updater_qemux86_64_non_ostree.py
kbushgit: If we want to reuse existing code(test cases) which is common, and run tests independently for `ostree` and `non-ostree `we have to do this, I didn't look into OESelftestTestCase implementation but  I think it works in the way like looking all classes which inherit OESelftestTestCase and runs all member functions with the prefix test_* so when we run `oe-selftest `we pass a python module(file) and if we want reuse tests from another python module(file) we have to do something like that
https://github.com/advancedtelematic/meta-updater/blob/fed844e7b6b0ffdf52346aaf23ed95689018c401/lib/oeqa/selftest/cases/updater_qemux86_64_non_ostree.py#L10

`oe-selftest` doesn't look into imported modules): to make them visible for `oe-selftes` we do this hack

lib/oeqa/selftest/cases/updater_native_non_ostree.py
patrickvacek: > what do you mean when you say, - The tests should be independent..., they are independent!

Yes, I see that this term is confusing! I mean they should be independent of your current build environment. You (ideally) should haven't to preconfigure anything to run the tests.

> we can now separately run, for example, `updater_qemux86_64` and `updater_qemux86_64_non_ostree`, but not together for the same setup.

That indicates that the tests are *not* independent of the current environment setup.

> we just reuse some common code and test cases from existing classes just to avoid duplication code (test cases) which is general for both cases `ostrere` and `non-ostree`.

We should not need separate test cases for the OSTree and non-OSTree cases. We don't need to re-test everything for the non-OSTree case. All we need to add is a single new test that reconfigures the environment to build and boot and non-OSTree image and then reset everything back to how it was before. The amount of reconfiguration is a bit more than what we've done for other tests, but each of the tests already does some amount of this already.

kbushgit: Ok, I think I catch an idea, previously I've considered performing those tests for the `non-ostree` env setup, so I'm going to rework, using one env and extend existing test files with the cases for `non-ostree`.
One more question, what do you mean, we do not need to re-test everything, actually, we haven't some special cases for `non-ostree`, anyway, we have to retest some of them, not all, cases for 
`non-sotree`, like a test of provisioning, etc.

patrickvacek: > what do you mean, we do not need to re-test everything, actually, we haven't some special cases for `non-ostree`, anyway, we have to retest some of them, not all, cases for `non-sotree`, like a test of provisioning, etc.

I don't think it's necessary to retest all of the variations. I think just covering the basic (shared credential) provisioning case is fine. We only need a single test to prove that we can still build and boot an image without OSTree.

lib/oeqa/selftest/cases/updater_qemux86_64.py
```diff
@@ -204,6 +207,26 @@ def test_provisioning(self):
...
+        global UBOOT_ENABLE
+        UBOOT_ENABLE = 'no'
+        self.append_config('DISTRO = "poky"')
+        self.append_config('DISTRO_FEATURES_append = " systemd"')
+        self.append_config('VIRTUAL-RUNTIME_init_manager = "systemd"')
+        self.append_config('PREFERRED_RPROVIDER_virtual/network-configuration ??= "networkd-dhcp-conf"')
+        self.append_config('PACKAGECONFIG_pn-aktualizr = ""')
+        self.append_config('IMAGE_INSTALL_append += "aktualizr"')
+        self.append_config('IMAGE_INSTALL_remove += " aktualizr-shared-prov"')
+        self.append_config('IMAGE_INSTALL_append += " aktualizr-device-prov"')
```
patrickvacek: For the ease of testing, it might be better to use shared credential provisioning here.

But either way, I like the idea of reusing one of the existing tests and extending it like this.

lib/oeqa/selftest/cases/updater_qemux86_64.py
```diff
@@ -204,6 +207,26 @@ def test_provisioning(self):
         verifyProvisioned(self, machine)
 
 
+class DeviceCredProvTestsNonOSTree(DeviceCredProvTests):
+    def setUpLocal(self):
+        global UBOOT_ENABLE
+        UBOOT_ENABLE = 'no'
```
patrickvacek: Do we have to treat this as a global?

patrickvacek:
Looks like CI keeps failing. I'm seeing things like this:
```
warning: GnuPG was not available during last "repo init"
warning: Cannot automatically authenticate repo."""
```
and
```
 error.GitError: repo var: 
*** Please tell me who you are.
Run
  git config --global user.email "you@example.com"
  git config --global user.name "Your Name"
to set your account's default identity.
Omit --global to set the identity only in this repository.
fatal: unable to auto-detect email address (got 'root@runner-PpvsN-sj-project-497-concurrent-0.(none)')
```
These seem like things that should've come up before. What changed?

lbonn:
> These seem like things that should've come up before. What changed?

Sounds like `repo` released a new version that requires that.

patrickvacek:
@lbonn thanks for the quick fix! @kbushgit you'll want to rebase now that https://github.com/advancedtelematic/meta-updater/pull/683 is merged.

patrickvacek:
Looks like CI is doing better, but my two prior comments still stand.

lib/oeqa/selftest/cases/updater_qemux86_64.py
kbushgit: Yes, in another case we will get a warning like ```Shadows name ...```

lib/oeqa/selftest/cases/updater_qemux86_64.py
kbushgit: So would you prefer to have both of those tests or only shared provisioning?

lib/oeqa/selftest/cases/updater_qemux86_64.py
patrickvacek: I was wondering if it could be passed as an argument like we do for other qemu configuration options.

lib/oeqa/selftest/cases/updater_qemux86_64.py
patrickvacek: Just shared credential provisioning is enough, I think.

lib/oeqa/selftest/cases/updater_qemux86_64.py
kbushgit: By default, this argument is set to ```yes```, only in case we want to run ```qemu``` without  ```u-boot``` we change it, and of course, at the end of the test we just restore it to default state, that's actually the whole idea. We have to have since we inherit from the test which is originally designed for the ```ostree``` case.

patrickvacek: Could we make a member variable of the `SharedCredProvTests`, overwrite it in this test's `setUpLocal` before calling the parent class, and then pass it to `qemu_launch` in the parent class function?

I'm concerned about the global variable having unintended side effects in the future.

kbushgit: I just realized that python unit tests are not designed to be used with inheritance mechanism, I'm trying to define class member var to use it across class instances, so where is the proper place to initialize class self var? 
The idea is that I want to initialize ```self.VAR``` before invoking ```setUpLocal(self)``` member func, the reason why I want to do that is that I inherit my test class from the existing one and I want to have ```self.VAR``` which I'll be able to initialize before invoking ```setUpLocal(self)``` from the parent class.

I can't use ```__init__ ```, so when I try to use ```setUp(self)``` like
```
setUp(self):
  self.uboot_enable = 'no'
```
...all others self.VAR's declared in the parent class become invisible in the subclass?

patrickvacek: > I just realized that python unit tests are not designed to be used with inheritance mechanism

Oh, strange. What's different about python unit tests where inheritance doesn't work? I do seem to recall that this `setUpLocal` function is basically the class initializer, but I seem to also recall from the Yocto docs that there was some other `setUpXyz` function that might be helpful here.

If necessary, you could try setting the variable in the subclass, and then in the superclass, check if it is empty and set it to the real "default" if it is.

kbushgit: The issue was that I call `super().setUpLocal` I did reworking over that, could you check out the latest changes?

kbushgit: I wonder how it affects but if I override` setUpLocal()` or `setUp` all `self.VAR's `from the parent class are not available in the derived class...

patrickvacek: >  I did reworking over that, could you check out the latest changes?

Yep, looks fine to me now! If CI passes, I think this is ready to merge.

PR #676, opened 2020-02-05
OTA-4388: Conf variable to specify an update type for secondary

mike-sul:

### DIFF
--- a/recipes-test/demo-config/files/30-ostree-pacman.toml
+++ b/recipes-test/demo-config/files/30-ostree-pacman.toml
@@ -1,2 +1,0 @@
-[pacman]
-type = "ostree"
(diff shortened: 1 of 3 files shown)


mike-sul:
> Looks fine, but we should probably update the doc in aktualizr about local.conf variables used by meta-updater.

In this PR https://github.com/advancedtelematic/aktualizr/pull/1550

mike-sul:
> And what options do we have for testing this? Do we have an oe-selftest that we can expand to cover both cases?

The oe-selftests that we have for IP Secondary test just device provisioning correctly and includes Secondary, but it does not test any update on Secondary.
Or do you mean checking specifically the configuration part, e.g. see if the expected value is in the toml?

mike-sul:
@Raigi you might find this feature interesting for your OTF tests (can choose update type for secondary)

Raigi:
> @Raigi you might find this feature interesting for your OTF tests (can choose update type for secondary)

yes, thank you! :)

PR #687, opened 2020-02-19
ota_image: pull-local by ref name

eu-smirnov:
This makes ref name visible in the target system.

Signed-off-by: Eugene Smirnov <evgenii.smirnov@here.com>

### DIFF
--- a/classes/image_types_ota.bbclass
+++ b/classes/image_types_ota.bbclass
@@ -63,7 +63,7 @@ IMAGE_CMD_ota () {
 
 	ostree_target_hash=$(cat ${OSTREE_REPO}/refs/heads/${OSTREE_BRANCHNAME}-${IMAGE_BASENAME})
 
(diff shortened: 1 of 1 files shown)


classes/image_types_ota.bbclass
```diff
@@ -63,7 +63,7 @@ IMAGE_CMD_ota () {
 
 	ostree_target_hash=$(cat ${OSTREE_REPO}/refs/heads/${OSTREE_BRANCHNAME}-${IMAGE_BASENAME})
 
-	ostree --repo=${OTA_SYSROOT}/ostree/repo pull-local --remote=${OSTREE_OSNAME} ${OSTREE_REPO} ${ostree_target_hash}
+	ostree --repo=${OTA_SYSROOT}/ostree/repo pull-local --remote=${OSTREE_OSNAME} ${OSTREE_REPO} "${OSTREE_BRANCHNAME}-${IMAGE_BASENAME}"
```
patrickvacek: It might be worth leaving a comment why this is preferred over `ostree_target_hash`, because the difference is non-obvious from merely looking at the code, and we use `ostree_target_hash` below for other things.

eu-smirnov:
Superseded by #716

PR #688, opened 2020-02-21
Fix uart output on raspberrypi3

leon-anavi:
Cherry-pick @lbonn commit to branch thud that provides device trees in /boot for recent versions of broadcom firmware. This change is required to fix the serial output on AGL image for Raspberry Pi 3 with enabled agl-sota.

Best regards,
Leon

### DIFF
--- a/classes/sota_raspberrypi.bbclass
+++ b/classes/sota_raspberrypi.bbclass
@@ -44,9 +44,9 @@ IMAGE_BOOT_FILES_sota = "bcm2835-bootfiles/* \
                          "
 
 # OSTree puts its own boot.scr to bcm2835-bootfiles
-# raspberrypi4 needs dtb in /boot partition so that they can be read by the
-# firmware
-IMAGE_BOOT_FILES_append_sota_raspberrypi4 = "${@make_dtb_boot_files(d)}"
+# rpi4 and recent rpi3 firmwares needs dtb in /boot partition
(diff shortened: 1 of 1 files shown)


lbonn:
In this case, it will have impact on thud images and create more files than necessary with old firmwares in `/boot`. But that's probably something we can live with.

However, have you looked into a way to achieve the same goal on some AGL layer? It might be tricky because it's in `sota_raspberrypi.bbclass`.

If this is still our best shot, I would then rewrite the comment a bit to clarify that the issue does not appear with the version of the firmware shipped with meta-raspberrypi.

leon-anavi:
Hi @lbonn,

Yes, if you prefer I am kind of confident I can achieve the same directly in AGL. I will give it a try tomorrow and provide feedback.

Thanks,
Leon

patrickvacek:
> Yes, if you prefer I am kind of confident I can achieve the same directly in AGL. I will give it a try tomorrow and provide feedback.

Hey @leon-anavi, did you have luck handling this from within AGL?

leon-anavi:
Hi @patrickvacek,

Yes, sorry I forgot to provide feedback here. [The fix has been merged in meta-agl](https://gerrit.automotivelinux.org/gerrit/c/AGL/meta-agl/+/24141). Thank you for the provided help with this issue.

Thanks,
Leon

PR #693, opened 2020-02-28
Feat/thud/2020.3

patrickvacek:


Raigi:
OTF passed, successfully created and updated rpi3 primary and secondary.

Version reported on the device seems wrong though: 
2020.2-106-g9f2cc5d7
SRCREV = "9f2cc5d7b026eb1a1404f0c77ef6cb9377245423"

patrickvacek:
> Version reported on the device seems wrong though:
> 2020.2-106-g9f2cc5d7
> SRCREV = "9f2cc5d7b026eb1a1404f0c77ef6cb9377245423"

Strange, just tried to reproduce and I got the expected result:
```
root@qemux86-64:/sysroot/home/root# aktualizr --version
Current aktualizr version is: 2020.3-0-g9f2cc5d7
```

Hard to say what happened, but since the commit was correct, it's not a big deal. There were some problems making the release, so possibly when you bitbaked, it didn't recognize that there was a new release yet.

Raigi:
> > Version reported on the device seems wrong though:
> > 2020.2-106-g9f2cc5d7
> > SRCREV = "9f2cc5d7b026eb1a1404f0c77ef6cb9377245423"
> 
> Strange, just tried to reproduce and I got the expected result:
> 
> ```
> root@qemux86-64:/sysroot/home/root# aktualizr --version
> Current aktualizr version is: 2020.3-0-g9f2cc5d7
> ```
> 
> Hard to say what happened, but since the commit was correct, it's not a big deal. There were some problems making the release, so possibly when you bitbaked, it didn't recognize that there was a new release yet.

Ok, that might be it then.

PR #700, opened 2020-03-24
Feat/zeus/2020.4

patrickvacek:


Raigi:
There was one failing test in OTF, I have checked it manually.
When installing updates on a Primary that has a Secondary it now takes a lot more time.
Previously it took OTF maximum of 2.5 minutes (normally it took about 1 minute) to get update complete, now it took 4 minutes.

Do you want to improve this or should I adjust waiting times in OTF?

patrickvacek:
>  When installing updates on a Primary that has a Secondary it now takes a lot more time.
> Previously it took OTF maximum of 2.5 minutes (normally it took about 1 minute) to get update complete, now it took 4 minutes.
> Do you want to improve this or should I adjust waiting times in OTF?

Sorry, that's really annoying. I'm confused, though, for the last release/backport, we tried to change the connection logic specifically to improve the likelihood that OTF would pass. I'm not quite sure what changed between then and now to make it worse in a different way. @mike-sul do you have insight here?

At some point, though, I'm going to say we have to give up on bandaids for this problem and we should just try to solve it the "right" way as we've been discussing, meaning a proper connection manager.

Raigi:
> Sorry, that's really annoying. I'm confused, though, for the last release/backport, we tried to change the connection logic specifically to improve the likelihood that OTF would pass. I'm not quite sure what changed between then and now to make it worse in a different way. @mike-sul do you have insight here?

Last time the problem was provisioning time, now it is installing the update time. It seems to me that it takes longer for the device to come back up after rebooting.

mike-sul:
> It seems to me that it takes longer for the device to come back up after rebooting.

Interesting, actually the timing should be improved in case of Primary update since Laurent added this `if (proviisioned)` statement which means that Primary won't try to connect to Secondaries after reboot if a device is provisioned. @Raigi Is/Are Secondary(ties) rebooted too in your case?

Raigi:
> > It seems to me that it takes longer for the device to come back up after rebooting.
> 
> Interesting, actually the timing should be improved in case of Primary update since Laurent added this `if (proviisioned)` statement which means that Primary won't try to connect to Secondaries after reboot if a device is provisioned. @Raigi Is/Are Secondary(ties) rebooted too in your case?

No, only primary got rebooted.
If I run journalctl -fu aktualizr on primary, it takes a bit time, before I get any logs. Sometimes it is much longer even.
`
-- Logs begin at Thu 2020-03-26 10:02:36 UTC. --
Mar 26 10:02:55 qemux86-64 aktualizr[229]: Connecting to and getting info about IP Secondary: 10.0.3.2:9050...
`

mike-sul:
> > > It seems to me that it takes longer for the device to come back up after rebooting.
> > 
> > 
> > Interesting, actually the timing should be improved in case of Primary update since Laurent added this `if (proviisioned)` statement which means that Primary won't try to connect to Secondaries after reboot if a device is provisioned. @Raigi Is/Are Secondary(ties) rebooted too in your case?
> 
> No, only primary got rebooted.
> If I run journalctl -fu aktualizr on primary after reboot, it takes a bit time, before I get any logs. Sometimes it is much longer even.
> `-- Logs begin at Thu 2020-03-26 10:02:36 UTC. -- Mar 26 10:02:55 qemux86-64 aktualizr[229]: Connecting to and getting info about IP Secondary: 10.0.3.2:9050...`

Now I understand, this delay is caused by this change https://github.com/advancedtelematic/meta-updater/commit/de993c695f58924fe8ee2b10a5b30af38afc42d3. But, IMHO, this is correct change as it makes aktualizr service not to start until all network interfaces are up and acquired IP addresses. Without this change, the Primary and Secondary discovery will fail in many cases because if this discovery process starts before a network interface acquires an IP address it will just simply fail because a connect timeout is not applied in this case and socket.connect() call just returns immediately with an error.

Raigi:
> > > > It seems to me that it takes longer for the device to come back up after rebooting.
> > > 
> > > 
> > > Interesting, actually the timing should be improved in case of Primary update since Laurent added this `if (proviisioned)` statement which means that Primary won't try to connect to Secondaries after reboot if a device is provisioned. @Raigi Is/Are Secondary(ties) rebooted too in your case?
> > 
> > 
> > No, only primary got rebooted.
> > If I run journalctl -fu aktualizr on primary after reboot, it takes a bit time, before I get any logs. Sometimes it is much longer even.
> > `-- Logs begin at Thu 2020-03-26 10:02:36 UTC. -- Mar 26 10:02:55 qemux86-64 aktualizr[229]: Connecting to and getting info about IP Secondary: 10.0.3.2:9050...`
> 
> Now I understand, this delay is caused by this change [de993c6](https://github.com/advancedtelematic/meta-updater/commit/de993c695f58924fe8ee2b10a5b30af38afc42d3). But, IMHO, this is correct change as it makes aktualizr service not to start until all network interfaces are up and acquired IP addresses. Without this change, the Primary and Secondary discovery will fail in many cases because if this discovery process starts before a network interface acquires an IP address it will just simply fail because a connect timeout is not applied in this case and socket.connect() call just returns immediately with an error.

OK, I see. So if this is expected, is it fine that I update tests to wait for update to finish for maximum of 5 minutes?

patrickvacek:
> So if this is expected, is it fine that I update tests to wait for update to finish for maximum of 5 minutes?

I suppose so. It's annoying that it takes so long, but if Mike is correct, then that's simply how long it takes for the network interface to come up, so that's how long we have to wait.

mikesul-vd:
> > So if this is expected, is it fine that I update tests to wait for update to finish for maximum of 5 minutes?
> 
> I suppose so. It's annoying that it takes so long, but if Mike is correct, then that's simply how long it takes for the network interface to come up, so that's how long we have to wait.

That's correct, it would be slightly faster if aktualizr depends just on an IP address assignment to the network interface used for communication between Primary and Secondary instead of waiting for such assignment to each network interface, but eventually we want aktualizr to communicate to the backend anyway...

PR #699, opened 2020-03-24
aktualizr: Bump to latest along with garage-sign.

patrickvacek:
Depends on https://github.com/advancedtelematic/aktualizr/pull/1613.

PR #703, opened 2020-03-25
Fix oe-selftest log message case

lbonn:
Was changed recently


lbonn:
ah nevermind, it would need to be on top of #699. I'll just push it there.

PR #704, opened 2020-03-25
layer.conf: update LAYERSERIES_COMPAT for dunfell

shr-project:

### DIFF
--- a/conf/layer.conf
+++ b/conf/layer.conf
@@ -12,7 +12,7 @@ BBFILE_PRIORITY_sota = "7"
 LAYERDEPENDS_sota = "openembedded-layer"
 LAYERDEPENDS_sota += "meta-python"
 LAYERDEPENDS_sota += "filesystems-layer"
-LAYERSERIES_COMPAT_sota = "thud warrior zeus"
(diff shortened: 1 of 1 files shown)


shr-project:
> Thanks! Is it not recommended to keep the previous branch names in `LAYERSERIES_COMPAT`? I seem to recall seeing other layers doing that as well.

It doesn't even parse with zeus, because of meta-updater/recipes-support/boost/boost_1.72.0.bbappend

Sometimes older release is kept at least temporarily, so that you don't need the very latest oe-core to parse with once the `LAYERSERIES_COMPAT` is updated, but this time dunfell is included for 2 weeks already since:
https://git.openembedded.org/openembedded-core/commit/?id=090bb3b44ba0cc01c29942c00d43e910d1ff735e
and this change in meta-updater isn't mandatory for parsing yet, because this time Richard kept zeus in oe-core temporarily as well:
https://git.openembedded.org/openembedded-core/commit/?id=1970c5f7838ff10cd6243e1624553bdc46e89fad
so oe-core and all the layers don't need to be updated at the same time, but I think 2 weeks of transition period should be enough.

PR #710, opened 2020-04-02
Fix envsetup when run in zsh

lbonn:
Well, we never promised it would work with zsh, but this should fix it.

As bash array were not bad enough, zsh has to do it differently... :cry:

### DIFF
--- a/scripts/envsetup.sh
+++ b/scripts/envsetup.sh
@@ -6,15 +6,23 @@ BUILDDIR="build"
 DISTRO="poky-sota-systemd"
 BASE_CONF="local.conf.base.append"
 
+if [ -n "$ZSH_VERSION" ]; then
+    # be more compatible with bash
+    setopt shwordsplit
+fi
+
 # A definition of a dictionary with a list of configuration files that must be appended
 # to resulting conf/local.conf file for each particular distribution.
-declare -A supported_distros=(
-    ["poky-sota-systemd"]="local.conf.systemd.append"
-    ["poky-sota"]="local.conf.base.append"
-    ["poky"]="local.conf.systemd.append local.conf.nonostree.append"
-)
+declare -A supported_distros
+supported_distros[poky-sota-systemd]="local.conf.systemd.append"
+supported_distros[poky-sota]="local.conf.base.append"
+supported_distros[poky]="local.conf.systemd.append local.conf.nonostree.append"
 
-[[ "$#" -lt 1 ]] && { echo "Usage: ${SCRIPT} <machine> [builddir] [distro=< poky-sota-systemd | poky-sota | poky >]"; return 1; }
+usage () {
+  echo "Usage: ${SCRIPT} <machine> [builddir] [distro=< poky-sota-systemd | poky-sota | poky >]"
+}
+
+[[ "$#" -lt 1 ]] && { usage; return 1; }
 [[ "$#" -ge 2 ]] && { BUILDDIR="$2"; }
 [[ "$#" -eq 3 ]] && { DISTRO="$3"; }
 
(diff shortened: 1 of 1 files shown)


alexhumphreys:
Copying this to my machine and running it I still see an error:
```
$ source meta-updater/scripts/envsetup.sh qemux86-64
meta-updater/scripts/envsetup.sh:16: no matches found: [poky-sota-systemd]=local.conf.systemd.append
```

mike-sul:
> Copying this to my machine and running it I still see an error:
> 
> ```
> $ source meta-updater/scripts/envsetup.sh qemux86-64
> meta-updater/scripts/envsetup.sh:16: no matches found: [poky-sota-systemd]=local.conf.systemd.append
> ```

maybe makes sense to try to initialize the `supported_distros` after declaration as it suggested here https://scriptingosx.com/2019/11/associative-arrays-in-zsh/

lbonn:
> Copying this to my machine and running it I still see an error:
> 
> ```
> $ source meta-updater/scripts/envsetup.sh qemux86-64
> meta-updater/scripts/envsetup.sh:16: no matches found: [poky-sota-systemd]=local.conf.systemd.append
> ```

Ah damn it works with my version of zsh (5.7.1). I'll have another look.

lbonn:
zsh supports the syntax we were using since 5.5 and Ubuntu Bionic ships with 5.4.2.

@alexhumphreys could you test another time?

alexhumphreys:
Ayy looks good :+1: 
```
➜  myproject source meta-updater/scripts/envsetup.sh qemux86-64
You had no conf/local.conf file. This configuration file has therefore been
created for you with some default values. You may wish to edit it to, for
example, select a different MACHINE (target hardware). See conf/local.conf
for more information as common configuration options are commented.

You had no conf/bblayers.conf file. This configuration file has therefore been
created for you with some default values. To add additional metadata layers
into your configuration please add entries to conf/bblayers.conf.

The Yocto Project has extensive documentation about OE including a reference
manual which can be found at:
    http://yoctoproject.org/documentation

For more information about OpenEmbedded see their website:
    http://www.openembedded.org/


### Shell environment set up for builds. ###

You can now run 'bitbake <target>'

Common targets are:
    core-image-minimal
    core-image-sato
    meta-toolchain
    meta-ide-support
```

scripts/envsetup.sh
```diff
@@ -6,13 +6,17 @@ BUILDDIR="build"
...
 # to resulting conf/local.conf file for each particular distribution.
-declare -A supported_distros=(
-    ["poky-sota-systemd"]="local.conf.systemd.append"
-    ["poky-sota"]="local.conf.base.append"
-    ["poky"]="local.conf.systemd.append local.conf.nonostree.append"
-)
+declare -A supported_distros
+supported_distros[poky-sota-systemd]="local.conf.systemd.append"
+supported_distros[poky-sota]="local.conf.base.append"
+supported_distros[poky]="local.conf.systemd.append local.conf.nonostree.append"
```
patrickvacek, line 19: 👍

scripts/envsetup.sh
```diff
@@ -35,23 +39,21 @@ fi
 METADIR=${METADIR:-${SOURCEDIR}/../..}
 
 if [[ ! -f "${BUILDDIR}/conf/local.conf" ]]; then
-  declare -a DISTRO_CONFIGS=${supported_distros[$DISTRO]}
-  [[ -n ${DISTRO_CONFIGS[@]} ]] && { echo "Using (${DISTRO_CONFIGS[*]}) for the specified distro '$DISTRO'"; } || { echo "The specified distro $DISTRO is not supported"; return 1; }
-
```
patrickvacek, line 40: While this sort of one-line if-replacement logic is a bit awkward to me, do you think the check against supported distros is not worth keeping?

scripts/envsetup.sh
```diff
@@ -35,23 +39,21 @@ fi
...
   cat "${METADIR}/meta-updater/conf/include/bblayers/sota_${MACHINE}.inc" >> conf/bblayers.conf
+
   sed -e "s/##MACHINE##/$MACHINE/g" \
       -e "s/##DISTRO##/$DISTRO/g" \
          "${METADIR}/meta-updater/conf/$BASE_CONF" >> conf/local.conf
 
-  for config in ${DISTRO_CONFIGS[@]}; do
+  for config in ${supported_distros[$DISTRO]}; do
     if [[ "$BASE_CONF" != "$config" ]]; then
       cat "${METADIR}/meta-updater/conf/$config" >> conf/local.conf
```
mike-sul, line 54: I suppose this `cat` will fail if a user specifies unsupported distro, although printing clear described failure reason and returning might be a bit better.

scripts/envsetup.sh
lbonn: You're right, I've put a check back.

scripts/envsetup.sh
lbonn: Good point

scripts/envsetup.sh
```diff
@@ -35,23 +39,26 @@ fi
 METADIR=${METADIR:-${SOURCEDIR}/../..}
 
 if [[ ! -f "${BUILDDIR}/conf/local.conf" ]]; then
-  declare -a DISTRO_CONFIGS=${supported_distros[$DISTRO]}
-  [[ -n ${DISTRO_CONFIGS[@]} ]] && { echo "Using (${DISTRO_CONFIGS[*]}) for the specified distro '$DISTRO'"; } || { echo "The specified distro $DISTRO is not supported"; return 1; }
+  if [ -z "${supported_distros[$DISTRO]}" ]; then
+    echo "The specific distro $DISTRO is not supported"
```
patrickvacek, line 43:
```suggestion
    echo "The specified distro $DISTRO is not supported"
```
Sorry for being pedantic. Maybe worth printing the Usage string here as well?

PR #713, opened 2020-04-27
OSTree task improvements

agners:
Hi, this are smaller improvements to the OSTree image tasks. Most notably it changes how /home folder is stored to align with what Fedora CoreOS is doing and what upstream OSTree today uses as the modern approach.

### DIFF
--- a/classes/image_types_ostree.bbclass
+++ b/classes/image_types_ostree.bbclass
@@ -38,9 +38,6 @@ IMAGE_CMD_ostree () {
     mkdir sysroot
     ln -sf sysroot/ostree ostree
 
-    rm -rf tmp/*
-    ln -sf sysroot/tmp tmp
-
     mkdir -p usr/rootdirs
 
     mv etc usr/
--- a/classes/image_types_ostree.bbclass
+++ b/classes/image_types_ostree.bbclass
@@ -71,13 +71,11 @@ IMAGE_CMD_ostree () {
     mkdir -p usr/share/sota/
     echo -n "${OSTREE_BRANCHNAME}" > usr/share/sota/branchname
 
-    # Preserve data in /home to be later copied to /sysroot/home by sysroot
-    # generating procedure
-    mkdir -p usr/homedirs
-    if [ -d "home" ] && [ ! -L "home" ]; then
-        mv home usr/homedirs/home
-        ln -sf var/rootdirs/home home
-    fi
+    # home directories get copied from the OE root later to the final sysroot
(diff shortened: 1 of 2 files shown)


patrickvacek:
Thanks for putting this together! It looks like a lot of good work. I'm reviewing it now and testing it in CI, but it might take just a bit of time to make sure that everything works like we expect.

patrickvacek:
CI is unfortunately broken apparently because Yocto's certs expired, which also means I can't test manually. Once that is resolved we'll retry.

I've reviewed the logic, though, and I don't see any obvious concerns. Everything looks like an improvement and update to modern OSTree usage.

BTW, I agree that the tmp handling logic we had in place was just a mistake. The code was committed in the earliest days of this project and has probably never been looked at too closely again until now. Thanks for catching that!

agners:
> CI is unfortunately broken apparently because Yocto's certs expired, which also means I can't test manually. Once that is resolved we'll retry.

Hehe, yeah our CI got tripped by it too today :-)

> I've reviewed the logic, though, and I don't see any obvious concerns. Everything looks like an improvement and update to modern OSTree usage.
> 
> BTW, I agree that the tmp handling logic we had in place was just a mistake. The code was committed in the earliest days of this project and has probably never been looked at too closely again until now. Thanks for catching that!

Yeah I was assuming that. Thanks for confirming.

agners:
@patrickvacek added two more commits, one of which addresses more or less what @eu-smirnov in #687 tried to achieve. If you prefer to have those two separate still I can pop them and create a separate merge request.

patrickvacek:
> added two more commits, one of which addresses more or less what @eu-smirnov in #687 tried to achieve. If you prefer to have those two separate still I can pop them and create a separate merge request.

Again, looks like good stuff! I probably would prefer if the new stuff were in a separate PR, though. It's a lot to test, and all of this has to be tested manually, since CI doesn't really cover the interesting parts here. Does your change to `ostree_target_hash` address https://github.com/advancedtelematic/meta-updater/issues/412?

lbonn:
The changes look good and well motivated in their commit messages, thanks!

agners:
@patrickvacek ok, dropped off the last two patches! Will create a separate pull request for those.

agners:
@patrickvacek and no, it does not address #412.

eu-smirnov:
Looks good to me!

agners:
Just wanted to point out that using the new `/home` location also allows to use ro `/sysroot` (see https://github.com/ostreedev/ostree/issues/2086).

PR #716, opened 2020-04-29
Use checksum as deployment source and refspec in the deployment

agners:
This merge request uses the OSTree commit checksum as source reference in the OSTree deployment task `do_image_ota`. This guarantees that the exact commit this build commited in `do_image_ostreecommit` actually gets deployed.

It also makes sure that `origin refspec` appears in `ostree admin status`, and hence will be used when using `ostree admin upgrade`. This is similar to what #687 tries to achieve.

### DIFF
--- a/classes/image_types_ostree.bbclass
+++ b/classes/image_types_ostree.bbclass
@@ -168,25 +168,19 @@ IMAGE_CMD_ostreecommit () {
     fi
 
     # Commit the result
-    ostree --repo=${OSTREE_REPO} commit \
+    ostree_target_hash=$(ostree --repo=${OSTREE_REPO} commit \
            --tree=dir=${OSTREE_ROOTFS} \
            --skip-if-unchanged \
            --branch=${OSTREE_BRANCHNAME} \
            --subject="${OSTREE_COMMIT_SUBJECT}" \
            --body="${OSTREE_COMMIT_BODY}" \
-           --add-metadata-string=version="${OSTREE_COMMIT_VERSION}" \
-           --bind-ref="${OSTREE_BRANCHNAME}-${IMAGE_BASENAME}"
+           --add-metadata-string=version="${OSTREE_COMMIT_VERSION}")
+
+    echo $ostree_target_hash > ${WORKDIR}/ostree_manifest
 
     if [ "${OSTREE_UPDATE_SUMMARY}" = "1" ]; then
         ostree --repo=${OSTREE_REPO} summary -u
     fi
-
-    # To enable simultaneous bitbaking of two images with the same branch name,
-    # create a new ref in the repo using the basename of the image. (This first
-    # requires deleting it if it already exists.) Fixes OTA-2211.
-    ostree --repo=${OSTREE_REPO} refs --delete ${OSTREE_BRANCHNAME}-${IMAGE_BASENAME}
-    ostree_target_hash=$(cat ${OSTREE_REPO}/refs/heads/${OSTREE_BRANCHNAME})
-    ostree --repo=${OSTREE_REPO} refs --create=${OSTREE_BRANCHNAME}-${IMAGE_BASENAME} ${ostree_target_hash}
 }
 
 IMAGE_TYPEDEP_ostreepush = "ostreecommit"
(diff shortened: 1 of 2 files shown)


classes/image_types_ostree.bbclass
```diff
@@ -165,25 +165,19 @@ IMAGE_CMD_ostreecommit () {
...
 
     if [ "${OSTREE_UPDATE_SUMMARY}" = "1" ]; then
         ostree --repo=${OSTREE_REPO} summary -u
     fi
-
-    # To enable simultaneous bitbaking of two images with the same branch name,
-    # create a new ref in the repo using the basename of the image. (This first
-    # requires deleting it if it already exists.) Fixes OTA-2211.
-    ostree --repo=${OSTREE_REPO} refs --delete ${OSTREE_BRANCHNAME}-${IMAGE_BASENAME}
-    ostree_target_hash=$(cat ${OSTREE_REPO}/refs/heads/${OSTREE_BRANCHNAME})
```
eu-smirnov, line 185: I don't think I fully understand what was happening here. If another parallel bitbake also did commit before this line but after the line 168 was executed, then it will give a wrong hash, won't it?

Anyway, I think the current change makes this logic more straightforward.

classes/image_types_ota.bbclass
```diff
@@ -63,13 +63,23 @@ IMAGE_CMD_ota () {
...
-	ostree admin --sysroot=${OTA_SYSROOT} deploy ${kargs_list} --os=${OSTREE_OSNAME} ${ostree_target_hash}
+	# Create the same reference on the device we use in the archive OSTree
+	# repo in ${OSTREE_REPO}. This reference will show up when showing the
+	# deployment on the device:
+	# ostree admin status
+	# If a remote with the name ${OSTREE_OSNAME} is configured, this also
+	# will allow to use:
+	# ostree admin upgrade
+	ostree --repo=${OTA_SYSROOT}/ostree/repo refs --create=${OSTREE_OSNAME}:${OSTREE_BRANCHNAME} ${ostree_target_hash}
+	ostree admin --sysroot=${OTA_SYSROOT} deploy ${kargs_list} --os=${OSTREE_OSNAME} ${OSTREE_OSNAME}:${OSTREE_BRANCHNAME}
```
eu-smirnov, line 82: Cool! That's definitely better than #687

classes/image_types_ostree.bbclass
agners: No, because this task is protected by a lockfile inside the repository location. So if multiple bitbake instances run at the same time, this code is guaranteed to run in a sequence.

https://github.com/advancedtelematic/meta-updater/blob/7a9dffea0efeb53d73a8f4065bdbf3b0638593c0/classes/image_types_ostree.bbclass#L161

agners: The `do_image_ota` task is not using that lock. So passing the checksum via the bitbake instance private TMPDIR guarantees that we do not end up using a different checksum in the `do_image_ota` task.

classes/image_types_ota.bbclass
agners: @patrickvacek currently I add the remote manually for testing, but we might add a recipe which then adds a default remote for a pure OSTree use-case.

agners:
Updated the merge request. Your fix seems correct to me too. Somehow forgot to check the other tasks in `image_types_ostree.bbclass`, I remember checking `image_types_ota.bbclass` quite thoroughly. Sorry about that.

> where `$OSTREE_REPO` in practice is something like `tmp/deploy/images/qemux86-64/ostree_repo/`

We overwrite `$OSTREE_REPO` to be in a common space for all our builds (hence #412...). That makes using branch names to pass the checksum of the current builds commit from one task to another susceptible to races.

> where `$WORKDIR` in practice is something like tmp/work/qemux86_64-poky-linux/core-image-minimal/1.0-r0/.

From my point of view the most important thing is that `$WORKDIR` is under `$TMPDIR`, which is a build specific directory. Running two builds in one `$TMPDIR` at the same time is not a supported use case by bitbake/OE. So with that restriction from bitbake/OE it is guaranteed that the commit hash is stored in a per build "private" location, no matter where `$OSTREE_REPO` points to.

We moved away from sharing/reusing build directories. What we do is sharing sstate and download, but create a separate build dir for basically every machine/distro/image configuration. I think in theory, one can build different machine/image combination in a single build dir without negative consequences (we fixed bugs in that domain quite a bit back when we did that, but it generally works), but changing distro (or distro features) often lead to larger problems not easily fixable. It might be better today, but at least 1-2 years ago that did not work well.

agners:
@patrickvacek is this (and #713) something you consider for backport? I would be interested mainly in this (but also #713) for zeus and dunfell.

patrickvacek:
> @patrickvacek is this (and #713) something you consider for backport? I would be interested mainly in this (but also #713) for zeus and dunfell.

Yep, I was planning on backporting this for sure and probably #713 as well. We're trying to issue new releases of aktualizr and backport meta-updater commits approximately every four weeks, so we'll get around it it in about 2-3 weeks if you can wait.

agners:
> so we'll get around it it in about 2-3 weeks if you can wait.

@patrickvacek our feature freeze is planned in 4 week, so that should work out.

PR #722, opened 2020-05-10
meta: change virtual/network-configuration to network-configuration

liuming50:
The name 'virtual' is a recipe specific concept but not a package
specific one, it's confusing that a package provides 'virtual/' names.

Let's drop 'virtual/' from network-configuration, to keep consistent
with yocto naming styles.

### DIFF
--- a/classes/sota_m3ulcb.bbclass
+++ b/classes/sota_m3ulcb.bbclass
@@ -8,5 +8,5 @@ OSTREE_BOOTLOADER ?= "u-boot"
 
 UBOOT_MACHINE_sota = "${@d.getVar('SOC_FAMILY').split(':')[1]}_ulcb_defconfig"
 
-PREFERRED_RPROVIDER_virtual/network-configuration ?= "connman"
-IMAGE_INSTALL_append_sota = " virtual/network-configuration "
+PREFERRED_RPROVIDER_network-configuration ?= "connman"
+IMAGE_INSTALL_append_sota = " network-configuration "
--- a/classes/sota_minnowboard.bbclass
+++ b/classes/sota_minnowboard.bbclass
@@ -8,5 +8,5 @@ IMAGE_FSTYPES_remove_sota = "live hddimg"
 OSTREE_KERNEL_ARGS ?= "ramdisk_size=16384 rw rootfstype=ext4 rootwait rootdelay=2 console=ttyS0,115200 console=tty0"
 IMAGE_INSTALL_append = " minnowboard-efi-startup"
 
-PREFERRED_RPROVIDER_virtual/network-configuration ?= "connman"
-IMAGE_INSTALL_append_sota = " virtual/network-configuration "
+PREFERRED_RPROVIDER_network-configuration ?= "connman"
+IMAGE_INSTALL_append_sota = " network-configuration "
--- a/classes/sota_porter.bbclass
+++ b/classes/sota_porter.bbclass
@@ -7,5 +7,5 @@ IMAGE_BOOT_FILES_sota += "porter-bootfiles/*"
 OSTREE_BOOTLOADER ?= "u-boot"
 UBOOT_MACHINE_sota = "porter_config"
 
-PREFERRED_RPROVIDER_virtual/network-configuration ?= "connman"
-IMAGE_INSTALL_append_sota = " virtual/network-configuration "
+PREFERRED_RPROVIDER_network-configuration ?= "connman"
+IMAGE_INSTALL_append_sota = " network-configuration "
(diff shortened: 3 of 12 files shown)


patrickvacek:
Looks fine to me. I'm running CI now. @lbonn I think you originally created that logic, do you wanna take a quick look?

lbonn:
We can change that but I also don't really get the argument and recipe/package distinction.

There are similar examples in poky, for example `RPROVIDES_${PN} += "virtual/systemd-bootconf"` in `meta/recipes-core/systemd/systemd-bootconf_1.00.bb` and IMO it's helpful to denote it is a more abstract feature that can be fulfilled by different packages.

liuming50:
> We can change that but I also don't really get the argument and recipe/package distinction.
> 
> There are similar examples in poky, for example `RPROVIDES_${PN} += "virtual/systemd-bootconf"` in `meta/recipes-core/systemd/systemd-bootconf_1.00.bb` and IMO it's helpful to denote it is a more abstract feature that can be fulfilled by different packages.

I attempt to say, the package naming schemes shall follow the real package backends like ipk, rpm, deb, in which, there is not a 'virtual/' syntax, if I understood correctly. (Please correct me, I might be wrong)

According to the 'virtual/systemd-bootconf', yes, I notice that they are in OE, but from my point of view, they should be fixed, I could make patch for that to ping OE.

lbonn:
Are you concerned that 'virtual/network-configuration' cannot appear in a package file name? It should not, it's only used as a proxy package name internally during bitbaking.

One thing that is idiomatic and documented is tweaking virtual/kernel's preferred provider to use another kernel recipe https://www.yoctoproject.org/docs/current/bitbake-user-manual/bitbake-user-manual.html#var-bb-PREFERRED_PROVIDER. Is there a fundamental difference from this use case and ours, that we do not know about?

liuming50:
> Are you concerned that 'virtual/network-configuration' cannot appear in a package file name? It should not, it's only used as a proxy package name internally during bitbaking.

Attempt to say that is not true, any name in RPROVIDES would be written into package, for instance rpm, that is 'Provides' field.

> 
> One thing that is idiomatic and documented is tweaking virtual/kernel's preferred provider to use another kernel recipe https://www.yoctoproject.org/docs/current/bitbake-user-manual/bitbake-user-manual.html#var-bb-PREFERRED_PROVIDER. Is there a fundamental difference from this use case and ours, that we do not know about?

It's different since that is PREFERRED_PROVIDER, but not PREFERRED_RPROVIDER, 'virtual/' in PREFERRED_PROVIDER is totally fine, it is a yocto internal variable for recipes.

patrickvacek:
> > Are you concerned that 'virtual/network-configuration' cannot appear in a package file name? It should not, it's only used as a proxy package name internally during bitbaking.
> 
> Attempt to say that is not true, any name in RPROVIDES would be written into package, for instance rpm, that is 'Provides' field.
> 
> > One thing that is idiomatic and documented is tweaking virtual/kernel's preferred provider to use another kernel recipe https://www.yoctoproject.org/docs/current/bitbake-user-manual/bitbake-user-manual.html#var-bb-PREFERRED_PROVIDER. Is there a fundamental difference from this use case and ours, that we do not know about?
> 
> It's different since that is PREFERRED_PROVIDER, but not PREFERRED_RPROVIDER, 'virtual/' in PREFERRED_PROVIDER is totally fine, it is a yocto internal variable for recipes.

I didn't realize there were such differences in how the two variables were treated. I think this is fine; it's just nomenclature, but I also think it would be even better to use `PREFERRED_PROVIDER` instead of `PREFERRED_RPROVIDER`, although we can't remember why that didn't work in the first place. Perhaps we should try that once more before going with this.

liuming50:
> I didn't realize there were such differences in how the two variables were treated. I think this is fine; it's just nomenclature, but I also think it would be even better to use `PREFERRED_PROVIDER` instead of `PREFERRED_RPROVIDER`, although we can't remember why that didn't work in the first place. Perhaps we should try that once more before going with this.

But PREFERRED_PROVIDER does not fit our need? It's to declare a recipe dependency, usually in "DEPENDS", but not for "RDEPENDS" and "IMAGE_INSTALL", that requires PREFERRED_RPROVIDER?

patrickvacek:
> But PREFERRED_PROVIDER does not fit our need? It's to declare a recipe dependency, usually in "DEPENDS", but not for "RDEPENDS" and "IMAGE_INSTALL", that requires PREFERRED_RPROVIDER?

Yeah, actually, that's a good point. This is probably fine as it is. Sorry for the delay here!

PR #724, opened 2020-05-11
sota_sanity.bbclass: introduce sota_check_boolean_variable

liuming50:
The current sanity check are too strict for some boolean variables,
introduce sota_check_boolean_variable to allow a boolean value to be
set like yes/y/true/t/1 or no/n/false/f/0.

### DIFF
--- a/classes/sota_sanity.bbclass
+++ b/classes/sota_sanity.bbclass
@@ -1,5 +1,12 @@
 # Sanity check the sota setup for common misconfigurations
 
+def sota_check_boolean_variable(var, d):
+    try:
+        oe.types.boolean(d.getVar(var))
+    except:
+        return False
+    return True
+
 def sota_check_overrides(status, d):
     for var in (d.getVar('SOTA_OVERRIDES_BLACKLIST') or "").split():
         if var in d.getVar('OVERRIDES').split(':'):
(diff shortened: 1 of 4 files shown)


classes/sota_sanity.bbclass
```diff
@@ -47,14 +54,14 @@ def sota_check_variables_validity(status, d):
...
-    if d.getVar("OSTREE_UPDATE_SUMMARY") and d.getVar("OSTREE_UPDATE_SUMMARY") not in ("0", "1", ""):
-        status.addresult("OSTREE_UPDATE_SUMMARY should be set to 0 or 1.\n")
-    if d.getVar("OSTREE_DEPLOY_DEVICETREE") and d.getVar("OSTREE_DEPLOY_DEVICETREE") not in ("0", "1", ""):
-        status.addresult("OSTREE_DEPLOY_DEVICETREE should be set to 0 or 1.\n")
-    if d.getVar("GARAGE_SIGN_AUTOVERSION") and d.getVar("GARAGE_SIGN_AUTOVERSION") not in ("0", "1", ""):
-        status.addresult("GARAGE_SIGN_AUTOVERSION should be set to 0 or 1.\n")
-    if d.getVar("SOTA_DEPLOY_CREDENTIALS") and d.getVar("SOTA_DEPLOY_CREDENTIALS") not in ("0", "1", ""):
-        status.addresult("SOTA_DEPLOY_CREDENTIALS should be set to 0 or 1.\n")
+    if not sota_check_boolean_variable("OSTREE_UPDATE_SUMMARY", d):
+        status.addresult("OSTREE_UPDATE_SUMMARY (=%s) should be set to yes/y/true/t/1 or no/n/false/f/0.\n" % d.getVar("OSTREE_UPDATE_SUMMARY"))
```
lbonn, line 58: The message does not match the actual logic handling these variables which has been kept simple until now.

https://github.com/advancedtelematic/meta-updater/blob/master/classes/image_types_ostree.bbclass#L180

liuming50: Thanks, you are absolutely right, will send a V2 soon.

liuming50:
Changes in V2:

1 Also change the checking code to use oe.types.boolean.

PR #726, opened 2020-05-25
Remove boost 1.72 bbappend

lbonn:
poky switched to 1.73 which includes this patch


lbonn:
oe-selftest passed after a retry but that didn't get reflected in the github status.

PR #728, opened 2020-05-29
image_types_ostree: allow specifying a device tree to deploy

antznin:
When setting `OSTREE_DEPLOY_DEVICETREE` to 1, it will by default deploy
all the device tree blobs present in `KERNEL_DEVICETREE`. Adding
`OSTREE_DEVICETREE` would allow specifying a specific device tree blob
(or several dtbs), thus resulting in only the specified ones in the
image. This is particularely useful because ostree selects the first
device tree it finds in /boot, and discards the remaining ones.

### DIFF
--- a/classes/image_types_ostree.bbclass
+++ b/classes/image_types_ostree.bbclass
@@ -10,6 +10,7 @@ OSTREE_COMMIT_BODY ??= ""
 OSTREE_COMMIT_VERSION ??= "${DISTRO_VERSION}"
 OSTREE_UPDATE_SUMMARY ??= "0"
 OSTREE_DEPLOY_DEVICETREE ??= "0"
+OSTREE_DEVICETREE ??= "${KERNEL_DEVICETREE}"
 
 BUILD_OSTREE_TARBALL ??= "1"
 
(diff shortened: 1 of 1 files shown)


patrickvacek:
>  Adding `OSTREE_DEVICETREE` would allow specifying a specific device tree blob  (or several dtbs), thus resulting in only the specified ones in the image. This is particularely useful because ostree selects the first device tree it finds in /boot, and discards the remaining ones.

This looks fine and CI passes. But what would be the point of selecting multiple dtbs if ostree will discard everything after the first?

antznin:
> This looks fine and CI passes. But what would be the point of selecting multiple dtbs if ostree will discard everything after the first?

Well, it's true that in my case, there's no point to having multiple device trees, and I'm failing to see where having several dtbs would be useful.

When adding the variable, my mindset was that this modification should not force someone to choose only one device tree, but let the user choose.

patrickvacek:
> When adding the variable, my mindset was that this modification should not force someone to choose only one device tree, but let the user choose.

Yep, that makes sense. What is stopping the user from changing `KERNEL_DEVICETREE` directly, though?

antznin:
> Yep, that makes sense. What is stopping the user from changing `KERNEL_DEVICETREE` directly, though?

Actually, I tried that, and it worked in my case. But then, since this is a machine-specific variable, defined in the machine configuration, my guess is that it was cleaner to leave the variable untouched (I'm not currently aware of the usage of this variable other than in image_types_ostree).

patrickvacek:
> since this is a machine-specific variable, defined in the machine configuration, my guess is that it was cleaner to leave the variable untouched (I'm not currently aware of the usage of this variable other than in image_types_ostree).

Yes, that's a good point. I was just reviewing our own documentation and came across https://github.com/advancedtelematic/meta-updater-raspberrypi/#device-tree-configuration, which mentions `EXTERNAL_KERNEL_DEVICETREE`. That seems to be designed to work around similar problems, but doesn't necessarily solve your problem or address your concerns. In fact, it makes me wonder if we really support `EXTERNAL_KERNEL_DEVICETREE`, since we only use `KERNEL_DEVICETREE` here. I still need to think a bit, but I'm leaning towards accepting this as is.

lbonn:
> In fact, it makes me wonder if we really support `EXTERNAL_KERNEL_DEVICETREE`, since we only use `KERNEL_DEVICETREE` here. I still need to think a bit, but I'm leaning towards accepting this as is.

Maybe not indeeed, but I don't really have a problem keeping it in the docs, ihis is just a suggested solution to the problem faced on rpi.

antznin:
Will this be backported to the supported branches ? Thanks for the quick answers.

patrickvacek:
> Will this be backported to the supported branches ?

Yes. This unfortunately just barely missed a round of backporting and QA, though, so it will be about a month or so before our next regularly scheduled round. What branches do you want it on, and what's the urgency?

antznin:
> Yes. This unfortunately just barely missed a round of backporting and QA, though, so it will be about a month or so before our next regularly scheduled round.

Ah, it's okay, it can wait. :-)

> What branches do you want it on, and what's the urgency?

Ideally, all of your supported ones, since I'm also working on a multi-release project, where not all machines are supported on the latest release.

No urgency here. Thanks!

PR #729, opened 2020-05-29
Bump aktualizr to 2020.7 revision

lbonn:
And garage-sign to 0.7.1-4

### DIFF
--- a/classes/image_types_ostree.bbclass
+++ b/classes/image_types_ostree.bbclass
@@ -196,7 +196,7 @@ IMAGE_CMD_ostreepush () {
 
     if [ -n "${SOTA_PACKED_CREDENTIALS}" ]; then
         if [ -e ${SOTA_PACKED_CREDENTIALS} ]; then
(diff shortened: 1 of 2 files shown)


lbonn:
Ah we removed the ability to run garage-push with log level trace via `-vv`, which is what's done in `ostreepush`.

Should we just use settle for debug log level or do we need to revert something?

lbonn:
Ah nevermind, we can use `-loglevel` instead

classes/image_types_ostree.bbclass
```diff
@@ -196,7 +196,7 @@ IMAGE_CMD_ostreepush () {
 
     if [ -n "${SOTA_PACKED_CREDENTIALS}" ]; then
         if [ -e ${SOTA_PACKED_CREDENTIALS} ]; then
-            garage-push -vv --repo=${OSTREE_REPO} \
+            garage-push -loglevel 0 --repo=${OSTREE_REPO} \
```
patrickvacek, line 199: Two dashes, oder?

PR #730, opened 2020-06-02
2020.7 on thud

lbonn:


lbonn:
Ah it seems like our ostree version is too old here. The `--modern` switch was added in 2019.3. We can either update ostree on thud (we ship our own) or drop some of the changes.

patrickvacek:
> Ah it seems like our ostree version is too old here. The `--modern` switch was added in 2019.3. We can either update ostree on thud (we ship our own) or drop some of the changes.

Oh, good catch. I'm fine with either way, but just dropping some of the changes seems easier. @agners mentioned in https://github.com/advancedtelematic/meta-updater/pull/716#issuecomment-628635230 that he was only interested in backporting these OSTree changes to zeus and dunfell, so there's no specific demand there as far as I know.

agners:
@patrickvacek yes, zeus is currently the oldest version we support and makes use of OSTree. So from my side not back-porting those patches is fine. I guess the only reason to backport them would be to avoid conflicts with potential upcoming changes. Depending on how long thud will be supported it might make sense to bump OSTree and still backport them.

lbonn:
I've opted for the more conservative option and dropped the infringing commit.

PR #735, opened 2020-06-17
Add recipe to build images with binary primaries

lbonn:


Raigi:
Nice, I've been wanting to have that functionality for creating some OTF tests.
Triggered OTF pipeline with this change.

lbonn:
@patrickvacek Ah good point,  that might be better, even if it's only used for tests

recipes-test/binary-pacman/aktualizr-binary-pacman.bb
```diff
@@ -0,0 +1,21 @@
+DESCRIPTION = "Configure aktualizr with a binary package manager"
+LICENSE = "MPL-2.0"
+LIC_FILES_CHKSUM = "file://${COMMON_LICENSE_DIR}/MPL-2.0;md5=815ca599c9df247a0c7f619bab123dad"
+
+inherit allarch
+
+SRC_URI = "\
+    file://10-pacman.toml \
```
mike-sul, line 8: I just wonder if it won't interfer with this https://github.com/advancedtelematic/meta-updater/blob/master/recipes-test/demo-config/files/30-pacman-config.toml in the case if `aktualizr-binary-pacman` is added to the local.conf and the secondary image is being built.

lbonn: They can probably interfere if they are installed at the same time but I don't see why you would enable both. Do you know of an use case?

mike-sul: I usually used to use the same build/conf/local.conf for building both `primary-image` and `secondary-image`, in this case, it means that I should comment/uncomment addition of aktualizr-binary-pacman each time I switch between building primary and secondary image.

You can add this `aktualizr-binary-pacman` unconditionally regardless of an image type so it can be used for both primary and secondary. Maybe just remove the stuff I introduced (`UPDATE_TYPE`) and use this recipe so primary and secondary are configured in a uniform way. 

I apologize for writing comments to it all as I don't work for HERE anymore :), I just see that your approach is more inline with OE/Yocto than the one I did for secondary.

lbonn: No need to apologize, every input is appreciated :).

In general, it's best to use `IMAGE_APPEND` in your image recipe and not in the local.conf indeed.

Raigi: Just FYI, OTF is currently using UPDATE_TYPE for testing secondaries with binary updates, let me know if you decide to change it, so I can update tests accordingly

PR #738, opened 2020-06-24
WIP: OLPSUP-10887 Add support for multiple HWIDs

tkfu:
A software image can be compatible with several different hardware IDs, because different hardware IDs can have the same MACHINE value.

WIP, not yet tested, DNM, etc.


tkfu:
Note: I realize that I still need to implement a check for invalid characters in `GARAGE_TARGET_COMPATIBLE_HWIDS`, but if this general approach is ok that should be a pretty quick fix.

patrickvacek:
Looks fine, we should just test thoroughly to make sure it really does what we think it does. I wish we could reuse an existing variables, but it looks like both `SOTA_HARDWARE_ID` and `OSTREE_BRANCHNAME` won't work, so this is fine. If we merge this, let's not forget to update https://docs.ota.here.com/ota-client/latest/build-configuration.html#_configuration_variables_in_meta_updater, too.

patrickvacek:
CI failure is unrelated.

PR #681, opened 2020-02-10
Introduce ostree kernel initramfs

liuming50:

PR #740, opened 2020-06-27
Cherry pick several patches from master to zeus

liuming50:


liuming50:
> Looks like it should be good, but we'll probably want the fix from #739, too.

Yes, will send a V2 to cherry pick that commit. (after it's merged in master)

liuming50:
Changes in patch set 2:

1 Also cherry picked the fix https://github.com/advancedtelematic/meta-updater/pull/739

patrickvacek:
Looks like CI is happy, and the code looks fine to me. @Raigi do you mind taking a look at this? I see OTF is already running. This is just a backport done a little ahead of schedule.

Raigi:
> Looks like CI is happy, and the code looks fine to me. @Raigi do you mind taking a look at this? I see OTF is already running. This is just a backport done a little ahead of schedule.

Creating and updating raspberrypi devices worked. @patrickvacek Anyting specific I should check?

patrickvacek:
> Creating and updating raspberrypi devices worked. @patrickvacek Anyting specific I should check?

Great, thanks! It's all been tested on master, too, and I don't see anything specific to be concerned about. I think we're good to go!

shr-project:
Hi, with the fix from #739 the recipes parses OK, but still fails to build, it would be nice to automatically skip it when it isn't supposed to work. Now it's the only issue in my world CI builds for OE master (and now in zeus as well)

```
tmp-glibc/work/qemux86-webos-linux/ostree-kernel-initramfs/0.0.1-r0/temp/run.do_install.3011' failed with exit code 1:
cp: -r not specified; omitting directory 'tmp-glibc/deploy/images/qemux86/'
WARNING: exit code 1 from a shell command.
```

I guess this is from
```
cp ${DEPLOY_DIR_IMAGE}/${OSTREE_KERNEL} $kerneldir/vmlinuz
```
line, so it's not only empty INITRAMFS_IMAGE but also empty OSTREE_KERNEL which is causing this to fail building.

patrickvacek:
> Hi, with the fix from #739 the recipes parses OK, but still fails to build, it would be nice to automatically skip it when it isn't supposed to work. Now it's the only issue in my world CI builds for OE master (and now in zeus as well)

I agree, but it isn't clear to me why you are building it. How did you get there without having already pulled in `classes/sota.bbclass`, which is where that variable gets defined?

shr-project:
> I agree, but it isn't clear to me why you are building it. How did you get there without having already pulled in classes/sota.bbclass, which is where that variable gets defined?

"bitbake world" builds everything which isn't explicitly marked as incompatible with given MACHINE.

meta-updater layer is included, because it's used by some other MACHINEs (e.g. raspberrypi4 and that does inherit sota).

shr-project:
Also now these changes are in zeus and master, but missing in the dunfell branch.

patrickvacek:
> "bitbake world" builds everything which isn't explicitly marked as incompatible with given MACHINE.

We don't build or test `qemux86` regularly, but it should work in theory.

> meta-updater layer is included, because it's used by some other MACHINEs (e.g. raspberrypi4 and that does inherit sota).

I'm still confused why you would include the meta-updater layer if you aren't using the `sota` class. But I suppose more important is trying to keep the recipe as widely functional as possible. Is setting `OSTREE_KERNEL` conditionally empty good enough? @liuming50 @antznin do either of you have other ideas or opinions here?

> Also now these changes are in zeus and master, but missing in the dunfell branch.

That is correct, we haven't done a full round of backporting yet. @liuming50 asked to if they could backport a few things specifically for zeus now and I had no problem with that. We'll catch up with the rest in a week or two.

shr-project:
> I'm still confused why you would include the meta-updater layer if you aren't using the sota class.

Because raspberrypi builds are using sota and the conf/bblayers.conf is the same for multimachine builds as it should be.

> Is setting OSTREE_KERNEL conditionally empty good enough?

Not in the current implementation, cp will fail as shown above, because it will expand to copying while deploy/image directory instead of just the kernel.

antznin:
Hi,

Going back to https://github.com/advancedtelematic/meta-updater/pull/681, I'm having trouble understanding the reason why the deployment of the kernel/initramfs/dtbs have moved to the `ostree-kernel-initramfs` recipes. From my understanding, the goal of this PR was to comply with ostree's new implementation, quoting @liuming50: 
> this patch mainly aims to adapt to the new ostree implementation about how it stores kernel, devicetrees and initramfs, which is in /lib/modules/${KERNEL_VERSION}, the major benefit could be a easier deployment.

What are the benefits of having these deployment operations into a standalone recipes? From my understanding, @shr-project's issue is that `bitbake world` triggers the build of `ostree-kernel-initramfs`. But how is this recipes supposed to build if it is not related, or does not inherit sota*.bbclass, which defines all OSTREE_* variables?

I'm having an issue understanding why the PR didn't simply change the bbclass in order to adapt to the new ostree's kernel/ramfs/dtb locations (`/lib/modules/$kver`).

liuming50:
Hi,

To not get involved by 'bitbake world', we could set "EXCLUDE_FROM_WORLD =
"1"" to fix that.

About the reason why I have put them into a standalone recipe instead of in
image bblcass, mainly because it needs refer to kernel version, hence need
inherit kernel-modules bbclass, which is not good to be done in image
bbclass.

the best,
thank you

Antonin Godard <notifications@github.com> 於 2020年7月1日 週三 下午1:53寫道：

shr-project:
> To not get involved by 'bitbake world', we could set "EXCLUDE_FROM_WORLD =
"1"" to fix that.

This will prevent only "bitbake world" trying to build it, but people who explicitly try to "bitbake ostree-kernel-initramfs" will still get that useless error from cp trying to copy whole deploy/images. Adding raise bb.parse.SkipRecipe in the recipe when OSTREE_KERNEL isn't set will skip it for world builds as well as provide reasonable explanation to user who tried to build it for wrong MACHINE or with incorrect config.

liuming50:
But who really want "bitbake ostree-kernel-initramfs" without inheriting
sota.bbclass? I attempt to say, if a developer likes to do so, that kind of
error could happen on a lot of recipes in meta-updater, not only
for ostree-kernel-initramfs.

the best,
thank you

shr-project <notifications@github.com> 於 2020年7月1日 週三 下午2:54寫道：

shr-project:
> But who really want "bitbake ostree-kernel-initramfs" without inheriting sota.bbclass?

Someone with wrong MACHINE or with incorrect config as said before.

Providing reasonable error message when OSTREE_KERNEL isn't set (and recipe blindly expects that it's always defined) is much better than

```
cp: -r not specified; omitting directory 'tmp-glibc/deploy/images/qemux86/'
```

And it takes only 3 more lines to do that
```
python() {
    if not d.getVar('OSTREE_KERNEL'):
        raise bb.parse.SkipRecipe('OSTREE_KERNEL isn't defined, maybe your MACHINE config doesn't inherit sota.bbclass?')
}
```
compared with useless
```
EXCLUDE_FROM_WORLD = "1"
```

liuming50:
Yes, this could be a better fix with a clear message. Thanks.

// Ming LIu


But who really want "bitbake ostree-kernel-initramfs" without inheriting

PR #743, opened 2020-06-30
ostree-kernel-initramfs: deploy dtbs from OSTREE_DEVICETREE

liuming50:
Signed-off-by: antznin <agodard@witekio.com>


liuming50:
@patrickvacek I noticed there is a new fix for ostree-kernel-initramfs, needs to be backported to zeus.

PR #745, opened 2020-07-03
ostree-kernel-initramfs: fix devicetree deployment

antznin:
Following the changes in ostree's deployment of the kernel, initramfs
and devicetree in /lib/modules/$kver, the deployment method of the
device tree also changed. Instead of picking the first device tree it
finds at a given location, ostree looks at a file named devicetree, next
to kernel and initramfs in /lib/modules/$kver.

This commit modifies ostree-kernel-initramfs to deploy the devicetree
from the sota-defined variable OSTREE_DEVICETREE. It will pick the
first one from the list of device trees that OSTREE_DEVICETREE defines,
and copy it to /lib/modules/$kver. Note that since OSTREE_DEVICETREE
equals to KERNEL_DEVICETREE when it isn't explicitly defined, it could
indeed be a list of device trees.

This defines the way ostree picks the devicetree https://github.com/ostreedev/ostree/blob/6ed48234ba579ff73eb128af237212b0a00f2057/src/libostree/ostree-sysroot-deploy.c#L1068 if I'm not mistaken.

Also, we could probably get rid of the `dtb` directory and pick the device tree from DEPLOY_DIR_IMAGE directly, since it isn't looked by ostree? Requesting comments from @patrickvacek and @liuming50 if you don't mind.

### DIFF
--- a/recipes-sota/ostree-kernel-initramfs/ostree-kernel-initramfs_0.0.1.bb
+++ b/recipes-sota/ostree-kernel-initramfs/ostree-kernel-initramfs_0.0.1.bb
@@ -14,7 +14,9 @@ ALLOW_EMPTY_ostree-devicetrees = "1"
 
 FILES_ostree-kernel = "${nonarch_base_libdir}/modules/*/vmlinuz"
 FILES_ostree-initramfs = "${nonarch_base_libdir}/modules/*/initramfs.img"
-FILES_ostree-devicetrees = "${nonarch_base_libdir}/modules/*/dtb/*"
+FILES_ostree-devicetrees = "${nonarch_base_libdir}/modules/*/dtb/* \
+    ${nonarch_base_libdir}/modules/*/devicetree \
+"
 
 PACKAGE_ARCH = "${MACHINE_ARCH}"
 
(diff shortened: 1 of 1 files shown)


liuming50:
My bad, I forgot a commit from ostree upstream is needed to support multiple devicetree selection: https://github.com/ostreedev/ostree/pull/2001

I am not opposite this change, but I think it might be better if we back port the above commit (or bump ostree revision to include it) ?

It supports multiple devicetrees existing in ostree sysroot and chosen by fdt_file in boot script.

antznin:
Hi @liuming50, AFAIK ostree is included in meta-openembedded and the version seems to rely on tags. I wasn't aware of the `dtb` directory, since I was taking a look at the code of 2020.3. Your implementation seems a bit ahead of the curve, so it will probably be relevant on the next release?

patrickvacek:
@liuming50 Do you see any conflict with this change and your implementation? Alternatively, is anything stopping us from bumping the version of OSTree in meta-openembedded?

liuming50:
> Hi @liuming50, AFAIK ostree is included in meta-openembedded and the version seems to rely on tags. I wasn't aware of the `dtb` directory, since I was taking a look at the code of 2020.3. Your implementation seems a bit ahead of the curve, so it will probably be relevant on the next release?

Yes, you are right, the latest release tag is 2020.03.

liuming50:
> @liuming50 Do you see any conflict with this change and your implementation? Alternatively, is anything stopping us from bumping the version of OSTree in meta-openembedded?

This should not conflict with my implementation, and it is a good complementary fix provided that we dont have https://github.com/ostreedev/ostree/pull/2001

I think we can merge it until next ostree release include https://github.com/ostreedev/ostree/pull/2001?

PR #749, opened 2020-07-09
Feat/dunfell/2020.8

patrickvacek:


patrickvacek:
Looks like that oe-selftest change that broke master for us has been backported to dunfell, but not my fix. I'll have to make another upstream PR for that. In the meantime, we can use a pinned commit of poky.

PR #752, opened 2020-07-09
Feat/thud/2020.8

patrickvacek:


Raigi:
Is something different in thud/warrior/dunfell compared to zeus? A test with secondaries failed, claiming `Initializer error: Server error: Error registering device: 412 {"code":"replace_ecu_assignment_exists","description":"Cannot replace ecus`

The test was not trying to replace ecus. It is installing binary update on secondary, then installing ostree update on primary, the secondary update completed, but primary update fails with this error message above. The same test passed on zeus branch.

patrickvacek:
@Raigi not to my knowledge. I can't explain that.

Raigi:
> @Raigi not to my knowledge. I can't explain that.

Ok, I'll try it out manually if it reproduces

Raigi:
This occurs when trying to install image on a device that was bitbaked differently: `bitbake core-image-minimal` , previously bitbaked with `bitbake primary-image secondary-image`. 

Do you think this is a problem? It used to also work this way before, but now it thinks that it is an ecu replacement.
`Jul 13 12:23:54 qemux86-64 aktualizr[253]: ECU serial 5501f7287fa535ce8866109f99b19ebd74775735cd8eef8eb9bba1c958195127 with hardware ID qemux86-64-sndry in storage was not found in Secondary configuration.
Jul 13 12:23:54 qemux86-64 aktualizr[253]: Initializer error: Server error: Error registering device: 412 {"code":"replace_ecu_assignment_exists","description":"Cannot replace ecus for DeviceId(a7c55af4-ebb7-4ed1-8efc-1ce6460a2614), the device has running assignments","cause":null,"errorId":"2c666da2-225e-48a9-aea4-187187356072"}
Jul 13 12:23:54 qemux86-64 systemd[1]: aktualizr.service: Main process exited, code=exited, status=1/FAILURE
Jul 13 12:23:54 qemux86-64 systemd[1]: aktualizr.service: Failed with result 'exit-code'.
`

steps:
1. bitbake primary-image secondary-image
2. create primary and secondary device
3. bitbake core-image-minimal
4. Install that image on primary

patrickvacek:
> This occurs when trying to install image on a device that was bitbaked differently: `bitbake core-image-minimal` , previously bitbaked with `bitbake primary-image secondary-image`.
> 
> Do you think this is a problem? It used to also work this way before, but now it thinks that it is an ecu replacement.

My guess is that the image bitbaked in step 3 does not know anything about Secondaries, so when it starts up and sees a Secondary in the database but not in the configuration, it tries to reregister. I wouldn't particularly recommend using `primary-image` and `core-image-minimal` images interchangeably. It can work if you make sure the configuration is suitable, but in this case, it sounds like it isn't.

Raigi:
> > This occurs when trying to install image on a device that was bitbaked differently: `bitbake core-image-minimal` , previously bitbaked with `bitbake primary-image secondary-image`.
> > Do you think this is a problem? It used to also work this way before, but now it thinks that it is an ecu replacement.
> 
> My guess is that the image bitbaked in step 3 does not know anything about Secondaries, so when it starts up and sees a Secondary in the database but not in the configuration, it tries to reregister. I wouldn't particularly recommend using `primary-image` and `core-image-minimal` images interchangeably. It can work if you make sure the configuration is suitable, but in this case, it sounds like it isn't.

Alritght, thanks. I'll update the tests.

PR #753, opened 2020-07-10
nfs-utils: fix nfs systemd daemons runtime issues

liuming50:
nfs-mountd.service and nfs-statd.service rely on /var/lib/nfs to be
present in rootfs, or else they would run into problems for some files
missing in it.

But a ostree/ota rootfs does not have that directory (removed during
do_image), to fix that, we now copy them to /var/local/lib/nfs and
mount it to /var/lib/nfs at runtime, which is done in
nfs-home-mount.service.

nfs-mountd.service and nfs-statd.service need run after
nfs-home-mount.service.

### DIFF
--- a/recipes-connectivity/nfs-utils/nfs-utils/nfs-home-mount.service
+++ b/recipes-connectivity/nfs-utils/nfs-utils/nfs-home-mount.service
@@ -1,0 +1,13 @@
+[Unit]
+Description=NFS home mount @LOCALSTATEDIR@/local/lib/nfs -> @LOCALSTATEDIR@/lib/nfs.
+DefaultDependencies=no
+Conflicts=umount.target
+After=local-fs.target
+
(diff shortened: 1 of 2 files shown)


patrickvacek:
Looks fine, I think, but why do you need the extra systemd script? It's not clear to me what it is doing.

liuming50:
> Looks fine, I think, but why do you need the extra systemd script? It's not clear to me what it is doing.

Hi, @patrickvacek:

Since both nfs-mountd.service and nfs-statd.service (no dependency between each other) are relying on a bind mount /var/local/lib/nfs -> /var/lib/nfs, so we need either a systemd service to do that and let them depend on it, or we make a script to do that (need checking if the directory already mounted in it), and add it to ExecStartPre= of both services, I choose the former solution. Or you might have other suggestions?

PR #754, opened 2020-07-26
nfs-utils: don't add nfs-home-mount.service without sota enabled

shr-project:
* nfs-home-mount.service is installed by do_install_append_sota, so only when sota is enabled
  but it was always added to SYSTEMD_SERVICE_ even when sota wasn't enabled resulting in
  do_patch failure:
  nfs-utils-2.5.1-r0 do_package: SYSTEMD_SERVICE_nfs-utils value nfs-home-mount.service does not exist

* use sota override everywhere

* fix issues introduced in
  https://github.com/advancedtelematic/meta-updater/pull/753


patrickvacek:
The CI failure looks unrelated to your PR. I'm looking into it. I don't see any problems with merging this but want to be sure our build system is doing the right thing first.

PR #756, opened 2020-07-29
qemucommand: Restore custom booloader parameter.

patrickvacek:
The functionality was (presumably unintentionally) removed, but the option was still present.


patrickvacek:
CI is gonna fail due to known problems, and this shouldn't affect CI anyway, so I'm proceeding with merging now.

PR #761, opened 2020-08-10
Remove boost 1.72 bbappend

psaavedra:
Backport necessary commit for dunfell from master

poky switched to 1.73 which includes this patch

Signed-off-by: Laurent Bonnans <laurent.bonnans@here.com>


patrickvacek:
> poky switched to 1.73 which includes this patch

I only see that poky has added this patch, but it doesn't look like they've actually upgraded to 1.73. Either way, this looks good to me. I'm running CI just to be sure. Thanks!

PR #762, opened 2020-08-10
image_repo_manifest.bbclass: explicitly call python3

liuming50:
Now we have changed to inherit python3native to provide python3 binary
for repo, then we need explicitly call python3 to execute repo, or else
it will firstly try to locate python which will fail if python is not
in HOSTTOOLS.

This fixes a following warning:
| Android repo tool failed to run; manifest not copied

### DIFF
--- a/classes/image_repo_manifest.bbclass
+++ b/classes/image_repo_manifest.bbclass
@@ -13,8 +13,9 @@ inherit python3native
 
 # Write build information to target filesystem
 buildinfo_manifest () {
-  if [ $(which repo) ]; then
-    repo manifest --revision-as-HEAD -o ${IMAGE_ROOTFS}${sysconfdir}/manifest.xml || bbwarn "Android repo tool failed to run; manifest not copied"
+  repotool=$(which repo)
+  if [ -n "$repotool" ]; then
+    python3 $repotool manifest --revision-as-HEAD -o ${IMAGE_ROOTFS}${sysconfdir}/manifest.xml || bbwarn "Android repo tool failed to run; manifest not copied"
   else
     bbwarn "Android repo tool not found; manifest not copied."
   fi


patrickvacek:
Are you sure this is the right solution? I think you might just need to upgrade the version of repo on your build machine, or install something like the `python-is-python3` package in Ubuntu.

liuming50:
> Are you sure this is the right solution? I think you might just need to upgrade the version of repo on your build machine, or install something like the `python-is-python3` package in Ubuntu.

Yes, I did not install `python-is-python3` which might help, but my repo tool seems is the latest one since it will update itself once being executed 'repo init'.

I think this solution could be sort of shoring up for old repo tools or for the users dont have python default to python3?

patrickvacek:
> I think this solution could be sort of shoring up for old repo tools or for the users dont have python default to python3?

Well, part of my concern here is that if your machine doesn't have a binary installed named `python3` because `python` is already v3, then you this will fail without an annoying workaround. I don't know how realistic that concern is, though; I'm mostly familiar with the Ubuntu/Debian world.

liuming50:
> > I think this solution could be sort of shoring up for old repo tools or for the users dont have python default to python3?
> 
> Well, part of my concern here is that if your machine doesn't have a binary installed named `python3` because `python` is already v3, then this will fail unless you use an annoying workaround. I don't know how realistic that concern is, though; I'm mostly familiar with the Ubuntu/Debian world.

Attempt to say that would not happen since we are now inheriting python3native, so python3 is from yocto sysroot instead of from host machine, this is why I started with this solution.

To verify that:
```
$ bitbake core-image-minimal -c devshell
$ which python3
/mnt/workspace/build/tmp/work/colibri_imx6-poky-linux-gnueabi/core-image-minimal/1.0-r0/recipe-sysroot-native/usr/bin/python3-native/python3
```

shr-project:
This PR breaks the build when there is no repo on host.

dash -x some-image/1.0-r2/temp/run.buildinfo_manifest.80233
```
+ export systemd_system_unitdir=/usr/lib/systemd/system
+ export systemd_unitdir=/usr/lib/systemd
+ export systemd_user_unitdir=/usr/lib/systemd/user
+ buildinfo_manifest
+ which repo
+ repotool=
+ bb_sh_exit_handler
+ ret=1
+ [ 1 != 0 ]
+ echo WARNING: exit code 1 from a shell command.
WARNING: exit code 1 from a shell command.
+ exit 1
```

shr-project:
https://github.com/advancedtelematic/meta-updater/pull/765 fixes this

liuming50:
Arha, thanks for the finding, yes, that looks like a proper fix.

//Ming Liu

shr-project <notifications@github.com> 於 2020年8月18日 週二 下午4:01寫道：

PR #763, opened 2020-08-11
Cherry pick nfs utils fixes to dunfell

liuming50:


liuming50:
@patrickvacek Could we have these two nfs-utils fixes for dunfell?

patrickvacek:
> Could we have these two nfs-utils fixes for dunfell?

Yes, it looks fine. I'm running tests now.

PR #772, opened 2020-09-03
sota: introduce OSTREE_MULTI_DEVICETREE_SUPPORT

liuming50:
ostree supports multiple devicetrees provided that there is no
/lib/modules/$kversion/devicetree installed on rootfs, in which case,
a uboot variable 'fdtdir' would be written to /boot/loader/uEnv.txt
instead of fdt_file.

Introduce OSTREE_MULTI_DEVICETREE_SUPPORT to control that, when it's
set to '0', there is no functional changes with current implementation,
while it's set to '1', then /lib/modules/$kversion/devicetree would not
be installed and the end users need set their own fdtfile/fdt_file in
boot script to locate the devicetree in 'fdtdir'.


patrickvacek (approved):
LGTM. We ran a test that failed where expected (unrelated to this) but otherwise looks good: https://main.gitlab.in.here.com/olp/edge/ota/connect/client/meta-updater/-/pipelines/1083447

PR #774, opened 2020-09-17
image_types_ostree: fix chmod

wilddom:
I'm pretty sure chown makes no sense here...


aodukha:
upd: pipelines fails for reasons not connected with this PR. approved.

aodukha (approved):

PR #775, opened 2020-09-28
Feat/rm upstreamed packages

patrickvacek:
Two packages have been added to meta-openembedded without us noticing:

* lshw
* softhsm

And two bbappends that were just adding native build support, but that is already available in the upstream version:

* gpgme
* libssh2

Hence we can remove all of these things.

These should probably get forward-ported to master and optionally backported where appropriate (I don't know if they are in the other repos in every branch we support).


atsjenkins (commented):
✅ OTF Tests **passed** for commit `24e1df0`.

aodukha (approved):

Raigi (approved):
Creating and updating rpi3 and 4 devices - OK

PR #777, opened 2020-10-15
ostree-kernel-initramfs: add back initramfs fit workaround

ricardosalveti:
Previous initramfs fit specific workaround was removed as part of commit
6eecf1593d, causing ostree to provide an init argument as part of the
kernel command line arguments when it is not really needed.

Bring back the workaround by simply generating an empty file in case
kernel image type is fit and initramfs is also used by the target.


ricardosalveti:
@liuming50 something I noticed when updating to latest meta-updater.

patrickvacek (approved):
I also wondered about that bit being removed. It looks correct to me to restore it. Some of the tests failed, but they are known to fail at present on master and they aren't particularly relevant to your changes. Thanks!

PR #779, opened 2020-10-21
aktualizr: Don't put unused configs in the image.

patrickvacek:
Because every aktualizr package was depending on aktualizr-configs, all
of the default configs were getting installed to /usr/lib/sota, despite
that by default none of them were ever used on the device. That
dependency has been removed.

The provisioning recipes still install the recipes they need to the
default location that aktualizr actually uses (/usr/lib/sota/conf.d). If
you want to use anything else, a specific recipe should be created to do
so.

This change rendered the aktualizr-host-tools package useless, so it has
been removed.

### DIFF
--- a/recipes-sota/aktualizr/aktualizr-device-prov-hsm.bb
+++ b/recipes-sota/aktualizr/aktualizr-device-prov-hsm.bb
@@ -7,7 +7,7 @@ LIC_FILES_CHKSUM = "file://${COMMON_LICENSE_DIR}/MPL-2.0;md5=815ca599c9df247a0c7
 
 inherit allarch
 
-# We need to get the config files from the aktualizr-host-tools package built by
+# We need to get the config files from the aktualizr-configs package built by
 # the aktualizr (target) recipe.
 DEPENDS = "aktualizr"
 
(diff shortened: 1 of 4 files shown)


patrickvacek:
@liuming50 can you take a look at this as well? This undoes some changes from https://github.com/advancedtelematic/meta-updater/pull/470 (specifically https://github.com/advancedtelematic/meta-updater/pull/470/commits/7ec1ffc436e8725e40e1853d084054b554929041). Why did you originally want to have every aktualizr package depend on the configs? None of the configs should be installed by default, only the ones that a user specifically wants.

aodukha (approved):

atsjenkins (commented):
✅ OTF Tests **passed** for commit `a1af0b4`.

liuming50:
> @liuming50 can you take a look at this as well? This undoes some changes from #470 (specifically [7ec1ffc](https://github.com/advancedtelematic/meta-updater/commit/7ec1ffc436e8725e40e1853d084054b554929041)). Why did you originally want to have every aktualizr package depend on the configs? None of the configs should be installed by default, only the ones that a user specifically wants.

@patrickvacek I am on board with this change, I think why I added ${PN}-configs to every single binary package was that I was not sure which binary is reading configs or not at that moment, so I did not want to break runtime behavior.

patrickvacek:
> I am on board with this change, I think why I added ${PN}-configs to every single binary package was that I was not sure which binary is reading configs or not at that moment, so I did not want to break runtime behavior.

Thanks for checking, and thanks again for the original work!

PR #778, opened 2020-10-16
layer.conf: add gatesgarth to LAYERSERIES_COMPAT

ricardosalveti:
Signed-off-by: Ricardo Salveti <ricardo@foundries.io>

PR #780, opened 2020-10-29
aktualizr: Upgrade to 2020.10 release

aodukha:
aktualizr: Upgrade to 2020.9 release.
Signed-off-by: Anatoliy Odukha <aodukha@gmail.com>

### DIFF
--- a/recipes-sota/aktualizr/aktualizr-device-prov-hsm.bb
+++ b/recipes-sota/aktualizr/aktualizr-device-prov-hsm.bb
@@ -7,7 +7,7 @@ LIC_FILES_CHKSUM = "file://${COMMON_LICENSE_DIR}/MPL-2.0;md5=815ca599c9df247a0c7
 
 inherit allarch
 
-# We need to get the config files from the aktualizr-host-tools package built by
+# We need to get the config files from the aktualizr-configs package built by
 # the aktualizr (target) recipe.
 DEPENDS = "aktualizr"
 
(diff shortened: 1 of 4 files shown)


patrickvacek:
I haven't touched master in a while, but if you want to update it, I would recommend forward-porting some of the recent changes that were merged directly to dunfell.

patrickvacek:
Specifically, I think this (and the other branches except dunfell) are missing the commits from https://github.com/advancedtelematic/meta-updater/pull/775.

Also, I don't see a PR for thud. I think we still have to support that for now, although maybe not forever.

patrickvacek:
Looks fine, but it's not great to use the merge commit; I prefer cherry-picking the individual commits from #775. Not a blocker, but it makes it a lot easier to compare what commits are where. The DCO check also isn't critical here, but normally I do `rebase master --signoff` to deal with that.

aodukha:
pushed cleaner version.

patrickvacek (approved):
LGTM. It'd be good to have CI running smoothly again, though.

aodukha:
> LGTM. It'd be good to have CI running smoothly again, though.

eh... I added ticket for this (OTA-5338)
Really configs are inconsistent and bitbake cannot build images. Probably #778 is responsible for the first error. Because other layers do not have gatesgarth.

patrickvacek:
> I added ticket for this (OTA-5338)

Sounds good, but I don't think that's the right ticket.

> Really configs are inconsistent and bitbake cannot build images. Probably #778 is responsible for the first error. Because other layers do not have gatesgarth.

Yeah, that's not helping. I think the RPi problems are still open issues here, too.

PR #785, opened 2020-10-31
Sync recipe/append removals from the dunfell branch

ricardosalveti:
Same changes apply at the master branch.

This is a set of cherry-picks from #775


patrickvacek:
LGTM, but @aodukha has included these changes in his PR: https://github.com/advancedtelematic/meta-updater/pull/780. I prefer the individual cherry-picks over squashing them into one commit, but I'll let him decide what to do from here.

ricardosalveti:
That works as well, but I would prefer landing these changes separately simply because they are really not related to the 2020.10 upgrade.

aodukha (approved):

aodukha:
it is better to have it as separate commits. it should not have conflicts with #780 (or I will update it)

PR #787, opened 2020-11-25
image_types_ota.bbclass: call oe_mkext234fs to make ota-ext4 image

liuming50:
We dont have to maintain our own function calculate_size and use
dd/mkfs.ext4 to generate ota-ext4 image, they have been done in OE
by get_rootfs_size/oe_mkext234fs functions, we could just use them.

The major benefit could be we can sync the future fixes/changes in
these functions from OE, also avoid maintaining some duplicated code in
image_types_ota.bbclass.

### DIFF
--- a/classes/image_types_ota.bbclass
+++ b/classes/image_types_ota.bbclass
@@ -1,42 +1,3 @@
-# Image to use with u-boot as BIOS and OSTree deployment system
-
-# Boot filesystem size in MiB
-# OSTree updates may require some space on boot file system for
-# boot scripts, kernel and initramfs images
-#
-calculate_size () {
-	BASE=$1
-	SCALE=$2
-	MIN=$3
-	MAX=$4
-	EXTRA=$5
-	ALIGN=$6
-
-	SIZE=`echo "$BASE * $SCALE" | bc -l`
-	REM=`echo $SIZE | cut -d "." -f 2`
-	SIZE=`echo $SIZE | cut -d "." -f 1`
-
-	if [ -n "$REM" -o ! "$REM" -eq 0 ]; then
-		SIZE=`expr $SIZE \+ 1`
-	fi
-
-	if [ "$SIZE" -lt "$MIN" ]; then
-		SIZE=$MIN
-	fi
-
-	SIZE=`expr $SIZE \+ $EXTRA`
-	SIZE=`expr $SIZE \+ $ALIGN \- 1`
-	SIZE=`expr $SIZE \- $SIZE \% $ALIGN`
-
-	if [ -n "$MAX" ]; then
-		if [ "$SIZE" -gt "$MAX" ]; then
-			return -1
-		fi
-	fi
-	
-	echo "${SIZE}"
-}
-
 OTA_SYSROOT = "${WORKDIR}/ota-sysroot"
 TAR_IMAGE_ROOTFS_task-image-ota = "${OTA_SYSROOT}"
 IMAGE_TYPEDEP_ota = "ostreecommit"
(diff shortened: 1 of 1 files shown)


liuming50:
Changes in patch set V2:

1 Add one more patch.

classes/image_types_ota.bbclass
```diff
@@ -108,24 +69,13 @@ IMAGE_CMD_ota () {
...
-	eval local MIN_COUNT=\"60\"
-	if [ ${OTA_ROOTFS_SIZE} -lt ${MIN_COUNT} ]; then
-		eval COUNT=\"${MIN_COUNT}\"
-	fi
-
-	dd if=/dev/zero of=${IMGDEPLOYDIR}/${IMAGE_NAME}${IMAGE_NAME_SUFFIX}.ota-ext4 seek=${OTA_ROOTFS_SIZE} count=${COUNT} bs=1024
-	mkfs.ext4 -O ^64bit ${IMGDEPLOYDIR}/${IMAGE_NAME}${IMAGE_NAME_SUFFIX}.ota-ext4 -L otaroot -d ${OTA_SYSROOT}
+	ln -sf ${STAGING_DIR_NATIVE}${base_sbindir_native}/mkfs.ext4 ${STAGING_DIR_NATIVE}${base_sbindir_native}/mkfs.ota-ext4
+	ln -sf ${STAGING_DIR_NATIVE}${base_sbindir_native}/fsck.ext4 ${STAGING_DIR_NATIVE}${base_sbindir_native}/fsck.ota-ext4
+	oe_mkext234fs ota-ext4 ${EXTRA_IMAGECMD}
```
patrickvacek, line 78: This looks cool, but I'm failing to see how the variables we used before to control the size get used by this function. I'm a bit worried about unexpected results here.

liuming50: @patrickvacek 

Previously we were using:
```
OTA_ROOTFS_SIZE=$(calculate_size `du -ks ${OTA_SYSROOT} | cut -f 1`  "${IMAGE_OVERHEAD_FACTOR}" "${IMAGE_ROOTFS_SIZE}" "${IMAGE_ROOTFS_MAXSIZE}" `expr ${IMAGE_ROOTFS_EXTRA_SPACE}` "${IMAGE_ROOTFS_ALIGNMENT}")
```

the logic is basically same with OE's function get_rootfs_size, which is being called for every image task in prefuncs, if you could add some print in openembedded-core/meta/classes/image.bbclass, for instance:
```
diff --git a/meta/classes/image.bbclass b/meta/classes/image.bbclass
index adc50c99bc..06cac9341a 100644
--- a/meta/classes/image.bbclass
+++ b/meta/classes/image.bbclass
@@ -518,6 +518,7 @@ def get_rootfs_size(d):
     initramfs_fstypes = d.getVar('INITRAMFS_FSTYPES') or ''
     initramfs_maxsize = d.getVar('INITRAMFS_MAXSIZE')
 
+    bb.warn("calling du -ks to: %s" % d.getVar('IMAGE_ROOTFS'))
     output = subprocess.check_output(['du', '-ks',
                                       d.getVar('IMAGE_ROOTFS')])
     size_kb = int(output.split()[0])
```

you will see how it's getting called for do_image_ota_ext4

patrickvacek: @aodukha is this something you can verify? I'm a bit too busy at the moment to dig into the details of how the image sizes are calculated.

aodukha: in order to verify it I will builds rapspi 3 & quemu (should not be affected at all) and flash pi3-4. but it will take some time. 
IMHO this should not affect our builds. but make sense integrate into master firstly and check differences between returned values from get_rootfs_size and current values dunfell/warrior/etc. It is possible (but unlikely) that somebody depends on it

patrickvacek: Yeah, I think just building a couple images and making sure the sizes look good is worth checking. I think this PR looks good and I'd prefer to use standardized code instead of this older, cruftier custom code, but I need some more confidence that it won't do something unexpected.

Since master CI is ugly, it shouldn't be hard to try this change out on dunfell instead of master if that helps.

aodukha: at least for qemu build existing version and PR calls dd with 209767. hm.. similar numbers (checked twice, have to check rasp version too)
make sense run it on CI. it seems to have identical functionality to existing code and a bit smaller code

patrickvacek: Cool, if it's similar, then it's probably good to go. I just wanted to make sure it was double-checked.

aodukha: Approved. there are no 
flashed and run rasp3 and quemu with this. Using local build for this.
I run CI for dunfell I https://main.gitlab.in.here.com/olp/edge/ota/connect/client/meta-updater/-/pipelines/1706123

aodukha (approved):

classes/image_types_ota.bbclass
shr-project: This doesn't seem to work very well as it produces ext2 instead of ext4 now, here are the fixes for master and dunfell:
https://github.com/advancedtelematic/meta-updater/pull/790
https://github.com/advancedtelematic/meta-updater/pull/791

liuming50: > This doesn't seem to work very well as it produces ext2 instead of ext4 now, here are the fixes for master and dunfell:
> #790
> #791

Yes, my bad, "-t ext4" is mandatory, but I have a concern about these pull request, had some comments on: https://github.com/advancedtelematic/meta-updater/pull/790/commits/64c24452a850001223df1054bfcbb1a9010732ac

aodukha: :( will run CI asap to check the fix

PR #788, opened 2020-12-15
Use python3 to get repo working again.

pattivacek:
The latest version of repo requires python3.6+, but the default Debian package is fine with the default python3 package (although it complains about experimental support).


aodukha (approved):

pattivacek:
Merging because this fixes the part that was broken, and we expect most of the actual tests to continue to fail on master.

PR #791, opened 2020-12-29
[master] Fix ota-ext4 IMAGE_FSTYPE

shr-project:


shr-project:
@aodukha, I've updated this one to match dunfell.

ricardosalveti:
This patch fixes the problem we currently have in master, which generates an ext2 fs, so +1.

One thing when comparing the older behavior, is that metadata_csum was enabled by default, so unsure why 4d34fa53db086955adf6bc58941ebf98a16778f8 was done (@liuming50 maybe you can extend here?).

I think we should just switch to the default options here, which includes 64 and metadata_csum.

ricardosalveti:
Using dumpe2fs on the ota-ext4 image created with the older logic:

```
Filesystem volume name:   otaroot
Last mounted on:          <not available>
Filesystem UUID:          be5dc5f0-d74b-4886-8077-b0dae6fab0c2
Filesystem magic number:  0xEF53
Filesystem revision #:    1 (dynamic)
Filesystem features:      has_journal ext_attr resize_inode dir_index filetype extent flex_bg sparse_super large_file huge_file dir_nlink extra_isize metadata_csum
Filesystem flags:         signed_directory_hash
Default mount options:    user_xattr acl
```

liuming50:
Hi, @ricardosalveti  

> One thing when comparing the older behavior, is that metadata_csum was enabled by default, so unsure why [4d34fa5](https://github.com/advancedtelematic/meta-updater/commit/4d34fa53db086955adf6bc58941ebf98a16778f8) was done (@liuming50 maybe you can extend here?).

Because uboot does not support saving env in EXT4 with metadata_csum enabled:
| Saving Environment to EXT4... Unsupported feature metadata_csum found, not writing

and some customers might like to save uboot in OTA partition. (For instance, if they dont have a boot partition)
 
> I think we should just switch to the default options here, which includes 64 and metadata_csum.

I am not sure about that, 64bit has been disabled at very beginning since this bbclass ported from AGL, now we enable it without knowing why it's disabled at the first place, that might be risky?

liuming50:
@ricardosalveti why not just add "-t ext4" to the current EXTRA_IMAGECMD_ota-ext4?

ricardosalveti:
> Hi, @ricardosalveti
> 
> > One thing when comparing the older behavior, is that metadata_csum was enabled by default, so unsure why [4d34fa5](https://github.com/advancedtelematic/meta-updater/commit/4d34fa53db086955adf6bc58941ebf98a16778f8) was done (@liuming50 maybe you can extend here?).
> 
> Because uboot does not support saving env in EXT4 with metadata_csum enabled:
> | Saving Environment to EXT4... Unsupported feature metadata_csum found, not writing
> 
> and some customers might like to save uboot in OTA partition. (For instance, if they dont have a boot partition)

Sure, but why would we have to change the default ext4 options because of one use case and tooling? Remember the same ext4 image will be used by other systems that won't even require u-boot.

I think we should fix the u-boot tools instead. Also, this option was enabled by default before as well.

> 
> > I think we should just switch to the default options here, which includes 64 and metadata_csum.
> 
> I am not sure about that, 64bit has been disabled at very beginning since this bbclass ported from AGL, now we enable it without knowing why it's disabled at the first place, that might be risky?

I tried to look back at what is provided in the tree and couldn't find the exact reason why it was added. I would guess it was added because the userspace tools were not supporting that option back when it was introduced, which is not the case anymore for most distros.

ricardosalveti:
> @ricardosalveti why not just append "-t ext4" to the current EXTRA_IMAGECMD_ota-ext4?

Sure, this is what this PR is proposing, but when comparing with the previous options I noticed metadata_csum was also set, which is a side effect of the same PR (that made it create ext2 fs instead).

liuming50:
@ricardosalveti 

> Sure, but why would we have to change the default ext4 options because of one use case and tooling? Remember the same ext4 image will be used by other systems that won't even require u-boot.

OK, I am on board.

> I tried to look back at what is provided in the tree and couldn't find the exact reason why it was added. I would guess it was added because the userspace tools were not supporting that option back when it was introduced, which is not the case anymore for most distros.

Yes, I also tried that, could not find the exact reason why it was added

ricardosalveti:
Cool, once this PR gets merged we can propose another one to change the default options used.

aodukha:
ci is ok. https://main.gitlab.in.here.com/olp/edge/ota/connect/client/meta-updater/-/pipelines/1855293

aodukha (approved):

PR #790, opened 2020-12-29
[dunfell] Fix ota-ext4 IMAGE_FSTYPE

shr-project:

PR #792, opened 2020-12-31
Revert "[dunfell] Fix ota-ext4 IMAGE_FSTYPE"

aodukha:
Reverts advancedtelematic/meta-updater#790 because of error on boot


aodukha (commented):
approved

PR #798, opened 2021-01-19
image_repo_manifest.bbclass: Improvement for builds outside the .repo…

wilddom:
… directory

The repo tool searches up the directory tree to find the .repo directory,
which doesn't work in a separated build directory.

### DIFF
--- a/classes/image_repo_manifest.bbclass
+++ b/classes/image_repo_manifest.bbclass
@@ -15,7 +15,7 @@ inherit python3native
 buildinfo_manifest () {
   repotool=`which repo || true`
   if [ -n "$repotool" ]; then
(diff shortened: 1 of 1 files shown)


pattivacek (commented):
Where does `THISDIR` come from? I think I see the point here and it should be fine, but I want to make sure it does what I expect.

wilddom:
Yocto Manual:
> `THISDIR` The directory in which the file BitBake is currently parsing is located

`THISDIR` was actually not exactly what I was searching for, but at least it worked. The `.repo` directory is usually at the top of the layers' directory and because `THISDIR` is the directory of the currently parsed `bb` file the repo tool should find the `.repo`directory from there.

pattivacek (approved):
Great, thanks! I took a look around and it looks fine to me.

PR #801, opened 2021-02-23
ostree-kernel-initramfs: update LIC_FILES_CHKSUM

shr-project:
* the license file was renamed in:
  https://git.openembedded.org/openembedded-core/commit/?id=5ecf139a31fa7bd813855f1235ea9f434fbcb2e0

  and the build with latest oe-core is now failing with:
  WARNING: Could not copy license file TOPDIR/oe-core/meta/files/common-licenses/GPL-2.0 to TOPDIR/BUILD/work/raspberrypi3-oe-linux-gnueabi/ostree-kernel-initramfs/0.0.1-r0/license-destdir/ostree-kernel-initramfs/GPL-2.0: [Errno 2] No such file or directory: 'TOPDIR/oe-core/meta/files/common-licenses/GPL-2.0'

  ERROR: QA Issue: ostree-kernel-initramfs: LIC_FILES_CHKSUM points to an invalid file: TOPDIR/oe-core/meta/files/common-licenses/GPL-2.0 [license-checksum]
  ERROR: Fatal QA errors found, failing task.

* unfortunately this would be first meta-updater commit which
  isn't backwards compatible with gatesgarth (for which there
  isn't separate branch). So either gatesgarth branch should
  be created before this or maybe we should discuss if this
  LICENSE value is actually correct (e.g. MIT license files
  weren't renamed) and this recipe doesn't have own source,
  the initramfs itself should be covered by the recipe which
  provides whatever is set in INITRAMFS_IMAGE and kernel from
  virtual/kernel, but ostree-kernel-initramfs recipe is only
  metadata, so it might make sense to use MIT (as COPYING.MIT
  in the root of the repo) or MPL-2 like other metadata-only
  recipes in this layer (is it intentionally MPL-2 instead of
  more common for metadata MIT?):

recipes-connectivity/networkd-dhcp-conf/networkd-dhcp-conf.bb:LIC_FILES_CHKSUM = "file://${COMMON_LICENSE_DIR}/MPL-2.0;md5=815ca599c9df247a0c7f619bab123dad"
recipes-sota/aktualizr/aktualizr-collectd.bb:LIC_FILES_CHKSUM = "file://${COMMON_LICENSE_DIR}/MPL-2.0;md5=815ca599c9df247a0c7f619bab123dad"
[...]

PR #802, opened 2021-02-24
ostree-kernel-initramfs: Use MIT license.

pattivacek:
It's a metadata-only recipe and all contributors have agreed that MIT would be preferable.

FYI @shr-project. This effectively replaces https://github.com/advancedtelematic/meta-updater/pull/801.


shr-project (approved):

pattivacek:
No need to test, so I'm going ahead with merging.

PR #803, opened 2021-03-02
ostree-kernel-initramfs: Use MIT license.

shr-project:
It's a metadata-only recipe and all contributors have agreed that MIT
would be preferable.

Signed-off-by: Patti Vacek <pattivacek@gmail.com>


shr-project:
Looks like recipe rename is on its way to dunfell as well:
https://git.openembedded.org/openembedded-core-contrib/commit/?h=stable/dunfell-nut&id=28e915bf8c89b40fc3c0431a3cd2ec7d85bca4b7
so better to backport this from master in advance to stay compatible with it.

pattivacek (approved):
LGTM, thanks!

PR #805, opened 2021-04-03
Fix big-update recipe

shr-project:
We had this PNBLACKLISTed for long time in webOS OSE builds, but fixing it is quite easy.


pattivacek (approved):

pattivacek:
LGTM, thanks!

PR #808, opened 2021-04-08
image_types_ostree.bbclass: add fakeroot varflag to prepare_ostree_ro…

liuming50:
…otfs

Since the ostree rootfs is generated by fakeroot, we should make
prepare_ostree_rootfs run with fakeroot as well, or else we might run
into a pseudo abort issue when removing OSTREE_ROOTFS directory.

Reference:
https://wiki.yoctoproject.org/wiki/Pseudo_Abort

### DIFF
--- a/classes/image_types_ostree.bbclass
+++ b/classes/image_types_ostree.bbclass
@@ -31,6 +31,7 @@ python prepare_ostree_rootfs() {
     image_rootfs = d.getVar("IMAGE_ROOTFS")
     oe.path.copyhardlinktree(image_rootfs, ostree_rootfs)
 }
+prepare_ostree_rootfs[fakeroot] = "1"
 
 do_image_ostree[dirs] = "${OSTREE_ROOTFS}"
 do_image_ostree[prefuncs] += "prepare_ostree_rootfs"


pattivacek (commented):
I don't understand what's going on here. Where does fakeroot get involved with ostree, and if it is necessary to change this task, why wouldn't we need to change all the others, too?

@aodukha Please take a look and see what you think.

liuming50:
> I don't understand what's going on here. Where does fakeroot get involved with ostree, and if it is necessary to change this task, why wouldn't we need to change all the others, too?
> 
> @aodukha Please take a look and see what you think.

@pattivacek @aodukha I would like to explain some details here.

All the updater image tasks ostree/ostreecommit/ostreepush/garagesign run under fakeroot, this is done in https://github.com/openembedded/openembedded-core/blob/master/meta/classes/image.bbclass#L494

So we need let their prefuncs/posfuncs (currently only one: prepare_ostree_rootfs) also run under fakeroot, or else it may run into the problems described in https://wiki.yoctoproject.org/wiki/Pseudo_Abort

I have observed this issue when I added meta-security/meta-integrity, and have verified it could be fixed by this patch.

pattivacek (approved):
@liuming50 CI has some issues right now, but from what I can tell, this looks fine. Thanks!

PR #810, opened 2021-06-19
image_types_ostree.bbclass: drop prepare_ostree_rootfs prefunc

liuming50:
The purpose of prepare_ostree_rootfs prefunc is to ensure
do_image_rootfs runs from a clean ${OSTREE_ROOTFS} and call
oe.path.copyhardlinktree from ${IMAGE_ROOTFS} to ${OSTREE_ROOTFS}.

We dont have to maintain a prefunc to achieve that, it could be easily
done in do_image_ostree itself.

But the major reason for this change is to avoid hard links from
${IMAGE_ROOTFS} to ${OSTREE_ROOTFS}, which might lead to a Pseudo
abortion in some cases, which was observed when IMA/EVM is enabled
in rootfs.

There was a commit aiming to fix the Pseudo abortion, commit db099053:
[ image_types_ostree.bbclass: add fakeroot varflag to prepare_ostree_rootfs ]

but it's incomplete per later tests, there is still the case some old
inodes being used in ${IMAGE_ROOTFS} when do_rootfs runs, which causes
'rm -rf ${IMAGE_ROOTFS}' fail on a Pseudo abortion error.

This fixes the Pseudo abortion completely.

### DIFF
--- a/classes/image_types_ostree.bbclass
+++ b/classes/image_types_ostree.bbclass
@@ -19,24 +19,13 @@ CONVERSIONTYPES_append = " tar"
 
 TAR_IMAGE_ROOTFS_task-image-ostree = "${OSTREE_ROOTFS}"
 
-python prepare_ostree_rootfs() {
-    import oe.path
-    import shutil
-
-    ostree_rootfs = d.getVar("OSTREE_ROOTFS")
-    if os.path.lexists(ostree_rootfs):
-        bb.utils.remove(ostree_rootfs, True)
-
-    # Copy required as we change permissions on some files.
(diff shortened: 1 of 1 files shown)


pattivacek (approved):
Thanks! This looks good and seems like an obvious improvement to me. @aodukha can you run the tests on this? I no longer have a way to do that.

pattivacek:
Well, I'm gonna go ahead and merge this...

aodukha:
I`m back. there are small issues with pipelines (gr.. again python) I will fix it late, for now it is ok merge it

PR #811, opened 2021-06-30
meta: support readonly sysroot

liuming50:
Currently ostree sysroot is being mounted as RW by ostree-remount
systemd service, add a variable OSTREE_SYSROOT_READONLY to control
that, so the end users could choose to mount sysroot as RO.


pattivacek (approved):
LGTM!

sergioprado:
@pattivacek moving forward, should we submit PRs here or in the uptane space?

pattivacek:
@sergioprado in the long term, I think moving towards the Uptane namespace is preferred. I no longer work for ATS/Here and I've been busy with other things, but for now I'm still keeping an eye on both. AFAIK the testing infrastructure at Here isn't in a great place at present, and since it was never publicly visible anyway, I'd like it if we could somehow get a better setup in place. I think the Uptane namespace would be a better place to do that work.

sergioprado:
That sounds great @pattivacek. I guess this is also valid for other repositories like aktualizr. Moving forward, we will send PRs against the Uptane namespace. Thanks!

ricardosalveti:
> That sounds great @pattivacek. I guess this is also valid for other repositories like aktualizr. Moving forward, we will send PRs against the Uptane namespace. Thanks!

Shouldn't we free the ones from advancedtelematic then? Or make one a mirror from the other?

Having two active projects/branches is quite confusing :-)

pattivacek:
> Shouldn't we freeze the ones from advancedtelematic then? Or make one a mirror from the other?
> 
> Having two active projects/branches is quite confusing :-)

Yes, probably. I don't think I can make that change, though, and if Here continues to use the advancedtelematic namespace, I can't stop them. I would prefer to unify in the uptane namespace and I would love it if we could have some testing infrastructure around that.

PR #813, opened 2021-07-09
image_types_ota.bbclass: add support for syslinux as OSTREE_BOOTLOADER

ricardosalveti:
Add support for ostree to automatically identify and use syslinux as the
preferred bootloader scheme when set via OSTREE_BOOTLOADER.

Not forcing a build-time dependency at do_image_ota as multiple
bootloaders are able to identify the syslinux format.


pattivacek (approved):
LGTM and shouldn't affect existing tests. @aodukha do you want to take a look?

PR #816, opened 2021-07-30
Convert to new override syntax

shr-project:


classes/image_types_ostree.bbclass
```diff
@@ -129,7 +129,7 @@ IMAGE_CMD_ostree () {
 IMAGE_TYPEDEP_ostreecommit = "ostree"
 do_image_ostreecommit[depends] += "ostree-native:do_populate_sysroot"
 do_image_ostreecommit[lockfiles] += "${OSTREE_REPO}/ostree.lock"
-IMAGE_CMD_ostreecommit () {
+IMAGE_CMD:ostreecommit () {
```
pattivacek, line 132: I'm a bit confused about the syntax changes here. Is this right that the `IMAGE_CMD` bits need colons, but `IMAGE_TYPEDEP` doesn't? Seems inconsistent to me. Anyway, these aren't overrides but rather task names... unless the underlying mechanism is somehow the same.

classes/sota.bbclass
```diff
@@ -46,11 +46,11 @@ GARAGE_TARGET_EXPIRE_AFTER ?= ""
...
 SOTA_MACHINE ??="none"
-SOTA_MACHINE_rpi ?= "raspberrypi"
+SOTA_MACHINE:rpi ?= "raspberrypi"
 SOTA_MACHINE_porter ?= "porter"
 SOTA_MACHINE_m3ulcb = "m3ulcb"
 SOTA_MACHINE_intel-corei7-64 ?= "minnowboard"
-SOTA_MACHINE_qemux86-64 ?= "qemux86-64"
+SOTA_MACHINE:qemux86-64 ?= "qemux86-64"
 SOTA_MACHINE_am335x-evm ?= "am335x-evm-wifi"
 SOTA_MACHINE_freedom-u540 ?= "freedom-u540"
```
pattivacek, line 55: Seems like some items here are skipped.

pattivacek (commented):
Thanks for working on this! Looks mostly good, but there are some edge cases here where I don't know what's right. I don't currently have a good way to test this, either, so I'm afraid I can't really judge if you've found everything or not.

classes/sota.bbclass
shr-project: Yes, the list of possible overrides is hardcoded in the conversion script, so these should be migrated as well.

conf/layer.conf
```diff
@@ -6,13 +6,13 @@ BBFILES += "${LAYERDIR}/recipes-*/*/*.bb \
...
+BBFILE_PRIORITY_sota = "7"
 
-LAYERDEPENDS:sota = "openembedded-layer"
-LAYERDEPENDS:sota += "meta-python"
-LAYERDEPENDS:sota += "filesystems-layer"
-LAYERSERIES_COMPAT:sota = "dunfell gatesgarth hardknott"
+LAYERDEPENDS_sota = "openembedded-layer"
+LAYERDEPENDS_sota += "meta-python"
+LAYERDEPENDS_sota += "filesystems-layer"
+LAYERSERIES_COMPAT_sota = "dunfell gatesgarth hardknott"
```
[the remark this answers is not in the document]
shr-project: This part is correct, unfortunately "sota" here is used as a layer name while elsewhere "sota" is used as an OVERRIDE.

classes/image_types_ostree.bbclass
shr-project: Currently this is correct as IMAGE_TYPEDEP is used like this:
```
meta/classes/image.bbclass:        deps = (d.getVar('IMAGE_TYPEDEP_' + t) or "").split()
meta/classes/image.bbclass:        vardeps.add('IMAGE_TYPEDEP_' + t)
```
while IMAGE_CMD was changed to:
```
meta/classes/image.bbclass:        vardepsexclude = (d.getVarFlag('IMAGE_CMD:' + realt, 'vardepsexclude', True) or '').split()
meta/classes/image.bbclass:        image_cmd = localdata.getVar("IMAGE_CMD")
meta/classes/image.bbclass:        vardeps.add('IMAGE_CMD:' + realt)
meta/classes/image.bbclass:            bb.fatal("No IMAGE_CMD defined for IMAGE_FSTYPES entry '%s' - possibly invalid type name or missing support class" % t)
meta/classes/image.bbclass:        # Since a copy of IMAGE_CMD:xxx will be inlined within do_image_xxx,
meta/classes/image.bbclass:        # prevent a redundant copy of IMAGE_CMD:xxx being emitted as a function.
meta/classes/image.bbclass:        d.delVarFlag('IMAGE_CMD:' + realt, 'func')
```

But I've asked Richard if it should be changed in oe-core as well to avoid this confusion.

pattivacek (dismissed):
This looks okay to me; @aodukha do you want to take a look?

aodukha:
unfortunatly I cannot check it right now, but I  may merge it into master. and check later. @shr-project  Is it finished? there was last changes 9hr ago.

shr-project:
> unfortunatly I cannot check it right now, but I may merge it into master. and check later. @shr-project Is it finished? there was last changes 9hr ago.

@aodukha 
There were more changes added in oe-core/master which required the 4th commit to be added, but it's finished for now. If there are more changes needed later I'll create follow-up PR.

classes/image_types_ostree.bbclass
shr-project: Richard changed the IMAGE_TYPEDEP and CONVERSION_CMD to use : as well, it was added to oe-core/master yesterday and I've added 4th commit here to do the same.

pattivacek (approved):
Thanks for staying on top of it!

aodukha (approved):

PR #829, opened 2022-08-05
Compatibility with kirkstone

albard00:
Hi,

We updated meta-updater internally to be compatible with kirkstone. Maybe you can reuse what we did.

Unfortunately I think that with the removal of the reproducible_build_simple the layer is not compatible anymore with previous versions, although I don't really understand why you used the "simple" variant of reproducible build.

### DIFF
--- a/conf/distro/sota.conf.inc
+++ b/conf/distro/sota.conf.inc
@@ -7,13 +7,9 @@
 DISTRO_FEATURES:append = " sota usrmerge"
 DISTRO_FEATURES_NATIVE:append = " sota"
 INHERIT += " sota"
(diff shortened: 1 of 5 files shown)


pattivacek:
These changes look fine, but most of them are already present in the more regularly maintained repo in the [uptane namespace](https://github.com/uptane/meta-updater). See https://github.com/uptane/meta-updater/pull/13 in particular for Kirkstone support. Note that there still isn't a separate kirkstone branch; it's just on master still for now.

albard00:
Thanks for the information, I did not notice there was another repo.

PR #834, opened 2025-05-17
ostree-initrd: move /run mount point to sysroot/run

codeshravan:
This change relocates the /run mount point from initramfs (/run) to the real root filesystem (sysroot/run). This ensures that systemd does not remount it during switchroot if it detects an existing mount point. Consequently, all contents of /run will be transferred to the real root filesystem.
This adjustment resolves the issue of mounting sysroot as  read-only. During the execution of ostree-remount.service, the system checks /run/ostree-booted to determine if sysroot should be mounted as read-only.


shrkum:
Closing the request as updated in some another repo

codeshravan:
Closing this as updated in uptane/meta-updater

